PR #173, opened 2019-09-08
[ZMR] - Update usage of Searcher::add_reader to new signature

codnee:
I am not sure if it makes sense, but it makes the app build. 😄 

Changed in https://github.com/kiwix/kiwix-lib/commit/1e94665c0791b951a24fa0a831ff6ef341a85cc8/


kelson42:
Could you please open a bug and PR on libkiwix instead?

rgaudin:
Yes, it's harmless and allows using latest likiwix as `reader.first` is the identifier string and `reader.second` the actual reader.

We shall still look into how this ABI change affects this structure as this identifier might not be needed anymore.

PR #276, opened 2020-06-09
travis setup

automactic:


kelson42:
This is sad you don't use Github Actions which is now the standard for our project (and obviously more efficient).

PR #301, opened 2020-08-11
Update README.md

thesenatorO5-2:
[Minor edit]: some minor fixes


README.md
```diff
@@ -1,4 +1,4 @@
-# Kiwix for iOS & macOS
+# Kiwix for iOS and macOS
```
automactic, line 1: nit: what is the rationale of this change? They are semantically the same

Issue #159, opened 2019-07-05
Introduce Continous Integration


kelson42:
The current repository has neither a CI (Continuous Integration) nor a CD (Continuous Deployement).

This is unoptimal because:
* Without CI (testing the compilation, the app with a automated tests, a code coverage solution) we have no baseline security to avoid regressions
* Without CI, no CD is possible and the pushing of new releases relies fully on a developer doing manually the operation. A way of doing which is unsecure and create a SPOC at the developer in charge.
* Without CD, no nightly either
* A complementary problem of this is that this is not really feasible because the core libraries (libkiwix and libzim) are not prebuilt/prepackaged for macOS. This is a blocker for this task, but doing it properly
  * Secure the quality of the software (no manual configuration)
  * Simplify the build process which should allow to onboard new (junior) developer (a thing which is "impossible" for the moment due to the complexity and the requirement of another field of expertise which is needed by libzim/libkiwix compilation).

kelson42:
I will try to implement it.

yash982000:
Continuous Integration (CI) is a development practice where developers integrate code into a shared repository frequently, preferably several times a day. Each integration can then be verified by an automated build and automated tests. ... Among them is a revision control, build automation and automated testing.

kelson42:
Github actions is far better and meanwhile almost everywhere used at @openzim and @Kiwix.

kelson42:
@automactic @rgaudin @mgautierfr I have rewritten this already 4 years old ticket for more clarity. To me the priority of this task a raised continously during these pas years considering the increase of our audience and the improvement of our internal processes. AFAIK this is one of the latest (if not the latest) repository without CI and CD.

This tickets depends on:
* libkiwix
  * https://github.com/kiwix/kiwix-build/issues/455
  * https://github.com/kiwix/kiwix-build/issues/418
* libzim
  * https://github.com/kiwix/kiwix-build/issues/590

kelson42:
This CI should, via the GitHub Action, compile from scratch for both iOS and macOS (once for x86, once for ARM) the app each time we have a PR. The user should be able via the CI result to know if its PR compiles. [Libzim Github Action workflow](https://github.com/openzim/libzim/blob/main/.github/workflows/ci.yml) can be taken as an example.

PR #312, opened 2020-08-22
travis

automactic:


kelson42:
> > Here it would have avoid you to loose time with travis and me to loose time asking you what you want to achieve... thing which is still unclear.
> 
> This sentence does not make a whole lot of sense. If you wish me to use GitHub actions, perhaps you should have noted in ticket like #159 that GitHub action is preferred.

yes, at the time the ticket has been created it was not free for opensource project, see:
https://github.blog/2019-08-08-github-actions-now-supports-ci-cd/

> Also why are there so many rules? If I knew I would get into so much bickering over this I wouldn't have bothered to create a PR on a Sat morning

rules?

PR #495, opened 2023-05-03
Download task on force quit

automactic:
Fixing an issue where task is not labeled as failed when user force quit the app.

### DIFF
--- a/Model/DownloadService.swift
+++ b/Model/DownloadService.swift
@@ -23,7 +23,6 @@ class DownloadService: NSObject, URLSessionDelegate, URLSessionTaskDelegate, URL
         configuration.allowsCellularAccess = true
         configuration.isDiscretionary = false
         configuration.sessionSendsLaunchEvents = true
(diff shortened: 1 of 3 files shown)


Model/DownloadService.swift
```diff
@@ -23,16 +23,15 @@ class DownloadService: NSObject, URLSessionDelegate, URLSessionTaskDelegate, URL
         configuration.allowsCellularAccess = true
         configuration.isDiscretionary = false
         configuration.sessionSendsLaunchEvents = true
-        configuration.timeoutIntervalForResource = 300
```
automactic, line 26: This time out is removed, because `timeoutIntervalForResource` is for the timeout period for the whole download task, and downloads usually takes way more than 5 mins to complete.

`timeoutIntervalForRequest` is the correct one to use, but when set I never get call back from `URLSession` even when I enable flight mode. So I am not sure there is a way to do timeout without implementing it myself

PR #496, opened 2023-05-07
Implementing HTTP 206

automactic:
Implementing HTTP 206 Partial Content based on spec from https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Notification for when a download task complete
 - Improved handling and error surfacing for failed download tasks
 - Fixed an issue where tasks are not labeled as failed when app is force quit (iOS & iPadOS)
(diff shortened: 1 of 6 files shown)


kelson42:
@automactic Wonder a bit why Kiwix downloader has to handle the HTTP at this level of details. No way to deal using a library which "does do the job"?

automactic:
@kelson42 Yea I know there is `zim::Item::getData`. The point right now is to experiment with if implementing HTTP 206 will resolve issue related to video

```
/** Get the data associated to the item
 *
 * Get the `size` bytes of data of the item, starting at offset.
 *
 * @param offset The number of byte to skip at begining of the data.
 * @param size The number of byte to read.
 * @return A blob corresponding to the data.
 */
Blob getData(offset_type offset, size_type size) const;
```

PR #526, opened 2023-10-02
Fix/518 - new tab on macos

### LINKED ISSUE #518: Non functional "Open Link in New Window"
Working with version 3.1 on Monterey (12.6.8): By selecting a link in the article with right mouse button, the context menu entry " Open Link in New Window" don't work. Nothing happens. No new Window.

anthony-bernardo:
I refactor the way the alert is receiving notification, because with `NotificationCenter` (if there were opened tabs) the alert were displayed in each tabs

## TODO: 

- [ ] Squash commit before merge

Fixes #518

### DIFF
--- a/ViewModel/BrowserViewModel.swift
+++ b/ViewModel/BrowserViewModel.swift
@@ -392,11 +392,6 @@ class BrowserViewModel: NSObject, ObservableObject,
             }
         }
     }
-    
-    // TODO: Utils
-    func isExternalLink(url: URL) -> Bool {
-        return url.scheme == "http" || url.scheme == "https"
-    }
 
     // MARK: - New tab in macos desktop
       
--- a/ViewModel/BrowserViewModel.swift
+++ b/ViewModel/BrowserViewModel.swift
@@ -37,6 +37,7 @@ class BrowserViewModel: NSObject, ObservableObject,
     
     // MARK: - Properties
     
+    @Published var openExternalLinkURL = URL(string: "")
     @Published private(set) var canGoBack = false
     @Published private(set) var canGoForward = false
     @Published private(set) var articleTitle: String = ""
@@ -179,7 +180,7 @@ class BrowserViewModel: NSObject, ObservableObject,
         } else if url.isKiwixURL {
             decisionHandler(.allow)
         } else if url.scheme == "http" || url.scheme == "https" {
-            NotificationCenter.default.post(name: .externalLink, object: nil, userInfo: ["url": url])
+            self.openExternalLinkURL = url
             decisionHandler(.cancel)
         } else if url.scheme == "geo" {
             if FeatureFlags.map {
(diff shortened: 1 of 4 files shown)


ViewModel/BrowserViewModel.swift
```diff
@@ -179,7 +180,7 @@ class BrowserViewModel: NSObject, ObservableObject,
         } else if url.isKiwixURL {
             decisionHandler(.allow)
         } else if url.scheme == "http" || url.scheme == "https" {
-            NotificationCenter.default.post(name: .externalLink, object: nil, userInfo: ["url": url])
```
automactic, line 182: Could you remove this as well? (I don't think we need this anymore, right?)

https://github.com/kiwix/apple/blob/fa397a905f95b946e1cf6b55328771b101bb91f8/SwiftUI/Patches.swift#L65

Views/ViewModifiers/ExternalLinkHandler.swift
```diff
@@ -11,12 +11,11 @@ import SwiftUI
 import Defaults
 
 struct ExternalLinkHandler: ViewModifier {
+    @Binding var externalURL: URL?
```
automactic, line 14:
```suggestion
    @Binding var externalLinkURL: URL?
```
(following the 2nd feedback from above)

Views/ViewModifiers/ExternalLinkHandler.swift
```diff
@@ -28,8 +27,10 @@ struct ExternalLinkHandler: ViewModifier {
     }
     
     func body(content: Content) -> some View {
-        content.onReceive(externalLink) { notification in
-            guard let url = notification.userInfo?["url"] as? URL else { return }
+        content.onChange(of: externalURL, perform: { _ in
```
automactic, line 30: nit: could you write this in the style of 
```
content.onChange(of: externalURL) {
    ...
}
```

ViewModel/BrowserViewModel.swift
```diff
@@ -179,7 +180,7 @@ class BrowserViewModel: NSObject, ObservableObject,
         } else if url.isKiwixURL {
             decisionHandler(.allow)
         } else if url.scheme == "http" || url.scheme == "https" {
```
automactic, line 182: If you are introducing `isExternalLink` on `URL`, could you change this line to use that variable instead?

ViewModel/BrowserViewModel.swift
```diff
@@ -391,4 +392,34 @@ class BrowserViewModel: NSObject, ObservableObject,
...
+
+          // create new tab
+          guard let currentWindow = NSApp.keyWindow,
+                let windowController = currentWindow.windowController else { return nil }
+          windowController.newWindowForTab(nil)
+          guard let newWindow = NSApp.keyWindow, currentWindow != newWindow else { return nil }
+          currentWindow.addTabbedWindow(newWindow, ordered: .above)
+          
+          // open url in new tab
+          NotificationCenter.openURL(newUrl, inNewTab: true)
```
automactic, line 419:
```suggestion
          NotificationCenter.openURL(newUrl)
```
Context: `inNewTab` is for iOS / iPadOS only, it doesn't do anything for Mac, so I suggest removing it

automactic, line 419: This works well most of the times, but not always. The issue I think is `@Environment(\.controlActiveState) var controlActiveState` is supposed to tell us which window is key, and only the key window should respond to the open URL notification, but there sometimes can be multiple key windows.

https://github.com/kiwix/apple/blob/fa397a905f95b946e1cf6b55328771b101bb91f8/App/App_macOS.swift#L141-L143

I am not an AppKit expert, but I was under the impression only one window can become key at a time, since only one window is supposed to be able to receive keyboard event

You can reproduce the issue with (on macOS 14):
- using right click, open in new window to open stay 4 tabs
- go to the 2nd tab, right click on something, and open in new window
- multiple windows responds to the notification 
  - with the additional problem that the newly opened window having the incorrect title (haven't investigated why)

Screenshot 2023-10-07 at 10 06 27 AM

automactic (commented):
I like your solution to resolving issue around external url alerts, but I don't think the open link in new tab stuff always works reliably

Not that it's your fault or anything -- a lot of swiftui stuff are very new, and to be frank buggy. My position on "Open in new tab" is: if it doesn't result in a good and (mostly) bug free experience, let's disable it

PR #531, opened 2023-10-21
Dependency versions & C++ interop

automactic:
## Summary
Use `kiwix::getVersions` as datasource of dependency list / table, also fixing a small issue regarding Swift / C++ interop
This commit also enables SwiftLint, which is currently enabled only on `Views/Settings`

## Screenshots
</img>Screenshot 2023-10-21 at 5 46 08 PM


Kiwix.xcodeproj/project.pbxproj
```diff
@@ -540,11 +543,11 @@
 			isa = PBXGroup;
 			children = (
 				97E94B1D271EF250005B0295 /* Assets.xcassets */,
+				9713F7782AE416E5007DD9EC /* CoreKiwix.modulemap */,
```
automactic, line 546: The module map has to be named differently than what's in the install / built dir

See [this](https://developer.apple.com/forums/thread/75674) for additional context

PR #534, opened 2023-11-11
Support for libkiwix main (13)

rgaudin:
libkiwix made important changes around Library management after last release. One of them is the enforcement of using a shared pointer.

See https://github.com/kiwix/libkiwix/commit/1dc97055974a95245e17d14cf27e6376cb8ca416 which introduces the `LibraryPtr` type.

This is required in order to use built-off-main CoreKiwix.xcframework that's provided by kiwix-build.


kelson42:
Now that libkiwix 13.0.0 has been released, we can merge

PR #536, opened 2023-11-13
Fix opening new tabs on macos(#518)

BPerlakiH:
Fixing: Open link "in a new window" on macOS.
Partially based on https://github.com/kiwix/apple/pull/526

Known issues:
- the opened tabs are not restoring their content on app restart, the tabID remains nil (might need another issue/ticket).
- command+click on a link would not open it in a new tab (as I would expect it, since native safari has this capability)

### DIFF
--- a/ViewModel/BrowserViewModel.swift
+++ b/ViewModel/BrowserViewModel.swift
@@ -13,9 +13,9 @@ import WebKit
 
 import OrderedCollections
 
-class BrowserViewModel: NSObject, ObservableObject,
-                        WKNavigationDelegate, WKScriptMessageHandler, WKUIDelegate,
-                        NSFetchedResultsControllerDelegate
+final class BrowserViewModel: NSObject, ObservableObject,
+                              WKNavigationDelegate, WKScriptMessageHandler, WKUIDelegate,
+                              NSFetchedResultsControllerDelegate
 {
     static private var cache = OrderedDictionary<NSManagedObjectID, BrowserViewModel>()
     
@@ -45,14 +45,17 @@ class BrowserViewModel: NSObject, ObservableObject,
     @Published private(set) var outlineItems = [OutlineItem]()
     @Published private(set) var outlineItemTree = [OutlineItem]()
     @Published private(set) var url: URL?
-    
+    @Published var externalURL: URL?
+
     let tabID: NSManagedObjectID?
     let webView: WKWebView
     private var canGoBackObserver: NSKeyValueObservation?
     private var canGoForwardObserver: NSKeyValueObservation?
     private var titleURLObserver: AnyCancellable?
     private var bookmarkFetchedResultsController: NSFetchedResultsController<Bookmark>?
-    
+    /// A temporary placeholder for the url that should be opened in a new tab, set on macOS only
+    private static var urlForNewTab: URL?
+
     // MARK: - Lifecycle
     
     init(tabID: NSManagedObjectID? = nil) {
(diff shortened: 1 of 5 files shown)


kelson42:
@BPerlakiH Thank you.
@automactic Could you please have a look?

kelson42:
@BPerlakiH Please fix the two weaknesses you have listed so the PR is feature completed. Once done, please update PR comment and request new review.

kelson42:
@BPerlakiH Rebasing to benefit of the new CI

automactic:
@BPerlakiH Oops, I am so sorry I did not see you have this PR already opened for the external URL alert issue. So I made a PR (#539) and merged it, because I created the issue in the first place. Apologies for the merge conflict

As I have communicated with @kelson42, I won't have any capacity for PR reviews in the foreseeable future, so I'll give this one a quick look, and let him decide what to do

ViewModel/BrowserUIDelegate.swift
```diff
@@ -0,0 +1,97 @@
...
+
+                    // bookmark
+                    let bookmarkAction: UIAction = {
+                        let context = Database.viewContext
+                        let predicate = NSPredicate(format: "articleURL == %@", url as CVarArg)
+                        let request = Bookmark.fetchRequest(predicate: predicate)
+                        if let bookmarks = try? context.fetch(request), !bookmarks.isEmpty {
+                            return UIAction(title: "Remove Bookmark",
+                                            image: UIImage(systemName: "star.slash.fill")) { [weak self] _ in
+                                self?.deleteBookMark = url
```
automactic, line 81: bookmark is one word

https://www.merriam-webster.com/dictionary/bookmark

ViewModel/BrowserViewModel.swift
```diff
@@ -45,41 +44,74 @@ class BrowserViewModel: NSObject, ObservableObject,
...
+        uiDelegate.$externalURL.assign(to: \.externalURL, on: self)
+            .store(in: &cancellables)
+
+        uiDelegate.$createBookMark.sink { [weak self] url in
+            self?.createBookmark(url: url)
+        }.store(in: &cancellables)
+
+        uiDelegate.$deleteBookMark.sink { [weak self] url in
+            self?.deleteBookmark(url: url)
+        }.store(in: &cancellables)
```
automactic, line 89: I disagree with splitting one view model class into three. Because now you have the same amount of logic (yes I know it's a lot), but scattered in four places. So you might have more specialized classes, but it is harder to get to the actual logic (so arguably less readable)

And this block of code above (line 72-89), there are more code to maintain to achieve the same results, so sorry I cannot get behind that

Views/BrowserTab.swift
```diff
@@ -8,10 +8,11 @@
 
 import SwiftUI
 
-struct BrowserTab: View {
+struct BrowserTab: View, Identifiable {
```
automactic, line 11: Does conforming a view struct to identifiable actually do anything? Not sure what exactly you are aiming for, but I have certainly not seen any swiftui code written like this. For example, swiftui view is probably not gonna use this as the view's identity

Please explain, maybe I'll learn something new from you

ViewModel/BrowserViewModel.swift
kelson42: I guess this came because of the Codefactor error, maybe here we should then just add an exception.

ViewModel/BrowserUIDelegate.swift
kelson42: I guess here is meant `deleteBookmark` in place of `deleteBookMark`...

kelson42:
@automactic Thanks for having a quick look
@BPerlakiH Please fix the problem reported by @automactic and then we should be OK.

ViewModel/BrowserViewModel.swift
BPerlakiH: Reverted to one file with SwiftLint ignores.

Views/BrowserTab.swift
BPerlakiH: Sorry, it was a left over, removed now.

ViewModel/BrowserUIDelegate.swift
BPerlakiH: Thx, reverted

kelson42 (approved):
I can not judge the can and for the moment can not test it on my own. But if the points revealed by @automactic are fixed, then should be OK.

PR #544, opened 2023-11-22
Introduce Continuous Deployment

### LINKED ISSUE #492: Introduce Continous Deployment
For the moment, Kiwix for macOS/iOS is built on a developer machine and then published.

This has a number of disadvantages:
* It is really unsecure (what about is something fishy runs/happens on this computer)
* No separation of duty between development and publishing duties (as well a security issue)
* The process as a consequence is not really optimised/documented/scripted/reproducable/tracable
* Not real pressure to simplify the process of publishing
* No nightly (or similar)
* A SPOC at the developer in charge
* ...

This is a recognized bad practice of the software industry anyway and should be avoided if possible. Kiwix for macOS/iOS is almost 10 years old, had a major and great revamping last years which makes it easier to maintain. The audience is growing as well, so the impact in case of problem. Actually this repository is one of the last without CD and it's time IMO to fix this weakness.

Blocked by #159

rgaudin:
Automatically building and deploying in Github Actions for Nightlies and Releases.

Triggered by the following:

- every day at 01:32am (nightly mode)
- manualy (nightly mode)
- on release publication (release mode)

This workflow makes extensive use of secrets with no additional safe-guard, given:

- `schedule` (nightly) runs only off `main` branch.
- `workflow_dispatch` (manual) can run on any in-repo branch (but uses the workflow from `main`)
- Release publication requires push access to repo.

There are thus two *modes*: Release and Nightly (also used on manual dispatch). The mode sets the `VERSION` either to the YYYY-MM-DD date for nightly or the tag-name for the release.

It has four *targets*: `macOS dmg`, `macOS app-store`, `iOS ipa` and `iOS app-store`

- **macOS dmg**: universal notarized macOS App in a dmg uploaded to `Kiwix-$VERSION.dmg`
- **macOS app-store**: universal notarized macOS App uploaded to the App Store.
- **iOS ipa**: iOS App uploaded to `Kiwix-$VERSION.ipa`
- **iOS app-store**: iOS App uploaded to the App Store

Code Signing is *automatic* (xcode decides which one to use based on availability). We use Apple Distribution one for the app-store targets. IPA uses Apple Development and dmg uses Developer ID.

⚠️ This allows updates CI workflow to make use of the shared xcbuild action

Fixes #492


kelson42 (dismissed):
Thank you very much for this very essential piece of software

rgaudin:
you merged broken code

kelson42:
@rgaudin Maybe but I don't get why https://github.com/kiwix/apple/actions/runs/6961075788 passes then!
@tvision251 This was your PR. Do you understand why it does not pass here?

PR #546, opened 2023-11-22
External PR access to CI 

rgaudin:
Instead of accepting `pull_request` event which doesn't passes secrets, we are now using `pull_request_target` event which does.

To prevent leaking secrets, we are using an initial job that sets the `environment` dynamically based on the event.
If it's an external PR, then it deploys to `external` environment.

This `external` environment has been created, includes required secrets and mandates approval by automactic, kelson or rgaudin

Because `pull_request_target` is meant to run on `main` code and not PR's one, we also change the checkout ref.

With this setup, CI runs unattended for pushes on main or PRs from our own repo but requires validation for external ones

Note: this is based on April 2023 instructions at
https://iterative.ai/blog/testing-external-contributions-using-github-actions-secrets


kelson42 (dismissed):
At this stage I don't really understand the concept behind `pull_request_targte` but we will discuss this at the occasion.

PR #563, opened 2023-11-27
Fix download notification name

### LINKED ISSUE #557: Download of books in macOS is wicked!
The download workflow is buggy at many levels IMO:
* Once you have chosen what to download, you click on download but nothing happens, the "download" button does not disappear, your are not moved to the "Downloads" tab. You might believe that nothing has happened.
* If you reclick on the "Download" button, same behaviour, but I actually believe it start a new download for the very same book in background
* If you go to the "Download" tab, you will actually see your download which seems to be done properly... but actually the progress bar is not update, there is no refresh at all of this page looks like. You have to leave the "Downloads" tab and come back to see progresses.
* So far, there is no way to cancel/pause the download... as the corresponding button are not displayed!

Tested with latest nightly on macOS 12.

On iOS things seem to work just fine.

BPerlakiH:
Fixes #557

Both iOS and macOS apps were missing the notification values, as the wrong one was used.
It turned out that the older value (coming from Obj-c under interop objc++ mode) should be used.

### DIFF
--- a/SwiftUI/Model/Database.swift
+++ b/SwiftUI/Model/Database.swift
@@ -8,16 +8,18 @@
 
 import CoreData
 
-class Database {
+final class Database {
     static let shared = Database()
     private var notificationToken: NSObjectProtocol?
     private var token: NSPersistentHistoryToken?
     private var tokenURL = NSPersistentContainer.defaultDirectoryURL().appendingPathComponent("token.data")
     
     private init() {
+        // due to objc++ interop, only the older notification value is working for downloads:
+        // https://developer.apple.com/documentation/coredata/nspersistentstoreremotechangenotification?language=objc
+        let storeChange: NSNotification.Name = .init(rawValue: "NSPersistentStoreRemoteChangeNotification")
         notificationToken = NotificationCenter.default.addObserver(
-            forName: NSNotification.Name(rawValue: "NSPersistentStoreRemoteChange"), object: nil, queue: nil
-        ) { notification in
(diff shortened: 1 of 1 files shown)


kelson42:
CI fails :(

rgaudin:
@kelson42 CI is green ; it's a weakness that's now identified. I'll fix it soon #566

SwiftUI/Model/Database.swift
```diff
@@ -78,7 +80,7 @@ class Database {
     
     /// Save image data to zim files.
     func saveImageData(url: URL, completion: @escaping (Data) -> Void) {
```
tvision251, line 82: @BPerlakiH `Data` should be optional.

```
func saveImageData(url: URL, completion: @escaping (Data?) -> Void) {
      URLSession.shared.dataTask(with: url) { data, response, _ in
          guard let response = response as? HTTPURLResponse,
                response.statusCode == 200,
                let mimeType = response.mimeType,
                mimeType.contains("image"),
                let data = data else {
              completion(nil)
              return
          }
          let context = self.container.newBackgroundContext()
          context.perform {
              let predicate = NSPredicate(format: "faviconURL == %@", url as CVarArg)
              let request = ZimFile.fetchRequest(predicate: predicate)
              guard let zimFile = try? context.fetch(request).first else { return }
              zimFile.faviconData = data
              try? context.save()
          }
          completion(data)
      }.resume()
  }
```

tvision251 (approved):

Issue #545, opened 2023-11-22
CD and Apple Store Versioning


rgaudin:
⚠️ I have a limited understanding of all this.

macOS and iOS apps are versioned using two data:

- `MARKETING_VERSION` which is the user-visible *main version*. It's a string and currently is `3.2`
- `CURRENT_PROJECT_VERSION` which is a build-number. It's an integer or float and currently is `119`.

Both are stored in file `Kiwix.xcodeproj/project.pbxproj`

Apple Store also references the `MARKETING_VERSION` one as *App Store version*.

The Store obviously keeps an history of builds sent to it and apparently manages versions as follows:

- A human creates a *Version* in the Store (per platform). That's a string. Builds for this version will need to have a matching `MARKETING_VERSION`
- When created, it is in a *Prepare Stage*.
- There can only be one Version in Preparation, so up to two version in store: one in a complete state (like *Ready for Sale*) and one in prep
- In Prepare stages, we can upload builds for this *target version*.
- If we have no Version in Prep, we can't send builds, those are rejected.
- If we build offline and send, we must bump the build-number first or store will reject the build.
- If we build with App-Store credentials (what we now do in CD), build-number is automatically incremented.
- Each build sent creates a candidate for Test Flight.

We thus **need to decide on a process** to manage those Versions and what we want to do in CD:

- When/how/who manages the Store versions
- When/how to manage version and build-number in source code
- When to submit a build to the Store.

Note that when building for direct distribution (macOS dmg), it doesn't matter. We can build over and over the same `3.2 (119)` and notarize it.

kelson42:
@rgaudin Thank you for asking the right questions! So far @automactic has handled this process manually and because he was alone on the project, following his own unwritten rules. Considering that we have now a CD and more people involved in the development, things have to be different, written and hopefully clearer. Here is what I propose.

First, in a matter of simplicity and coherence, I want AFAP to follow here the same approach as for other repository, ie.:
* `MARKETING_VERSION` (or Apple Store version) is the public facing and the most important thing. Here what is really important is to achieve a bijection between version and changelog. If we publish a version `x.y.z` then this version is a full release and there is no way back, any change has to be made in an other version number.
* `MARKETING_VERSION` should follow [Sementic versioning](https://semver.org/)
* There is a clear process to release with changelog, release ticket, milestone, software version update and release event, ... Actually would be a good idea to make a first version of such a release ticket to list what has to be done at this time to release properly.
[...]

PR #568, opened 2023-11-28
only upload to Apple upon releases

rgaudin:
At least until #545 is figured out


kelson42:
I guess this PR explains why we had many kiwix testflight releases these last days...

Issue #520, opened 2023-09-24
Implement support of custom app for both macOS/iOS


kelson42:
A custom app is a version of Kiwix which is prepackaged with a ZIM file and branded (in iTunes) as the content itself.

We do that on Android already, but many years ago have published such an app manually already, see https://apps.apple.com/us/app/wikimed/id1281693200.

Technically, this is Kiwix which does not appear as Kiwix (the app is renamed) and without a few features:
* Everything related to the library mgmt
* In the Settings a few options related to library/download are hidden

kelson42:
@BPerlakiH #569 fixes only part of this ticket. But at this stage, not sure keeping this ticket in this form is really helpful. Could you please open the tasks left to implement - to completion of this ticket - as new issues and link them here (in a comment)?

PR #569, opened 2023-11-29
#520 custom app support, fixing all the core issues in apple code base

BPerlakiH:
Fixing: #520 
Adding support for custom apps, includes the changes only related to Kiwix main repo.


App/App_iOS.swift
```diff
@@ -50,6 +45,24 @@ struct Kiwix: App {
...
+                        LibraryOperations.scanDirectory(URL.documentDirectory)
+                        LibraryOperations.applyFileBackupSetting()
+                        LibraryOperations.applyLibraryAutoRefreshSetting()
+                        DownloadService.shared.restartHeartbeatIfNeeded()
+                    } else if let url = Brand.mainZimFileURL {
+                        LibraryOperations.open(url: url) {
+                            navigation.navigateToMostRecentTab()
+                        }
+                    } else {
+                        assertionFailure("App should support library, or should have a main zip file")
```
kelson42, line 63: The test here are not explicit enough IMO. We clearly need a primitive isCustomApp()  - or similar - which returns `true` or `false` and act accordingly. This primitive would be then used in many places. I would keep the `FeatureFlags.has...()` only for options which can be `true` or `false` within a custom app, depending how the custom app is configured. Doing so would allow to avoid the last `else` as well.

kelson42, line 63: ZIM file, not zip file

Model/Utilities/FeatureFlags.swift
```diff
@@ -1,27 +1,19 @@
-//
-//  FeatureFlags.swift
-//  Kiwix
-//
-//  Created by Chris Li on 9/25/20.
-//  Copyright © 2020 Chris Li. All rights reserved.
-//
+//  Copyright © 2023 Kiwix. All rights reserved.
```
kelson42, line 1: Please remove everything here. If there is one header to put, this is the GPLv3 header

Support/kiwix.xcconfig
```diff
@@ -0,0 +1,10 @@
+//  Copyright © 2023 Kiwix. All rights reserved.
+
+// Configuration settings file format documentation can be found at:
+// https://help.apple.com/xcode/#/dev745c5c974
+HAS_LIBRARY = true
```
kelson42, line 5: Remove these headers which are in contradction with the GPLv3 license

App/App_iOS.swift
BPerlakiH: I see, I think it can be grouped together, although the last ``else`` statement will still be present, in a way. 
That is more for the case if we configure the custom app in a wrong way, eg. if the zim file cannot be read (eg. wrong filename or permission issues).

Model/Utilities/FeatureFlags.swift
BPerlakiH: opened a dedicated ticket for these here:
https://github.com/kiwix/apple/issues/575

App/App_iOS.swift
BPerlakiH: Thanks, updated it.

Support/kiwix.xcconfig
BPerlakiH: also see dedicated ticket for the headers are: https://github.com/kiwix/apple/issues/575

App/App_iOS.swift
BPerlakiH: I managed to simplify this config for the Kiwix app, now it only contains the bare minimum, in the info.plist file and in the xccoonfig. Additionally I managed to cut down on this if else, as this validation is now managed in one place.

BPerlakiH:
Moved these changes in a more clean form to: https://github.com/kiwix/apple/pull/576

BPerlakiH:
Moved these changes here: https://github.com/kiwix/apple/pull/576

PR #574, opened 2023-12-01
Update UI srings for better localisation

### LINKED ISSUE #571: New lines are not supported by Translatewiki
See https://phabricator.wikimedia.org/T351949#9370182

tvision251:
Fixes #571


App/SplitViewController.swift
```diff
@@ -128,7 +128,7 @@ class SplitViewController: UISplitViewController {
             let controller = UIHostingController(rootView: Settings())
             setViewController(UINavigationController(rootViewController: controller), for: .secondary)
         default:
-            let controller = UIHostingController(rootView: Text("Not yet implemented".localized))
```
BPerlakiH, line 131: I am not sure if this part needs localisation at all... The user should never see this case.

SwiftUI/Model/Enum.swift
```diff
@@ -135,9 +160,9 @@ enum LibraryLanguageSortingMode: String, CaseIterable, Identifiable, Defaults.Se
     var name: String {
         switch self {
         case .alphabetically:
-            return "A-Z"
+            return "title-a-z".localized
         case .byCounts:
-            return "By Count"
+            return "title_by_country".localized
```
BPerlakiH, line 165:
```suggestion
            return "title_by_count".localized
```

Support/en.lproj/Localizable.strings
```diff
@@ -6,198 +6,193 @@
...
+"title-downloads" = "Downloads";
+"title-new" = "New";
+"title-map" = "Map";
+"title-tab-new" = "New Tab";
+"title-disabled" = "Disabled";
+"title-first-paragraph" = "First Paragraph";
+"title-first-sentence" = "First Sentence";
+"title-matches" = "Matches";
+"title-a-z" = "A-Z";
+"title_by_country" = "By Count";
```
BPerlakiH, line 198:
```suggestion
"title_by_count" = "By Count";
```

Views/BrowserTab.swift
```diff
@@ -22,7 +22,7 @@ struct BrowserTab: View {
                     Button {
                         NotificationCenter.toggleSidebar()
                     } label: {
-                        Label("Show Sidebar".localized, systemImage: "sidebar.left")
+                        Label("zim-file-show-slider".localized, systemImage: "sidebar.left")
```
BPerlakiH, line 25:
```suggestion
                        Label("zim-file-show-sidebar".localized, systemImage: "sidebar.left")
```

Views/Bookmarks.swift
```diff
@@ -53,7 +53,7 @@ struct Bookmarks: View {
                     Button {
                         NotificationCenter.toggleSidebar()
                     } label: {
-                        Label("Show Sidebar".localized, systemImage: "sidebar.left")
+                        Label("zim-file-show-slider".localized, systemImage: "sidebar.left")
```
BPerlakiH, line 56:
```suggestion
                        Label("zim-file-show-sidebar".localized, systemImage: "sidebar.left")
```

App/App_macOS.swift
```diff
@@ -108,7 +108,7 @@ struct RootView: View {
                     responder.tryToPerform(#selector(NSSplitViewController.toggleSidebar(_:)), with: nil)
                 } label: {
                     Image(systemName: "sidebar.leading")
-                }.help("Show sidebar".localized)
+                }.help("zim-file-show-slider".localized)
```
BPerlakiH, line 111:
```suggestion
                }.help("zim-file-show-sidebar".localized)
```

Support/en.lproj/Localizable.strings
```diff
@@ -6,198 +6,193 @@
...
+"refresh-time-never" = "Never";
+"message-nothing" = "There is nothing to see";
+"zim-file-missing" = "Zim file is missing.";
+"zim-articles" = "articles";
+"library-select-zim-file" = "Select a zim file to see detail";
+"library-main-page" = "Main Page";
+"library-copy-url" = "Copy URL";
+"library-copy-id" = "Copy ID";
+"zim-file-no-opened" = "No opened zim file";
+"zim-file-show-slider" = "Show Sidebar";
```
BPerlakiH, line 77:
```suggestion
"zim-file-show-sidebar" = "Show Sidebar";
```

BPerlakiH (commented):
I was wondering if we really gain by naming these keys ourselfs, such as "title-something-something" (eg. title-build)?
Maybe a simpler approach would be to convert the actual value to be the key, eg. with lowercased, underscored version:
build: "Build"
our_website: "Our Website"
gnu_general_public_license_v3: "GNU General Public License v3"
maybe with some limit on the length of the key, eg max: 120 chars.
This way it becomes more of a dictionary, that can be translated to other languages.

The downside of the current approach is that it needs an additional mental mapping from key to value or the other-way around, and makes the code harder to read, eg:
```
Label("button-article-random-title".
```
instead of:
```
Label("random_page"

most of the keys became "titles".

With the current approach it looks more like a CMS (content management solution), which it really is not, these are static values, which will be baked into the app.

Naming things is usually hard, I think in this case we could save us the effort of not doing so.

BPerlakiH:
I just wanted to share another approach to this. Anyways, I am happy to keep it as it is now.
@tvision251  Maybe the "common" prefix mentioned in the 2nd article, could be useful for the re-usable cases.

kelson42:
@tvision251 Status?

BPerlakiH (approved):

Issue #575, opened 2023-12-02
Replace file License headers to match GLPv3


BPerlakiH:
In an attempt to unify the header of each source file, including new ones, we want to apply a common template, and replace the existing ones across the code base.

BPerlakiH:
Can we define here how the header should look like, to be GLPv3 compliant?
Then I can replace them across the project, and set it in a template, so new files are created with that from then on.

kelson42:
Yes I will, but not a prio for the moment.

PR #576, opened 2023-12-03
#520 custom app support, fixing all the core issues in apple code baseCustom app support

BPerlakiH:
Fixing: https://github.com/kiwix/apple/issues/520
Adding support for custom apps, includes the changes only related to Kiwix main repo.


App/App_iOS.swift
```diff
@@ -1,10 +1,4 @@
-//
```
kelson42, line 1: Copyright is not owned by Kiwix. Headers should be changed in a dedicated PR, see ticket #575. Lets keep this PR for the custom apps.

Kiwix.xcodeproj/project.pbxproj
```diff
@@ -655,7 +664,7 @@
 				};
 			};
 			buildConfigurationList = 97A2AB831C1B80FF00052E74 /* Build configuration list for PBXProject "Kiwix" */;
-			compatibilityVersion = "Xcode 13.0";
+			compatibilityVersion = "Xcode 15.0";
```
kelson42, line 667: This is not something we can changed just like this usually. @rgaudin Is that OK?

BPerlakiH: No problem, reverted to version 13.

kelson42 (approved):

Issue #573, opened 2023-12-01
Hide library-related menu entries on macOS


rgaudin:
In custom-app, those entries should not be visible.

<img width="914" alt="Screenshot_2023-12-01_at_11_22_11" src="https://github.com/kiwix/apple/assets/57929/6cad0e38-62a6-4150-be15-b351c8d384d4">

PR #577, opened 2023-12-03
Hide library menus on macOS

BPerlakiH:
Fixing: [#573](https://github.com/kiwix/apple/issues/573)

Hide library menus on macOS.


BPerlakiH:
No longer needed.

PR #587, opened 2023-12-08
Remove unused file

BPerlakiH:
Remove file that is not in use anymore (not linked to any target).


rgaudin:
There might be a reason @automactic left it on the repo. Can be restored from git anyway though.

PR #590, opened 2023-12-09
Remove unused file

BPerlakiH:
Removing unused test file:
Screenshot 2023-12-09 at 21 25 44


kelson42 (approved):
Not really sure why this is unused honestly and therefore not sure this is a good idea to remove it.

PR #595, opened 2023-12-17
Enforce localization for custom apps

### LINKED ISSUE #594: Specify/force language for custom apps
Kiwix custom apps should have the ability to force UI language to a specific local.

BPerlakiH:
Fixes #594 

We want to enforce the localization of the apps.
There are a couple of limitations in this:
- Apple is choosing which language should be selected by default on it's own.
- It is possible to override the UserDefaults "AppleLanguages", but swift UI cannot do this early enough (main function is provided by Apple and we don't know how to re-implement it in a safe way to override), 
- The result is that t**he full enforcement of the language only works on the second launch of the app** (not fresh install)
- On fresh install we can override the bundle we use to search for localised strings, which works OK, except the search bars' cancel button, which is currently using the system provided button.

Simulator Screenshot - iPad mini (6th generation) - 2023-12-17 at 14 53 08
Screenshot 2023-12-17 at 14 52 40

Since this is a first launch only issue, and not really visible without using the search functionality, I think we can live with this.

### DIFF
--- a/App/App_iOS.swift
+++ b/App/App_iOS.swift
@@ -17,8 +17,11 @@ struct Kiwix: App {
     @UIApplicationDelegateAdaptor(AppDelegate.self) private var appDelegate
     
     private let fileMonitor: DirectoryMonitor
-    
+
     init() {
+        if let enforcedLanguage: String = Config.value(for: .enforcedLanguage) {
+            DefaultLanguages.enforce(language: enforcedLanguage)
+        }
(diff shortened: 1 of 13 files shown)


kelson42:
@BPerlakiH How do apps which are only available in German? I guess, first launch can not be in "default language" if it does not exist.

kelson42 (approved):
LGTM from a user perspective

kelson42:
@tvision251 Please make the code review

PR #596, opened 2023-12-18

translatewiki:
Translation updates


BPerlakiH:
So far the PR itself looks good.
Overall there are 3 issues found with the localisation:
1) if a value is not translated it will fall back to the key, instead of English.
2) the new languages are not added automatically to the project, for this I will create a separate PR to handle it via XCodeGen.
3) the qqq.proj which we should not include in the project, here there are 2 options:
 - we can .gitignore it, and it won't be under source control
 - we can exclude it via XcodeGen, so it won't include it in the project itself, but will be source controlled

kelson42:
@BPerlakiH We can not have fallback to english?

rgaudin:
> @BPerlakiH We can not have fallback to english?

If I understood correctly, being able to fallback to English means including EN language thus having the app displayed in EN for all those not having their system in German

BPerlakiH:
> @BPerlakiH We can not have fallback to english? Xcodegen shoukd resolve the other points.

Actual we can fallback with the changes added here: https://github.com/kiwix/apple/pull/597
That PR also addresses the other concerns.
**One questions remains, do we need the qqq.proj file in git?** 
We can either .gitignore it, or keep it in git and ignore it on the project level (so it won't make it into our builds).

BPerlakiH:
I've added the relevant localisation changes to this PR:
- make sure we are falling back to English if a term / key has no translation yet.
- fixing the icons on the article page, as the name of the image to be displayed was also using the translated values, which caused the icons to disappear in other (than English) languages.

BPerlakiH:
This PR will add the new language files to Git, but won't add it to the project itself.
- We can do that manually (which I think we agreed on not to do)
- We can do it automatically using Xcode gen, via this PR: https://github.com/kiwix/apple/pull/597

BPerlakiH (approved):

PR #597, opened 2023-12-20
Adding XcodeGen

BPerlakiH:
We want to replace the manual project file with an auto-generated one.
This gives us the flexibility to include new files in the project automatically, eg. new translations.

In order to keep it in sync, it requires to re-generate the project with each file added or removed.
This is currently achieved via shared git-hooks, with the help of pre-commit.

Readme file updated with the necessary 1 extra step for new developers to start with the project.


project.yml
```diff
@@ -0,0 +1,143 @@
...
+        INFOPLIST_KEY_CFBundleDisplayName: Kiwix
+    type: application
+    supportedDestinations: [iOS, macOS]
+    # info:
+      # path: Support/Info.plist
+      # CFBundleName: Kiwix
+    sources: 
+      - path: App
+      - path: Support/Kiwix-Bridging-Header.h
+        destinationFilters: [iOS]
```
rgaudin, line 76: Why would this not be required for macOS?

README.md
```diff
@@ -20,6 +20,16 @@ downloadable for older versions of macOS and iOS on the Mac App Store.
 
 Kiwix developers usually work with latest macOS and Xcode. Check our [Continuous Integration Workflow](https://github.com/kiwix/apple/blob/main/.github/workflows/ci.yml) to find out which XCode version we use on Github Actions.
 
+After checkout please run:
```
rgaudin, line 23: I thought we wanted to ease onboarding (cf. the move to a swift package) ; this goes the other way.
I'd be reading the README so I'm OK with this but is @kelson42? Because we discussed otherwise:

- having the xcproj in the repo
- having it in .gitignore (so it's not pushed accidentaly)
- in the CI (and/or pre-commit), if `project.yml` is modified, ensure the project is added (with --force) and pushed otherwise fail.

Pros:

- does not require any initial setup from developers
- prevents accidental un-sync

Cons:

- no `git diff Kiwix.xcproject` (but that's the same with what's proposed here ; consequence of xcodegen)
- potential frustration of newcomer not reading readme and wanting to push change to xcproj before realizing it's ignored
- Initial CI failures in PRs when modifying the template and not force-pushing the gen'd xcproj

rgaudin (commented):
I don't know if this is because it targets `main` which would lack ongoing progress but this PR includes a lot of unrelated stuff.

Looks good but I think we should sort this method thing first: should we include the proj in the repo or not?

project.yml
BPerlakiH: This had to be removed for macOS in the original project file as well, in order to make it compile, please see it here:
https://github.com/kiwix/apple/blob/37532b303f7b6fff80a2234745164e2cd1470b3b/Kiwix.xcodeproj/project.pbxproj#L106C23-L106C23
at the end of line 106 we have: "platformFilter = ios; };"

README.md
BPerlakiH: I don't have a strong opinion on this one.
Now I've added back the project files generated by Xcode-gen, maybe as a step 2 we can try to figure out how to keep it in sync.
Any swift file / asset file added to the repo would change the project file, therefore it needs to be re-generated.
Maybe we could apply special GitHub rules who can commit to that file, so on a successful build it will be updated from the CI side, and any change pushed to it will be ignored / overwritten.

BPerlakiH:
> I don't know if this is because it targets `main` which would lack ongoing progress but this PR includes a lot of unrelated stuff.
I think you are right, it's a bit confusing with the other changes in the same place.
I've moved the translation related changes to the other PR: https://github.com/kiwix/apple/pull/596
and left here only the project changes.

project.yml
rgaudin: OK

README.md
kelson42: > I thought we wanted to ease onboarding (cf. the move to a swift package) ; this goes the other way.
> I'd be reading the README so I'm OK with this but is @kelson42?

I confirm

kelson42:
What is the status here? How this PR articulates with #606? Whihc ticket does it fix?

BPerlakiH:
> What is the status here? How this PR articulates with #606? Whihc ticket does it fix?

It has been superseded by https://github.com/kiwix/apple/pull/606, therefore I am closing this one, to make things cleaner.

PR #599, opened 2023-12-21
Generate and commit Xcode project files in CI

### LINKED ISSUE #607: Improve handling of (code generated) XCode project files
# The origin of the problem

Xcode needs a so called project folder in order to compile an app.
This is a collection of XMLs that is referencing other files, has settings per target and per scheme, and other general compile options.

In order to dynamically work with similar settings to the main repo, but building custom apps, we need to change some of these settings, and might not need to include all the files in the project (eg. selecting only a single language file to be included), or add special files, such as a custom zim file.

Editing the XML file doesn't seem like a viable option here, the most important file is over 1500 lines long.

For these reasons, we wanted to use XCodeGen, which based on a yml project file and the files available in the repository, can create the very same Xcode project directory. It has templating, scheme and other options as well, which is very handy for our goals.

# The current problem

What can cause a change in the project directory? It can be a manual setting change in Xcode, but it can also be a file change, eg. adding a new file / renaming a file or deleting it (outside of the Xcode project folder that is, eg: adding a new Model or Utility class).

We wanted to leave the Xcode project folder in git, so that a new developer can clone the project and start using it. This raises a question how can we make sure that the project file is up to date?

**We wanted to add the project folder to .gitignore. This is not working as we expected.** The project folder is already "tracked" by git, so every change to it needs to be committed. Adding the project folder to .gitignore, has no effect, it is still "tracked" by git. From the [git documentation](https://git-scm.com/docs/gitignore):
> A gitignore file specifies intentionally untracked files that Git should ignore. Files already tracked by Git are not affected;

This means that we are left we 2 options. Either leaving the project file in git, and make sure it's always up to date, or removing it from git.

# 1) Alternative approach keep tracking the project folder, but regenerate on CI side:
In order to make sure that the project folder is up to date, it needs to be re-generated on the CI side, either at the level of building the Pull Request (a successful build means it's a valid project folder) or at the level of the main branch, when merging PRs.
This means we will apply either **automatic commits** on top of pull requests or as part of merging pull requests.

# 2) Alternative approach, remove the project folder from git:
Keeping in mind that we want to make sure that onboarding developers is  easy, we can combine the first steps of getting started with the project into **a single command**, that will both install the necessary tools, automate the project generation and synchronisation for the developers, and downloads CoreKiwix.xcframework at the same time. A single command that needs to be run only once after cloning the project.

BPerlakiH:
Fixes #607 

Adding a post build step to commit back the xcodeproject files via CI.


kelson42:
@BPerlakiH I don't know what exactly which ticket you try to fix, but all commits should be done by humans. You don't need this MO.

BPerlakiH:
Please see the full description how this attempt might (or might not) fit into our goals here:
https://github.com/kiwix/apple/pull/606

kelson42:
@BPerlakiH I require clearly defined issue(s) for each PR. This is not the case neither here or for #606. This is one of the reason why I don't understand you. Please follow processes accurately.

@rgaudin Maybe you better understand than me what is going on here (beside "wild" PRs. But to me there seems to be behind this PR a misunderstanding how works git.

rgaudin:
Can we close this PR? It _fixes_ an issue that is closed and we don't have the project in the repo anymore.

BPerlakiH:
I am happy to close this.

PR #600, opened 2023-12-22
Adding fastlane to support beta builds to TestFlight

BPerlakiH:
Initial setup of Fastlane to support beta builds directly to TestFlight.


rgaudin:
This is still in draft but looks good to me ; except that it's using your account to connect to ITC. Can fastlane use the API token and password we already have configured in secrets and in CD?

BPerlakiH:
Closing this as we had little success with direct fastlane usage.

Issue #603, opened 2023-12-25
Proper handling of UI languages


kelson42:
Kiwix is now able to handle many languages and the syncrhonisation with Translatewiki has been achieved. But it is still important that the "correct" language is chosen to display the UI:
* Custom apps
  * `enforced_lang` lang is specified in the `info.json`
    * UI language should not be customizable (in the OS - but app related - settings)
    * UI language should be the one specified in `info.json`
  * Otherwise works like the Kiwix app
* Kiwix app
  * UI languages should be configurable (in the OS - but app related - settings) on iOS (but nothing for the moment on macOS, see #604)
  * Default language should be the one of the OS

This is valid for iOS and macOS.

kelson42:
@BPerlakiH Do you confirm this is clear? This makes sense? Does it already works like this?

rgaudin:
there is no such thing as app préf in OS for macOS. It's only iOS

kelson42:
> there is no such thing as app préf in OS for macOS. It's only iOS

Then, no way to configure the UI language for the moment. I have open [an other ticket](https://github.com/kiwix/apple/issues/604) for this enhancement (not a priority for the moment).

BPerlakiH:
Thank you @kelson42 it makes sense. 
This really means we can should include only the enforced language in the custom app.
I will reflect these changes in the latest release.

BPerlakiH:
I've created the releases for DWDS app 2023.12.3 (build 133) to reflect these changes, and created a PR to revert the unnecessary changes in the main repo: https://github.com/kiwix/apple/pull/605

The project level (automatic) handling of the enforced language will be included on the kiwix-apple-custom repo side.

PR #605, opened 2023-12-29
Remove default/enforced language code solution

BPerlakiH:
## The problem as of  #603 

Our former assumption was that if a language is enforced for a custom app, we need to force its selection from the available list of languages, to be the default one.

This assumption is no longer true.

What we want to do instead, is that if a language is enforced for a custom app, we only include this single language, and no other languages can be selected for this app. This means that we can handle this change on the project level, and no code change is required.

## Solution 

Remove the former changes in code, as they are no longer needed.


kelson42:
@BPerlakiH Thank you. If it full fixes #603,lease write "Fixes #603" in the PR description. As PR title, please only write a synthetix description of what the PR does (so no related issue number).

BPerlakiH:
Thank you, I have updated the PR title, pushed a fix for the CI build.
This PR is not directly solving the issue, the solution comes from the project file settings of the custom app (which is not part of the Kiwix/main). This PR is removing the unnecessary code that was introduced earlier for handling the language settings, which we no longer need.

kelson42:
@BPerlakiH If i get it right, if we merge only this we will create actually a regression. So in which PR comes the full fix?

BPerlakiH:
@kelson42 it won't cause a regression, as the solution to the problem is to include only the German language file in the DWDS app settings, which is outside of this repository. It's OK to merge this.

kelson42 (approved):

PR #606, opened 2023-12-29
XcodeGen integration

BPerlakiH:
# XcodeGen integration

Fixes: #607 
https://github.com/kiwix/kiwix-apple-custom/pull/9 - and https://github.com/kiwix/kiwix-apple-custom/issues/1 depend on this

## The problem
We want to dynamically handle the files that goes into the Xcode project, for example the translation files.

## The solution is XcodeGen
We would like to have all the benefits of XcodeGen, namely:
- ✅ Generate projects on demand and remove your `.xcodeproj` from git, which means **no more merge conflicts**!
- ✅ Groups and files in Xcode are always **synced** to your directories on disk
- ✅ Easy **configuration** of projects which is human readable and git friendly
- ✅ Easily copy and paste **files and directories** without having to edit anything in Xcode
- ✅ Share build settings across multiple targets with **build setting groups**
- ✅ Automatically generate Schemes for **different environments** like test and production
- ✅ Easily **create new projects** with complicated setups on demand without messing around with Xcode
- ✅ Generate from anywhere including on **CI**
- ✅ Distribute your spec amongst multiple files for easy **sharing** and overriding
- ✅ Easily create **multi-platform** frameworks

## The side effects
Since we want to avoid merge conflicts in the .xcodeproj files, the easiest (and the recommended way) is to remove it from git, meaning we stop tracking the changes to this folder.
This is in contrast with our aim to make developer onboarding easy. There are 2 ways to solve it:
### 1) keep tracking the .xcodeproj file
We need to deal with the conflicts manually. 

In order to keep it up to date, we need to regenerate the project file, ideally on the CI side, so that new users to the repository should get the latest, and up to date version.
Our former attempt to simple add it to .gitignore is not working as expected. From the [git documentation](https://git-scm.com/docs/gitignore):
> A gitignore file specifies intentionally untracked files that Git should ignore. Files already tracked by Git are not affected;

What we need in this case is, ignoring any incoming changes via pull requests, and re-generate the project file on CI side. Which means we either need to apply automatic commits to each PR as we build it, or we need to apply an automatic commit on the main branch after merging a PR. (see attempts of this here: https://github.com/kiwix/apple/pull/599)

### 2) we do not track the .xcodeproj file
The side effect of this, is that after a git clone the new developer won't have access to this file. On the same note a new developer (without reading the Readme) won't have the CoreKiwix.xcframework either.

To automate the xcodeproject generation locally (on the developer machine), it is possible to use pre-commit hooks, contained in this PR. Additionally to keep our main goal of easying the developer onboarding, the CoreKiwix.xcframework download and all the necessary dependencies, like xcodegen, precommit, is now proposed to be included in a single command: 
`brew bundle`.

This way we can reduce the Readme and the initial steps to a single command.


kelson42:
> Since we want to avoid merge conflicts in the .xcodeproj files, the easiest (and the recommended way) is to remove it from git, meaning we stop tracking the changes to this folder.

This sentence seems wrong by many things:
* merge conflicts happen only if base branch get out of sync with feature branch. It has nothing to do with what is going on here.
* stop tracking a file is not the same at all than removing from git.

BPerlakiH:
@kelson42 I have tried to sum it up in this new issue here: https://github.com/kiwix/apple/issues/607,
if you have any questions, please let me know.

.gitignore
```diff
@@ -70,3 +70,6 @@ fastlane/test_output
 .DS_Store
 xcuserdata
 Libraries
+
+# ignore xcode project files, they are now generated by XcodeGen
```
kelson42, line 74: Yes, but they should be tracked by git

.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
+# docs: https://pre-commit.com
+default_install_hook_types: [post-merge, post-checkout, post-rewrite]
+repos:
+  - repo: local
+    hooks:
+      - id: xcodegen
+        name: Generate project files for XCode
```
kelson42, line 7: Xcode, not XCode

project.yml
```diff
@@ -0,0 +1,143 @@
+name: Kiwix
+options:
+  xcodeVersion: 13.0
+  deploymentTarget:
+      iOS: 15.0
```
kelson42, line 5: Here we should put a comment explaining what is the rule to apply regarding supported versions… or maybe just a link to the readme file.

project.yml
```diff
@@ -0,0 +1,143 @@
...
+      macOS: 12.0
+  generateEmptyDirectories: true
+  useTabs: false
+  groupSortPosition: top
+
+settings:
+  base:
+    DEVELOPMENT_TEAM: L7HWM3SP3L
```
kelson42, line 13: This should not be a secret?

project.yml
```diff
@@ -0,0 +1,143 @@
...
+    sources: 
+      - path: App
+      - path: Support/Kiwix-Bridging-Header.h
+        destinationFilters: [iOS]
+      - path: Model
+      - path: Support
+        excludes: 
+          - "Launch.storyboard"
+          - "**/*.zim"
+          - "DWDS.xcassets"
```
kelson42, line 82: We should not have custom app specific things in this file IMO

kelson42 (changes requested):
I have made a new path and have found a few details. The README should be uüdated to explain everything around Xcodegen:
* principle of work
* How to generate new project files
* related git rules

project.yml
BPerlakiH: It is not considered a secret, please see: https://stackoverflow.com/questions/44417171/should-i-keep-the-apple-developer-team-id-secret

BPerlakiH: We were also including this value in the original pbxproject file (several times actually).

project.yml
BPerlakiH: updated

.pre-commit-config.yaml
BPerlakiH: fixed

project.yml
BPerlakiH: removed

BPerlakiH:
> I have made a new pass and have found a few details. The README should be uüdated to explain everything around Xcodegen:
> 
> * principle of work
> * How to generate new project files
> * related git rules

I've added a section for this in our Readme.

kelson42:
@BPerlakiH Please rebase (no merge!) on main branch.

kelson42 (approved):
LGTM

kelson42:
Please explicitly write which tickets are fixed with this PR:
* https://github.com/kiwix/kiwix-apple-custom/issues/1 ?
* #607 ?

BPerlakiH:
@kelson42 I've rebased this PR, and updated the issue links in the main description.

kelson42:
@rgaudin I merge to ensure @BPerlakiH can move forward in confortable condition, but would be glad to get your feedback anyway.

rgaudin:
- I find odd that the libkiwix version is defined in the Brewfile. libkiwix is our top dependency and exposing it solely in a URL in the Brewfile feels hackish.
- This PR doesn't update the CI and CD workflows which now relies on this mechanism. We thus end up with a `LIBKIWIX-VERSION` environ on those files that makes it look like it controls the version used while it doesn't anymore.
- README says that the Brewfile will install **latest** CoreKiwix. That's not true. It will install the version specified in the brewfile. We have a URL for the latest version but we're not using it (on purpose) so this sentence is misleading.
- `DEVELOPMENT_TEAM` is not a secret value but it's part of the several values associated with signing. In all repos we need them, we store it as a secret for convenience: when updating signing params, everything's in the same place. Given it doesn't change when certificates are renewed, we should remove it from the secrets… but you need to remove it from the places that are looking for it (CI, CD).

kelson42:
> * I find odd that the libkiwix version is defined in the Brewfile. libkiwix is our top dependency and exposing it solely in a URL in the Brewfile feels hackish.

@rgaudin I think I understand, the problem is that actually libkiwix is not a Brew dependence. What would be a better alternative to your opinion?

>     * This PR doesn't update the CI and CD workflows which now relies on this mechanism. We thus end up with a `LIBKIWIX-VERSION` environ on those files that makes it look like it controls the version used while it doesn't anymore.

@BPerlakiH I kind of assumed you didn't do that because everything will chnage a lot with fastlane. What is the explanation of this lack of workflow update? When is fastlane introduction expected to work?

>     * README says that the Brewfile will install **latest** CoreKiwix. That's not true. It will install the version specified in the brewfile. We have a URL for the latest version but we're not using it (on purpose) so this sentence is misleading.

Please fix, we should anyway not use unspecific dependency version in both CI/CD.

>     * `DEVELOPMENT_TEAM` is not a secret value but it's part of the several values associated with signing. In all repos we need them, we store it as a secret for convenience: when updating signing params, everything's in the same place. Given it doesn't change when certificates are renewed, we should remove it from the secrets… but you need to remove it from the places that are looking for it (CI, CD).

I prefer to have this as a secret (for convenience), but if integration of this secret is a nightmare, then lets be pragmatical... anyway things have to be fixed to be coherent.

Issue #609, opened 2024-01-01
Polish Xcode integration


kelson42:
- I find odd that the libkiwix version is defined in the Brewfile. libkiwix is our top dependency and exposing it solely in a URL in the Brewfile feels hackish.
- This PR doesn't update the CI and CD workflows which now relies on this mechanism. We thus end up with a `LIBKIWIX-VERSION` environ on those files that makes it look like it controls the version used while it doesn't anymore.
- README says that the Brewfile will install **latest** CoreKiwix. That's not true. It will install the version specified in the brewfile. We have a URL for the latest version but we're not using it (on purpose) so this sentence is misleading.
- `DEVELOPMENT_TEAM` is not a secret value but it's part of the several values associated with signing. In all repos we need them, we store it as a secret for convenience: when updating signing params, everything's in the same place. Given it doesn't change when certificates are renewed, we should remove it from the secrets… but you need to remove it from the places that are looking for it (CI, CD).

Reported at https://github.com/kiwix/apple/pull/606#issuecomment-1873280722

kelson42:
> * I find odd that the libkiwix version is defined in the Brewfile. libkiwix is our top dependency and exposing it solely in a URL in the Brewfile feels hackish.

@rgaudin What would be an better altenative approach?
 
>     * This PR doesn't update the CI and CD workflows which now relies on this mechanism. We thus end up with a `LIBKIWIX-VERSION` environ on those files that makes it look like it controls the version used while it doesn't anymore.

@BPerlakiH Why?

>     * README says that the Brewfile will install **latest** CoreKiwix. That's not true. It will install the version specified in the brewfile. We have a URL for the latest version but we're not using it (on purpose) so this sentence is misleading.

Dependencies version should be explicit anyway. Readme should be fixed.
 
>     * `DEVELOPMENT_TEAM` is not a secret value but it's part of the several values associated with signing. In all repos we need them, we store it as a secret for convenience: when updating signing params, everything's in the same place. Given it doesn't change when certificates are renewed, we should remove it from the secrets… but you need to remove it from the places that are looking for it (CI, CD).

I prefer the secret for convenience, but if this approach is proven to be combersome, then lets hardcode it everywhere. In any case, this has to be fixed.

@BPerlakiH Please treat ticket in priority.

BPerlakiH:
I've addressed these issues here: https://github.com/kiwix/apple/pull/610

The Development Team value. Yes, I think that can be removed from secrets, I do not think it is a priority atm.

[...]

PR #610, opened 2024-01-01
post XcodeGen: Remove libkiwix version from gitactions, update readme

BPerlakiH:
Fixing #609 

- Update the CI and CD workflows which were relying on a specific libkiwix version.
- Update the README to reflect that the Brewfile is setting the version of libkiwix to be used.


rgaudin (approved):
LGTM 👍

PR #612, opened 2024-01-02
Use full urls for websites

BPerlakiH:
Part of fixing: https://github.com/kiwix/kiwix-apple-custom/issues/10

We want to replace hard coded https://, to be able to use a full url in our custom info.jsons.


rgaudin (approved):
LGTM

PR #614, opened 2024-01-07
Fix first time build of the project

### LINKED ISSUE #613: Fix first time build
# The problem
If the project is checked out fresh (which is the case every single time on CI builds), it cannot be built right away.
Only the second time we build it, it will succeed.

# Cause
The module maps are copied into the CoreKiwix.xcframework as a building step.
We do this for every single build, although we really need to do it only once.
The real problem with this is that this step is carried out slightly late in the process, therefore the first build will fail, while subsequent builds will succeed.

# Solution

Move the copy module maps commands as a post step to downloading the CoreKiwix.xcframework, and remove it from the build steps.
This way the first build will succeed (and all the subsequent ones as well).

BPerlakiH:
Fixes #613

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -8,6 +8,6 @@ at_exit do
     # Copy Clang module map to xcframework for Swift C++ Interoperability 
     system "cp Support/CoreKiwix.modulemap CoreKiwix.xcframework/ios-arm64/Headers/module.modulemap"
     system "cp Support/CoreKiwix.modulemap CoreKiwix.xcframework/ios-arm64_x86_64-simulator/Headers/module.modulemap"
-    system "cp Support/CoreKiwix.modulemap"
+    system "cp Support/CoreKiwix.modulemap CoreKiwix.xcframework/macos-arm64_x86_64/Headers/module.modulemap"
     system "xcodegen"
 end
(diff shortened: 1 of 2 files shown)


kelson42 (approved):
LGTM

rgaudin:
@kelson42, I was asked to review this as well. 10h, during night time is too short. Especially considering there is no urgency regarding this ticket.

This PR lacks the removal of the hack/workaround in the workflows. @BPerlakiH please make another PR for it.

@kelson42 we should discuss this today. I think this should be in the xcframework. If we agree on this, we should discuss whether to put it in kiwix-build or libkiwix.

BPerlakiH:
@rgaudin I understand your point. My point of view was a bit the other way around: 
As I tried to add the fastlane integration to do the building of the apps, I did not wanted to introduce a similar work around (with retries) to make the build green.
Since we are about to change the CI workflow, the retry work around will be cleaned up that way.

BPerlakiH:
I do like the idea of putting it into the XCFramework, that would simplify this even more.

rgaudin:
> Since we are about to change the CI workflow, the retry work around will be cleaned up that way.

OK 👍

rgaudin:
> I do like the idea of putting it into the XCFramework, that would simplify this even more.

We've agreed on this with @mgautierfr and @kelson42 ; I'll make a PR and we'll have it in XCF for the next version

Issue #615, opened 2024-01-10
Remove xcconfig from main project and custom projects


BPerlakiH:
It turned out we can simplify the app configuration by relying only on .plist file values, therefore we can remove the xcconfig files.

BPerlakiH:
Currently the xcconfig file was containing the final value for APP_STORE_ID, and it was used in custom apps for other values as well.
Each target (the Kiwix app, and each custom app) had its own Info.plist file and xcconfig.
It turned out that we can simplify this process with inlining the values required for both Kiwix app and custom apps into the info plist file, and we can remove all xcconfig files, as this way they won't be needed.
This simplifies the whole process.

BPerlakiH:
In order to achieve this both the kiwix/apple and the kiwix/kiwix-apple-custom repositories needs to be updated.

BPerlakiH:
Implemented

PR #616, opened 2024-01-10
Remove xcconfig file, inline App Store ID

BPerlakiH:
Fixing #615

### DIFF
--- a/Support/Info.plist
+++ b/Support/Info.plist
@@ -3,7 +3,7 @@
 <plist version="1.0">
 <dict>
 	<key>APP_STORE_ID</key>
-	<string>$(APP_STORE_ID)</string>
+	<string>id997079563</string>
 	<key>BGTaskSchedulerPermittedIdentifiers</key>
 	<array>
 		<string>org.kiwix.library_refresh</string>
--- a/Support/kiwix.xcconfig
+++ b/Support/kiwix.xcconfig
@@ -1,5 +1,0 @@
-//  Copyright © 2023 Kiwix.
-
-// Configuration settings file format documentation can be found at:
-// https://help.apple.com/xcode/#/dev745c5c974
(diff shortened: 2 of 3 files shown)


Support/Info.plist
```diff
@@ -3,7 +3,7 @@
 <plist version="1.0">
 <dict>
 	<key>APP_STORE_ID</key>
-	<string>$(APP_STORE_ID)</string>
+	<string>id997079563</string>
```
rgaudin, line 6: How does this connects to the appstore ID ? where is it coming from ?

BPerlakiH: We already had this value set, it was just not so direct.
We had: Info.plist containing key: APP_STORE_ID  -- pointing to --> ${APP_STORE_ID} which is a built time variable, which was defined in kiwix.xcconfig with the value: id997079563.
Now we have Info.plist key APP_STORE_ID with the value: id997079563.

The value itself is the published ID of the app in the App Store (generated by Apple):
https://apps.apple.com/us/app/kiwix/id997079563

We need this value to get the correct URL to redirect users to App Review (of this specific app).

BPerlakiH: On the custom app side it turned out that we can do everything we want without the xcconfig files, so this extra layer can be removed.

Support/kiwix.xcconfig
```diff
@@ -1,5 +0,0 @@
-//  Copyright © 2023 Kiwix.
-
-// Configuration settings file format documentation can be found at:
-// https://help.apple.com/xcode/#/dev745c5c974
-APP_STORE_ID = id997079563
```
BPerlakiH, line 5: @rgaudin formerly it was here

rgaudin (approved):
Got it ; LGTM then

PR #619, opened 2024-01-13
Validate if secrets were set

BPerlakiH:


rgaudin:
Should we remove it now?

BPerlakiH:
Removing now.

PR #621, opened 2024-01-14
Download dependencies on build step according to parameter

BPerlakiH:
Adding an additional flag to the build workflow, if the dependencies should be downloaded as part of this workflow or not.
Will be handy for custom apps, for which we resolve the dependencies earlier.


BPerlakiH:
Once this PR is merged, this can merged as well: https://github.com/kiwix/kiwix-apple-custom/pull/46

Issue #624, opened 2024-01-14
CD is broken since two days!


kelson42:
https://github.com/kiwix/apple/actions/workflows/cd.yml

rgaudin:
It seems we overlooked https://github.com/kiwix/apple/pull/618 as the retry exit code for the DMG was ensuring we wait until the notarization is complete. This must be restored

BPerlakiH:
The fix is here: https://github.com/kiwix/apple/pull/625

PR #625, opened 2024-01-15
Revert CD archive to use python re-try

BPerlakiH:
Fix: #624


rgaudin (approved):
Proper way of this this would have been to revert the commit that broke the feature and add another one to remove it where the script is not needed anymore.

PR #639, opened 2024-01-27
Fastlane for iOS

### LINKED ISSUE #638: Use Fastlane for app builds
We would like to replace our current builds steps with Fastlane, and handle our CI and CD steps with it.

BPerlakiH:
Fixes #638


kelson42:
We won‘t use fastlane, see #638 for details

Issue #647, opened 2024-02-11
Bundle install fails on CI


BPerlakiH:
CI (temporarily) fails with:
```
Bash completion has been installed to:
  /usr/local/etc/bash_completion.d
Installing pre-commit has failed!
Installing xcodegen
Upgrading curl
Homebrew Bundle failed! 1 Brewfile dependency failed to install.
Check for differences in your Brewfile.lock.json!
```

Suggested solution (by brew) is to remove the dependencies we don't need and use --force option.
For CI we do not really need pre-commit, so we can dynamically remove that.

BPerlakiH:
Re-opening, as we need to merge the same changes to kiwix/apple as well, see here: https://github.com/kiwix/apple/pull/648

PR #648, opened 2024-02-11
Fix Bundle install on CI with --force

BPerlakiH:
Fixes: #647


rgaudin (approved):
LGTM ; I suppose there's no harm in having more stuff in the lockfile

PR #652, opened 2024-02-15
Fix default language for library filtering

### LINKED ISSUE #651: Fix default language for Feed filtering
Our feeds when returning the list of zim files, that can be downloaded, we filter those by the current language set in the app.
Since the app is setting the language as a 2 char language code, it cannot filter the zim files properly.

Ideally we should use 3 char language code as a default in our app, it is possible out of the box for iOS 16 and above, and requires a bit of a workaround for iOS 15.

BPerlakiH:
Fixes #651

### DIFF
--- a/ViewModel/LibraryViewModel.swift
+++ b/ViewModel/LibraryViewModel.swift
@@ -57,7 +57,7 @@ public class LibraryViewModel: ObservableObject {
             
             // populate library language code if there isn't one set already
             let currentLanguageCode: String?
-            if #available(iOS 16, *) {
+            if #available(iOS 16, macOS 13, *) {
                 currentLanguageCode = Locale.current.language.languageCode?.identifier(.alpha3)
             } else {
                 // Locale.current.languageCode is returning a 2 char lang code, eg: "en"
--- a/ViewModel/LibraryViewModel.swift
+++ b/ViewModel/LibraryViewModel.swift
@@ -56,10 +56,19 @@ public class LibraryViewModel: ObservableObject {
             Defaults[.libraryLastRefresh] = Date()
             
             // populate library language code if there isn't one set already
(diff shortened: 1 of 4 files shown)


ViewModel/LibraryViewModel.swift
```diff
@@ -56,10 +56,19 @@ public class LibraryViewModel: ObservableObject {
...
-            if Defaults[.libraryLanguageCodes].isEmpty, let currentLanguageCode = Locale.current.languageCode {
-                Defaults[.libraryLanguageCodes] = [currentLanguageCode]
+            let currentLanguageCode: String?
+            if #available(iOS 16, *) {
+                currentLanguageCode = Locale.current.language.languageCode?.identifier(.alpha3)
+            } else {
+                // Locale.current.languageCode is returning a 2 char lang code, eg: "en"
+                // we want a 3 char value, eg: "eng", otherwise we filter out every results
+                // and end up with an empty list in the categories
+                currentLanguageCode = "eng"
```
BPerlakiH, line 66: There's no easy way to get the 3 char value below iOS 16, so the easiest we can do is to fall back to English.

BPerlakiH:
@kelson42 @rgaudin please re-check, it is updated.

rgaudin (approved):
LGTM although the important part is Swift code which I dont know 😉

rgaudin:
haven't tested with this PR code but current main branch on first run has no language selected. I guess this fixes it ?

Screenshot 2024-02-19 at 10 14 00
Screenshot 2024-02-19 at 10 13 35

BPerlakiH:
@rgaudin yes, it is fixing that issue as well.

BPerlakiH:
Also the other PR is also fixing this issue in a way that we no longer allow the user to untick all the languages, it is restricting to have always at least one selected.

PR #653, opened 2024-02-16
Update kiwix feeds

### LINKED ISSUE #650: Update Kiwix Feeds
Currently the Apple apps are using the old feed:
https://library.kiwix.org/catalog/root.xml

We want to update to the latest feeds, as of OPDS API
- https://kiwix-tools.readthedocs.io/en/latest/kiwix-serve.html#http-api
- https://library.kiwix.org/catalog/v2/entries?lang=eng

BPerlakiH:
Fixes #650 
Use the v2 of the API.
Currently we do not do pagination, just downloading the full list of items as before.
Also adjusting the UnitTests.


kelson42 (approved):
Lgtm, but we should make default langauage configuration at app start, not at the time we fetch.

Issue #656, opened 2024-02-17
Filter out empty categories


BPerlakiH:
Filter the displayed categories automatically, so we do not display empty ones. 
Eg. for a language with less content, some of the categories will become empty. 
We want the user to find content in an easy way.

Eg. there's no TED content in French:

We display the list of categories the user can select:

![Image](https://github.com/kiwix/apple/assets/6784320/fb97a778-f4a4-4c8d-a476-bee445361ae4)

However after selecting TED, we display this:

![Image](https://github.com/kiwix/apple/assets/6784320/4374709f-f867-4aa7-9df0-a98296a7521c)

We do know upfront, there's no TED content in French, so we can filter out this category and not display it.

PR #657, opened 2024-02-18
Filter categories by content languages

BPerlakiH:
Fixes: #656 

Also fixes: the language count in settings, and a translations of drop downs in settings.

With this we filter down the categories to non empty ones, depending on the content language settings. Eg: setting it to German only (we won't have TED, which is currently only available in English):

Simulator Screenshot - iPhone 15 - 2024-02-18 at 09 21 49

### DIFF
--- a/App/App_macOS.swift
+++ b/App/App_macOS.swift
@@ -97,7 +97,7 @@ struct RootView: View {
                 if FeatureFlags.hasLibrary {
                     Section("app_macos_navigation.button.library".localized) {
                         ForEach(libraryItems, id: \.self) { navigationItem in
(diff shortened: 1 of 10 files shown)


rgaudin (dismissed):
Looks OK to me but I'm in no capacity to review the code. Seems like behavior is what we want given it's a user-defined filter on languages.

kelson42 (changes requested):
Please fix Codefactor.

BPerlakiH:
> Please fix Codefactor.

Please see it updated.

rgaudin:
@BPerlakiH @kelson42 forced-pushed and only requested codefactor to be fixed. You can merge

PR #662, opened 2024-02-21
Filter out igl lang code

### LINKED ISSUE #661: 3.3.0 Release issue

BPerlakiH:
Fixes #661


rgaudin (approved):
LGTM ; we don't wont want any 3 chars language code as this would indicate ISO-639-3 but we want to fail should it occur so we can actu upon it and fix.

PR #665, opened 2024-02-22

translatewiki:
Translation updates


BPerlakiH (commented):
It seems to me Dagbani is another example of a ISO-639-3 only language code.

BPerlakiH:
I've added changes to filter out for now this language code, as Apple won't approve non ISO-693-2 codes.

BPerlakiH (approved):

PR #666, opened 2024-02-22
663 feed fix for language changes

### LINKED ISSUE #663: Feed fix for language changes
I did test the following scenario:
- clean up everything, download the AppStore version
- update it via TestFlight to our latest release candidate

We have the following problem, the user's default / selected language will get updated properly, once the catalog is fetched, yet
the user will still face an empty list of categories.
The reason is the following:
- the feed now returns 3 char language codes, the default / selected language is also in this format. So far so good.
- the listing of the categories are based on what is stored in the local DB
- the DB won't get updated with the new language values coming from the feed, it is only comparing if the DB has entry for the given ZIM file
- the DB entry remains with the old language values: eg: "en",
- while trying to list the categories, we will search for eg: "eng"
- result: nothing is displayed for the user

BPerlakiH:
Fixes #663 

I have tested and fixed both scenarios:
- Fresh install of this version = new user experience
- Install the AppStore version, fetch the catalog, then install this version, check the catalog before and after fetching within this build

There was also a performance issue that had to be solved, the updates were really slow (we have about 3000 records). 
One solution was to do them in batch, one batch per language. 
Later it turned out that even that is not fast enough, as we have this "database changed" event hooked in (partially for iCloud), and it was doing a lot of extra work for each language update batch. The solution was to skip that part while we do the fetching / updating.

### DIFF
--- a/Model/CategoriesToLanguage.swift
+++ b/Model/CategoriesToLanguage.swift
@@ -11,8 +11,8 @@ struct CategoriesToLanguages {
     private let dictionary: [Category: Set<String>] = Defaults[.categoriesToLanguages]
 
     func has(category: Category, inLanguages langCodes: Set<String>) -> Bool {
-        guard !langCodes.isEmpty else {
-            return true // no languages provided, do not filter
+        guard !langCodes.isEmpty, !dictionary.isEmpty else {
+            return true // no languages or category filters provided, do not filter
         }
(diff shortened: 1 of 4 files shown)


rgaudin (dismissed):
LGTM

BPerlakiH:
Fix is included in: https://github.com/kiwix/apple/pull/670/

PR #670, opened 2024-02-24
Fix by reverting to 2 letter lang codes in DB

### LINKED ISSUE #663: Feed fix for language changes

### LINKED ISSUE #664: Incoherency in Settings (and empty list of books to download)
After updating to latest testflight.

My online list of books was empty.

My settings show „French“
image

But once I display the list, french is not selected anymore
image

I should see french books and settings should be coherent

### LINKED ISSUE #667: Update 2 letters language code settings to 3 letters
Should applied just right after the new update (run once):
* on iOS15: based on an ad-hoc hardcoded mapping list of X importances languages codes, apply the update. If no matching is found, then remove the language code. If at the end, list of preferend languages is empty, will default to `eng`.
* on iOS16+: use native primitives to convert to 3 letters language code.

This piece code should be removed after a few additional releases:
* Create a dedicated issue for that
* Put clear comment in the code that this is temporary code

### LINKED ISSUE #668: Fix root cause of language problems
After throughout analysis, why the AppStore version is working as it is, what were the changes in the meantime.
I did found the following:
- it all started with: https://github.com/kiwix/apple/issues/588
- I removed the deprecated function in https://github.com/kiwix/apple/pull/629/files
- The former function was returning a single string, whereas the new one might return a comma separated list of strings.
- The mapping from a list of strings to a list of Locales, was done a "level higher" not in the objective-c function but in swift.
- The result is that the objective-c function was in fact converting the language codes from the feed to a 2 letter language code (across all iOS versions), and that is what we were storing in the database, which is accidentally or purposefully but in an undocumented way is working
- When adding the swift function I was checking against the feed values, and from [that PR](https://github.com/kiwix/apple/pull/629/files) on the DB was using 3 letter codes.
- Unfortunately the unit-tests were not run by me on that PR, the CI is currently not running them
- The tests were updated at a later point in time here: https://github.com/kiwix/apple/pull/652/files#diff-edd62679239139566e8891207312793f6afaf35c590661caebec8c72f81bbfc3

## Proposition
The simplest, and most straightforward solution at this point is, use the existing working solution and store the language codes as 2 letter codes, apply the same for user settings. This solves the issue with iOS 15 and below, without having to manually map codes, as suggested in: https://github.com/kiwix/apple/issues/667

**This purposeful workaround has to be clearly documented in code.**

To avoid similar problems in the future, we need to make sure [unit tests are run during CI builds](https://github.com/kiwix/apple/issues/669) (issue created)

BPerlakiH:
Fixes #668 
Fixes #667 
Fixes #664 
Fixes #663


rgaudin (approved):
LGTM in that corresponds to the proposed solution but I though we were using this opportunity to switch to ISO-639-3 everywhere, fixing the ISO-639-1 parts.

Now this is the opposite, we are committing to ISO-639-1 and converting from ISO-639-3.

We don't **have to** use the same standard but if we don't it should be clearly agreed-upon and documented.
This looks like a shortcut to make it work while we had an opportunity for a longer-term fix with the discussed approach.

PR #672, opened 2024-02-25
Switch to ISO-639-3 language codes everywhere

### LINKED ISSUE #667: Update 2 letters language code settings to 3 letters

BPerlakiH:
Fixes #667


rgaudin (dismissed):
LGTM ; comments are really helpful 👍

kelson42:
@BPerlakiH Please resolve conflicts. In the future, just base this branch/PR on the depending branch and not on `main` to avoid this kind of problem.

BPerlakiH:
I've updated it.

kelson42:
@BPerlakiH This is not rebased properly. Please learn how to do that properly.

BPerlakiH:
I've updated it, now it looks better.

rgaudin (approved):

PR #678, opened 2024-03-03
601 bookmarks intial cleanup

BPerlakiH:
Initial clean up PR, that removes unused functions, clarifies the misleading naming of file URL Bookmark, which has nothing to do with the Bookmarks really.

No change of functionality, just more clarity in code.


rgaudin (approved):
I guess it makes sense

kelson42:
I would have love to release `3.3.0` first but it seems this still needs a bit of work, so merging this as well.

PR #679, opened 2024-03-03
601 bookmarks initial libkiwix implementation

BPerlakiH:
Initial attempt to use libkiwix's bookmark functionality.

Not to be merged!

After some battle with the interoperability on 3 levels swift / objective-c / c++, this is an initial implementation I could do to re-use the LibKiwix bookmark functionality.

There are several issues with this approach, and several things are missing compared to what we currently have via CoreData (on device SQL DB):
- it is a lot of bridging between those languages. It is a headache to maintain such code, it's not "safe", and hard to follow
- in order to implement this, and make it actually work, I had to look under the C++ source code to understand how it is working
- there are a lot of intermediate steps taken and types to be created in order to create a bookmark
- the current functionality **is not persisting the bookmarks**, meaning it is a memory only storage, so when the app is closed the bookmarks are lost
- the `kiwix::Bookmark` is not supporting "html snippet", and not supporting thumbnail url (see the screenshot below)
- I haven't found any built in way to add the "search bookmarks" functionality
- We would need to write all the "reactive" part of the code, to make sure any bookmarks change is triggering the appropriate UI change as well (eg. bookmark added / removed), currently we even have animated tiles as they are added / removed, thanks to CoreData bindings

Screenshot 2024-03-03 at 20 42 47
As you can see on the screenshot, in order to list the Bookmarks, we need the following minimum data:
- url to the Article
- title
- snippet (the short description)
- thumbnail url
The above values are currently HTML parsed out from the Article, but the LibKiwix::Bookmark struct is not supporting these additional fields

At this stage I feel that going down this route we will add a lot of extra binding code, and we would need to re-implement most of the things the CoreData is offering out of the box.

### DIFF
--- a/Model/LibBookmarks/LibBookmarksBridge.mm
+++ b/Model/LibBookmarks/LibBookmarksBridge.mm
@@ -33,7 +33,8 @@
 - (BOOL) isBookmarked: (nonnull NSURL *) url inZIM: (nonnull NSUUID *) zimFileID {
     std::string url_c = [url fileSystemRepresentation];
     std::string fileID_c = [[[zimFileID UUIDString] lowercaseString] cStringUsingEncoding:NSUTF8StringEncoding];
-    for (kiwix::Bookmark bookmark: self.library->getBookmarks()) {
+    const std::vector<kiwix::Bookmark> bookmarks = self.library->getBookmarks(true);
+    for (kiwix::Bookmark bookmark: bookmarks) {
         if (bookmark.getUrl() == url_c && bookmark.getBookId() == fileID_c) {
             return true;
         }
@@ -42,11 +43,14 @@
 }
 
 - (void) add: (LibBookmark *) bookmark {
-    self.library->addBookmark([bookmark bridged]);
+    kiwix::Book book = [bookmark book];
+    self.library->addBook(book);
+    kiwix::Bookmark bridgedBookmark = kiwix::Bookmark(book, bookmark.url_c, bookmark.title_c);
+    self.library->addBookmark(bridgedBookmark);
 }
 
 - (void) remove: (LibBookmark *) bookmark {
-    self.library->removeBookmark(bookmark.zimFileID_c, bookmark.url_c);
+    self.library->removeBookmark(bookmark.zimID_c, bookmark.url_c);
 }
 
 @end
--- a/Model/LibBookmarks/LibBookmarksBridge.mm
+++ b/Model/LibBookmarks/LibBookmarksBridge.mm
@@ -1,0 +1,52 @@
+//
+//  LibBookmarksBridge.mm
+//  Kiwix
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdocumentation"
+#include "kiwix/library.h"
+#include "kiwix/bookmark.h"
+#pragma clang diagnostic pop
+
+#import "LibBookmarksBridge.h"
+
+# pragma mark - Private properties
+@interface LibBookmarksBridge ()
+
+@property kiwix::LibraryPtr library;
+
+@end
+
+
+# pragma mark - Implementation
+
+@implementation LibBookmarksBridge
+
+- (instancetype _Nonnull)init {
+    self = [super init];
+    if (self) {
+        self.library = kiwix::Library::create();
+    }
+    return self;
+}
+
+- (BOOL) isBookmarked: (nonnull NSURL *) url inZIM: (nonnull NSUUID *) zimFileID {
+    std::string url_c = [url fileSystemRepresentation];
+    std::string fileID_c = [[[zimFileID UUIDString] lowercaseString] cStringUsingEncoding:NSUTF8StringEncoding];
+    for (kiwix::Bookmark bookmark: self.library->getBookmarks()) {
+        if (bookmark.getUrl() == url_c && bookmark.getBookId() == fileID_c) {
+            return true;
+        }
+    }
+    return false;
+}
+
+- (void) add: (LibBookmark *) bookmark {
+    self.library->addBookmark([bookmark bridged]);
+}
+
+- (void) remove: (LibBookmark *) bookmark {
+    self.library->removeBookmark(bookmark.zimFileID_c, bookmark.url_c);
+}
+
+@end
--- a/Model/ZimFileService/ZimFileService.mm
+++ b/Model/ZimFileService/ZimFileService.mm
@@ -88,7 +88,7 @@
 }
 
 - (void)close:(NSUUID *)zimFileID {
-    self.archives->erase([self zimfileID_C: zimFileID]);
+    self.archives->erase([ZimFileService zimfileID_C: zimFileID]);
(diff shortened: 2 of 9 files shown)


Tests/LibBookmarkTests.swift
```diff
@@ -0,0 +1,30 @@
...
+    private let testTitle: String = "Dionysius of Halicarnassus On Literary Composition"
+    private let bookmarkURL = URL(string: "kiwix://FF7D59DE-0FD8-F486-09FA-C57904646707/A/Dionysius%20of%20Halicarnassus%20On%20Literary%20Composition.50212.html")!
+
+    func testIfZimFileExists() throws {
+        XCTAssertNotNil(testZimURL)
+    }
+
+    func testAddingBookmark() {
+        let fileURLData = ZimFileService.getFileURLBookmarkData(for: testZimURL)!
+        try! ZimFileService.shared.open(fileURLBookmark: fileURLData)
```
BPerlakiH, line 20: This is a bit cumbersome, we have a hash of stored `Archive` under `ZimFileService` that came from parsing things, using `kiwix::library`` mostly (in `OPDSParser`), but we are not holding on to the `library`.
In order for the bookmarks to be successfully added, we do need to add them to the `library` itself.
So we have a kind of dualism here, where the list of items are going through one temporary instance of `library` to be parsed and inserted into the DB, whereas in order to store them we need another instance of `library` where the `kiwix:Book` had to be added, otherwise the added `kiwix::Bookmark` will be `invalid`.

Model/ZimFileService/ZimFileService.mm
```diff
@@ -115,6 +115,16 @@ + (ZimFileMetaData *)getMetaDataWithFileURL:(NSURL *)url {
...
 }
 
+- (kiwix::Book) getBookBy:(nonnull NSUUID *) fileZimID {
+    kiwix::Book book = kiwix::Book();
+    try {
+        NSURL *fileURL = [ZimFileService.sharedInstance getFileURL: fileZimID];
+        book.update(zim::Archive([fileURL fileSystemRepresentation]));
+    } catch (std::exception e) { }
+    return book;
+}
```
BPerlakiH, line 125: It is a lot of legwork at the moment:
- we need a `kiwix::Book` in order to be able create a `kiwix::Bookmark`
- we need a `kiwix::Archive` in order to be able to create `kiwix::Book`
Whereas all we want is to store:
- 2 strings (title, description) a thumbnail URL, keyed by the article URL, which is more or less this much code:
```
struct SwiftBookmark {
    let title: String
    let description: String
    let thumbnail: URL
}

final class SwiftBookmarkManager {
    private var dict: [URL: SwiftBookmark] = [:]
}
```

Model/LibBookmarks/LibBookmarksBridge.mm
```diff
@@ -0,0 +1,56 @@
...
+    if (self) {
+        self.library = kiwix::Library::create();
+    }
+    return self;
+}
+
+- (BOOL) isBookmarked: (nonnull NSURL *) url inZIM: (nonnull NSUUID *) zimFileID {
+    std::string url_c = [url fileSystemRepresentation];
+    std::string fileID_c = [[[zimFileID UUIDString] lowercaseString] cStringUsingEncoding:NSUTF8StringEncoding];
+    const std::vector<kiwix::Bookmark> bookmarks = self.library->getBookmarks(true);
```
BPerlakiH, line 36: if this call can be used without the argument and it works equally good:
`self.library->getBookmarks();`
The tricky part is, which is not much documented, is that the `kiwix::Book` needs to be added to the library,
as on line 47.
If the `Book` is not added to the `library`, the `self.library->getBookmarks();` won't return it, as it is an "invalid" bookmark..
The other solution is not to add the Book to the library and call `getBookmarks(false)`.
Well... without looking into the C++ code it is not obvious...

Tests/LibBookmarkTests.swift
```diff
@@ -0,0 +1,30 @@
...
+        let bookmarks = LibBookmarks.shared
+        XCTAssertFalse(bookmarks.isBookmarked(url: bookmarkURL))
+
+        let bookmark = LibBookmark(withUrl: bookmarkURL, withTitle: testTitle)!
+        bookmarks.addBookmark(bookmark)
+        
+        XCTAssertTrue(bookmarks.isBookmarked(url: bookmarkURL))
+    }
+
```
BPerlakiH, line 29: This can be compared with testing the pure swift proposed implementation:
```
    func testSwiftBookmars() {
        let bookmarks = SwiftBookmarkManager()
        XCTAssertFalse(bookmarks.isBookmarked(withURL: bookmarkURL))
        let bookmark = SwiftBookmark(title: testTitle, description: "test snippet", thumbnail: nil)
        bookmarks.add(bookmark: bookmark, withURL: bookmarkURL)
        XCTAssertTrue(bookmarks.isBookmarked(withURL: bookmarkURL))
    }
```

kelson42:
>     * it is a lot of bridging between those languages. It is a headache to maintain such code, it's not "safe", and hard to follow

That everything has to be bridged is normal. That this is not "safe", is not normal. Please do the necessary - like reporting problems at libkiwix if necessary - that things are "safe" (understandable and clean).
 
>     * in order to implement this, and make it actually work, I had to look under the C++ source code to understand how it is working

This is also not normal, please report things in issues at `openzim/libkiwix`.

>     * there are a lot of intermediate steps taken and types to be created in order to create a bookmark

Definitively and this is normal.

>     * the current functionality **is not persisting the bookmarks**, meaning it is a memory only storage, so when the app is closed the bookmarks are lost

Obviously it has to, and libkiwix offer primitive to save/load bookmarks to/from the filesystem.

>     * the `kiwix::Bookmark` is not supporting "html snippet", and not supporting thumbnail url (see the screenshot below)

Any argument to keep it? What is the value of such snippet?

>     * I haven't found any built in way to add the "search bookmarks" functionality

How does it technically work today? Do we have a dedicated DB with search index? Or just we go through all bookmarks? 

>     * We would need to write all the "reactive" part of the code, to make sure any bookmarks change is triggering the appropriate UI change as well (eg. bookmark added / removed), currently we even have animated tiles as they are added / removed, thanks to CoreData bindings.

Anyway to avoid getting ride of these animations by moving to libkiwix?

> 
> Screenshot 2024-03-03 at 20 42 47
> 
> As you can see on the screenshot, in order to list the Bookmarks, we need the following minimum data:
> 
>     * url to the Article
> 
>     * title
> 
>     * snippet (the short description)
> 
>     * thumbnail url
>       The above values are currently HTML parsed out from the Article, but the LibKiwix::Bookmark struct is not supporting these additional fields

Please open an issue for the two last items at `kiwix/libkiwix` which are currently not supported.

> At this stage I feel that going down this route we will add a lot of extra binding code, and we would need to re-implement most of the things the CoreData is offering out of the box.

Binding code should not be a complicated thing, actually there is tools to do that more or less automatically. For the rest, what is exactly the difficulty (you have talked about this animation but what else?)

PR #684, opened 2024-03-05
Fix language selection initial state UI problems

### LINKED ISSUE #680: No language selected
Screenshot 2024-03-04 at 12 02 05

Clean environment, latest TF build (139):

- opened the App
- Was greeted with option to open a ZIM file or fetch the catalog.
- Didn't fetch the catalog, looked at the side bar menu, all empty until UI froze for a bit
- then I had the categories tab filled: it's all English
- Opened settings. Saw that no language is selected.

I might have misunderstood but I thought the default (`eng`) would be selected because showing a filtered (to `eng`) catalog with no language selected in settings is deceptive IMO.

BPerlakiH:
Fixes #680

The initial state of the applications is that we do not have a list of languages.
We need to indicate it to the user in a reasonable way.
On macOS we also need to update the list of languages if the initial refresh starts from the settings (the same tab, where the list of languages are displayed).
On iPhones the list of languages is under a different section in the settings menu, but that navigation link should also be disabled, when there are no languages to select from.

Also while we are refreshing the language selection should be disabled for the user.

After the changes:
Screenshot 2024-03-05 at 13 51 37 Medium
Screenshot 2024-03-05 at 13 51 42 Medium
Screenshot 2024-03-05 at 13 51 50 Medium

Similarly on iOS:
Simulator Screenshot - iPad mini (6th generation) - 2024-03-05 at 13 48 21 Medium
Simulator Screenshot - iPad mini (6th generation) - 2024-03-05 at 13 48 31 Medium

### DIFF
--- a/SwiftUI/Model/DefaultKeys.swift
+++ b/SwiftUI/Model/DefaultKeys.swift
@@ -34,7 +34,6 @@ extension Defaults.Keys {
     static let libraryAutoRefresh = Key<Bool>("libraryAutoRefresh", default: true)
     static let libraryUsingOldISOLangCodes = Key<Bool>("libraryUsingOldISOLangCodes", default: true)
     static let libraryLastRefresh = Key<Date?>("libraryLastRefresh")
-    static let libraryLastRefreshTime = Key<Date?>("libraryLastRefreshTime")
     
     static let isFirstLaunch = Key<Bool>("isFirstLaunch", default: true)
     static let downloadUsingCellular = Key<Bool>("downloadUsingCellular", default: false)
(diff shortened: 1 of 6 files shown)


rgaudin (approved):
LGTM

SwiftUI/Model/DefaultKeys.swift
```diff
@@ -34,7 +34,6 @@ extension Defaults.Keys {
     static let libraryAutoRefresh = Key<Bool>("libraryAutoRefresh", default: true)
     static let libraryUsingOldISOLangCodes = Key<Bool>("libraryUsingOldISOLangCodes", default: true)
     static let libraryLastRefresh = Key<Date?>("libraryLastRefresh")
-    static let libraryLastRefreshTime = Key<Date?>("libraryLastRefreshTime")
```
BPerlakiH, line 37: This one wasn't in use at all.

PR #686, opened 2024-03-08
634 ZIM download progress fix

### LINKED ISSUE #634: ZIM download experience is terrible
I don't know if it has ever worked properly but I am testing this with macOS nightly.

Downloading a ZIM from the in-app catalog is very difficult and frustrating

- In _Categories_ view, selecting a ZIM and clicking _Download_ has no visual feedback. Button is still there, unchanged.
- Going to _Downloads_ view, there's usually nothing. If I clicked many times at previous step, sometimes I get an entry in _Downloads_ view.
- Progress bar in _Downloads_ view is never updated.
- Going back and forth in _Opened_/_Downloads_ view, I can figure out if download completed (only tested with small ZIMs obviously)
- ZIM might appear in _Opened_ view if it completed download.
- In _Opened_ view, clicking on the ZIM brings same right-column content as in _Categories_ view: with just the download button, no open one.
- Closing the app completely and re-opening allows _Opened_ view to reveal proper buttons and access the ZIM content.

BPerlakiH:
Fixes #634 

Tested on the attached "broken" Library/container.
The very specifics how downloads and progress are working:
- download starts via http, and we get progress updates, eg: 500 bytes loaded out of 2MB
- these progresses are marked in the SQLite Database under "DownloadTasks"
- the DB is updated from the background context, this allows CoreData syncing to the "view context", and swift ui views are automagically updated
- for this syncing to happen it looks up the list of transactions from a specific time stamp, called token, which is stored in UserDefaults

## What went wrong:
The token is invalid in the "broken" container, meaning there are no transactions to be found after this date.
This can happen for various reasons, either some transactions were corrupted, or old transactions were deleted (as there's a cleanup of transactions at some point as well).
Since the token was invalid, the synchronisation phase could not happen, and there was no recovery.
**The fix for that is to reset the token**. From then on the next synchronisation attempt will carry out successfully, and a new valid token value will be set.

The other issue, I have found that it was crashing from time to time, and that is because the dictionary holding the progress value was read from / writen to different threads, in short in a non thread safe manner. 
Fix: I've added a thread safe wrapper to this dictionary.

### DIFF
--- a/Model/DownloadService.swift
+++ b/Model/DownloadService.swift
@@ -36,8 +36,7 @@ final class DownloadService: NSObject, URLSessionDelegate, URLSessionTaskDelegat
     
     private let queue = DispatchQueue(label: "downloads")
     private let progress = DownloadProgress()
-    private var heartbeat: Timer?
-    
+    @MainActor private var heartbeat: Timer?
     var backgroundCompletionHandler: (() -> Void)?
     private lazy var session: URLSession = {
         let configuration = URLSessionConfiguration.background(withIdentifier: "org.kiwix.background")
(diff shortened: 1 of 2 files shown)


rgaudin (approved):
Thank you for tracking this down and finding the root cause.
LGTM from what I understand. Can't wait to test a build including it

Issue #601, opened 2023-12-22
Handling bookmarks from previous zim files


BPerlakiH:

BPerlakiH:
Currently if we update a ZIM file, the bookmarks created formerly remains linked to to old ZIM file, and stop working.
Create a plan how we can improve this on the iOS/macOS side.

kelson42:
@BPerlakiH **Please save ASAP (and update DWDS) app, the ZIM name in your bookmarks.** Once libkiwix ready, implementing this ticket will imply relying on the libkiwix for that AND create a migration plan.

BPerlakiH:
# Bookmarks

## The problem
If a newer version of a zim file is added to an app the bookmarks created with the old zim file won't work anymore.

## How the bookmarks are working on iOS/macOS?

It is an entity in the local database, based on the `articleURL`, eg:
`kiwix://A992BF76-CA94-6B60-A762-9B5BC89B5BBF/wb/Saftladen`
or
`kiwix://A992BF76-CA94-6B60-A762-9B5BC89B5BBF/wb/08_.15`
where `kiwix` is our custom namespace
the `A992BF76-CA94-6B60-A762-9B5BC89B5BBF` is the `fileID` of the `ZimFile` (also stored as an entity).

The `fileID` is comming from the zim file metadata itself, and it is not changing. It has been tested:
- across devices
- with fresh install
- even with renaming the file

The `Bookmark` entity is also referencing in the `ZimFile` in the DB via a unique (auto-incremented) primary key.

## Custom apps problem

Our current solution is enforcing to always use a single file per custom app. It is imported into the DB on launching the app.

When we update the zim file and create a new app release, the new zim file will be inserted into the DB next to the old one.

The problem we are facing is that the bookmarks will still point to the old file, both in referencing the primary key of the old file (which is no longer loaded),
and their url will contain the old value as the first part of their path (see above).

## Custom apps solution

In case of an update, we can assume that if a new file is imported, the old entries in the DB should be updated.

The following solution has been tested (by manually editing the DB):
1) change the old bookmarks url, by replacing the first part of the path to the value comming from the fileID of the new zim file, eg changing the one above to:
`kiwix://64C3EA1A-5161-2B94-1F50-606DA5EC0035/wb/Saftladen`
2) update the old bookmarks zim file primary key reference, so they are pointing to the new file inserted, eg changing primary key from `1` to `2`.

[...]

PR #688, opened 2024-03-09
Custom apps ZIM migration

BPerlakiH:
Fix for #601 - Migrating bookmarks for custom app.

There were a couple of thing to do:
- make sure we migrate the bookmarks
- migrate the tabs, which have "interactive data" - a kind of snapshot of the tab for webView, in a form of Data, which is now migrating at the point of opening the URL from that data
- make sure we delete the old file reference in the DB - as currently we were displaying a list of local ZIM files in the menu, (up to 5). For custom apps it should really be only 1 item, so we make sure we remove the reference to the old ZIM file.


kelson42:
@BPerlakiH Can you please fix the codefactor?

kelson42:
@BPerlakiH Ready to review?

BPerlakiH:
There's one more thing need to be fixed here, which is the "interactive data" saved together with the state of each tab.
Will update the PR soon.

BPerlakiH:
@kelson42 @rgaudin it is ready for review.

SwiftUI/Model/ZimMigration.swift
```diff
@@ -0,0 +1,96 @@
...
+                  let latest = zimFiles.popLast() else {
+                return
+            }
+
+            context.mergePolicy = NSMergePolicy.mergeByPropertyObjectTrump
+            for zimFile in zimFiles {
+                migrateFrom(zimFile: zimFile, toZimFile: latest, using: context)
+            }
+        }
+        debugPrint("migrationForCustomApps Complete")
```
rgaudin, line 35: Is this supposed to stay?

rgaudin (approved):
Can't say much

SwiftUI/Model/ZimMigration.swift
BPerlakiH: Thanks, good catch, I've removed it now.

rgaudin (approved):

PR #693, opened 2024-03-16
Fix tab selection post deletion

### LINKED ISSUE #692: Incorrect tab selection after tab deletion
## Deleting the currently selected tab, should always select another tab if that's possible.
The issue is that it is not always working.
The current solution is attempting to select the tab that was "historically" created before the one deleted.
This works in the following tabs scenario:
```
Tab 1 (created @ 09:00)
Tab 2 (created @ 10:00) - selected and now marked to be deleted
```
Correctly, Tab 1 is selected.

The same approach won't work here:
```
Tab 1 (created @ 09:00) - selected and now marked to be deleted
Tab 2 (created @ 10:00) 
```
There are no tabs "created before 09:00" therefore the result is:
Tab 1 is deleted, but nothing is selected.

## Solution
Do not rely on the timestamp of the tabs, they are already ordered by that anyways, instead select the Tab before, if there's such, or select the Tab after if there's such.

BPerlakiH:
Fixes #692

### DIFF
--- a/ViewModel/NavigationViewModel.swift
+++ b/ViewModel/NavigationViewModel.swift
@@ -62,7 +62,9 @@ class NavigationViewModel: ObservableObject {
     /// - Parameter tabID: ID of the tab to delete
     func deleteTab(tabID: NSManagedObjectID) {
         Database.performBackgroundTask { context in
-            guard let tabs: [Tab] = try? context.fetch(Tab.fetchRequest()),
+            let sortByCreation = [NSSortDescriptor(key: "created", ascending: false)]
+            guard let tabs: [Tab] = try? context.fetch(Tab.fetchRequest(predicate: nil,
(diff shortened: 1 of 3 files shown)


ViewModel/NavigationViewModel.swift
```diff
@@ -20,6 +20,8 @@ class NavigationViewModel: ObservableObject {
         let tab = Tab(context: context)
         tab.created = Date()
         tab.lastOpened = Date()
```
BPerlakiH, line 22: We were using the same set of operations after creating a tab:
- obtaining an DB id for it
- saving it
so it makes sense to bring those under the same function.

kelson42:
@BPerlakiH Please update changelog

BPerlakiH:
CHANGELOG updated.

PR #694, opened 2024-03-16
532 remove long press and re-arrange buttons

### LINKED ISSUE #474: Ability to hide/move "random article" button
On my iPhone (13 Pro), the "random article" button takes up valuable space in the action bar, making "Settings" and "Library" collapse into the overflow menu (sorry, not sure if these are the correct UI terms).

This is purely anecdotal/personal, but I use either option much more frequently than opening a random article, and it's annoying having to do the extra tap/step every time.

### LINKED ISSUE #532: iOS 17.1 Zim list inaccessible
On iOS 17.1, once a zim file is loaded there does not seem to be a way to exit the file, which would allow the user to either choose another, or download another to be used in tandem. The only solution I have found so far is deleting the zim file from the phone via storage management. This feels unintentional, though my experience with this app is limited so I am free to stand corrected. Maybe adding the "no file loaded" dialogues always could fix this?image

BPerlakiH:
Fixes #532 
Fixes #474 

Re-arranged the buttons, so no long press is needed to access them:

Simulator Screenshot - iPhone 6s iOS 15 0 - 2024-03-17 at 00 23 07

### DIFF
--- a/App/CompactViewController.swift
+++ b/App/CompactViewController.swift
@@ -1,48 +1,36 @@
 //
 //  CompactViewController.swift
 //  Kiwix
-//
-//  Created by Chris Li on 9/4/23.
-//  Copyright © 2023 Chris Li. All rights reserved.
-//
 
 #if os(iOS)
 import Combine
 import SwiftUI
 import UIKit
-import SwiftBackports
 
 final class CompactViewController: UIHostingController<AnyView>, UISearchControllerDelegate, UISearchResultsUpdating {
-
-    private enum Const {
-        #if os(macOS)
-        static let randomButtonTitle: String = "article_shortcut.random.button.title.mac".localized
-        #else
-        static let randomButtonTitle: String = "article_shortcut.random.button.title.ios".localized
-        #endif
-    }
     private let searchViewModel: SearchViewModel
     private let searchController: UISearchController
     private var searchTextObserver: AnyCancellable?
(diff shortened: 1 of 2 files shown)


kelson42:
@BPerlakiH Thank you but:
* Can you please update the changelog (have move the tickets to 3.3.0)
* Not sure what you have done actually, so I guess basically there is no impact beside fixing the problem. Is thatbright?

rgaudin:
> * Not sure what you have done actually, so I guess basically there is no impact beside fixing the problem. Is thatbright?

I don't know if it's bright 😀 but it's in the PR description: _Re-arranged the buttons, so no long press is needed to access them_

- Random is moved outside of bar to go to top, next to search
- tab icon only reacts to single-tap and displays the tab list
- library icon is moved from menu to bar and avail on single-tap
- menu icon is added (we now have 7 icons instead of 6) and serves on single-tap what was on long-press-tab-icon

Bookmark icon used to work with single-tap to list bookmarks and long-press to choose to either see bookmarks or add one. That's another hard-to-guess behavior but I #695 fixes it by removing the long-press and having the bookmark-this page action inside the bookmarks drawer

rgaudin (approved):
LGTM from a user's standpoint although I haven't tested it. @kelson42 does that look OK to you? Do you want a testflight build with those two PRs to test it live?

PR #695, opened 2024-03-17
532 bookmark no long press

### LINKED ISSUE #532: iOS 17.1 Zim list inaccessible

BPerlakiH:
Fixes #532 additional feature change:

Adding the option to add and remove bookmarks without long-press gesture as well, by adding the bookmark icon to the opened sheet top right menu item.
It can have 3 states: 
- disabled if no url is loaded
- enabled, but off - when the url is not bookmarked
- enabled and on - when the url is bookmarked already

Simulator Screenshot - iPhone 6s iOS 15 0 - 2024-03-17 at 01 02 55 Medium
Simulator Screenshot - iPhone 6s iOS 15 0 - 2024-03-17 at 01 02 58 Medium
Simulator Screenshot - iPhone 6s iOS 15 0 - 2024-03-17 at 01 02 42 Medium

### DIFF
--- a/Views/Buttons/BookmarkButton.swift
+++ b/Views/Buttons/BookmarkButton.swift
@@ -80,8 +80,11 @@ struct BookmarkButton: View {
                             }
                         } label: {
                             Label {
-                                Text(browser.articleBookmarked ?
-                                     "common.dialog.button.remove_bookmark".localized : "common.dialog.button.add_bookmark".localized)
+                                Text(
+                                    browser.articleBookmarked ?
+                                    "common.dialog.button.remove_bookmark".localized :
+                                        "common.dialog.button.add_bookmark".localized
+                                )
                             } icon: {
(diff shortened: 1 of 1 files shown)


rgaudin (approved):
LGTM ; Is the bookmark icon in toolbar changing state based on whether current page is already bookmarked?

BPerlakiH:
> Is the bookmark icon in toolbar changing state based on whether current page is already bookmarked?

Yes it is updating dynamically.

kelson42:
@BPerlakiH Please rebase and fix conflicts

BPerlakiH:
@kelson42 I've rebased it

PR #700, opened 2024-03-18
DWDS Fix app crash, and migrate tabs interactionState data

### LINKED ISSUE #696: Custom app open tabs are not properly migrated
Steps:
* Open a tab in a custom app
* Run the cutom app update
* Relaunch the app

You get:
Your tabs can not be open/refeshed, you get one error per tab and something like this
image

I expect:
The tabs to reopen properly

### LINKED ISSUE #697: Latest dwds testflight crash at start
I have make the report using testflight

BPerlakiH:
Fixes #696 
Fixes #697 
Make sure we migrate the so called "interactionState" of each tab into the new ZIM file.
Make sure we do the DB fetching from the main thread context to avoid the crash.

InteractionState:
It's a data format that Apple uses, to restore the whole web view state with the current URL scroll position, and even with a bit of previous navigation history.

The problem with that is that it is not transparent, how it is encoded / decoded, and there's no direct way of editing the content.

The former solution, I had was to restore this state, and once the page started to load the URL fix the URL to be the migrated one.
This is not fully working as expected, since setting the interactionState data on the WebView already triggers loading the URL, and even if we change it, it will trigger the error pop-up, even if the 2nd load from setting the right URL succeeds.
Temporarily turning off the pop-up doesn't seemed like a good idea for me.

So instead of trying to work around the workaround, I found it easier to just "edit" the interactionState data (by decoding it to string, editing and encoding again) during migration. This fixes the current URL to be loaded in each tab, and can fix the navigation history as well.


App/App_macOS.swift
```diff
@@ -120,6 +120,10 @@ struct RootView: View {
                         if let windowNumber = window?.windowNumber {
                             browser.restoreByWindowNumber(windowNumber: windowNumber,
                                                           urlToTabIdConverter: navigation.tabIDFor(url:))
```
BPerlakiH, line 122: Fallback, if nothing can be re-opened it should load the home page, instead of a blank page.

SwiftUI/Model/ZimMigration.swift
```diff
@@ -34,15 +34,6 @@ enum ZimMigration {
         }
     }
```
BPerlakiH, line 36: This URL mapping to custom apps is no longer needed, as this was applied "after" the interactionState was restored on the WebView, and the URL was re-set with the result of this.

SwiftUI/Model/ZimMigration.swift
```diff
@@ -74,7 +66,7 @@ enum ZimMigration {
     private static func latestZimFileHost() async -> String {
         if let newHost = await Self.newHost { return newHost }
         // if it wasn't set before, set and return by the last ZimFile in DB:
-        guard let zimFile = try? requestLatestZimFile.execute().first else {
+        guard let zimFile = try? Database.viewContext.fetch(requestLatestZimFile).first else {
```
BPerlakiH, line 69: This part was causing the crash, as the fetch started from the wrong "context".

Tests/BookmarkMigrationTests.swift
```diff
@@ -13,4 +13,18 @@ final class BookmarkMigrationTests: XCTestCase {
         let expectedURL = URL(string: "kiwix://A992BF76-CA94-6B60-A762-9B5BC89B5BBF/wb/Saftladen")!
         XCTAssertEqual(url.updateHost(to: newHost), expectedURL)
     }
+
```
BPerlakiH, line 16: This is how the interactive data looks like more or less in a string format.

ViewModel/BrowserViewModel.swift
```diff
@@ -211,24 +211,9 @@ final class BrowserViewModel: NSObject, ObservableObject,
     private func restoreBy(tabID: NSManagedObjectID) {
         if let tab = try? Database.viewContext.existingObject(with: tabID) as? Tab {
             webView.interactionState = tab.interactionState
-            if AppType.isCustom {
```
BPerlakiH, line 214: This is where we were trying to fix the URL that was already set on the WebView from interactionState data.
Since we fix the URL in the data itself, this is no longer needed.

ViewModel/BrowserViewModel.swift
```diff
@@ -416,7 +401,7 @@ final class BrowserViewModel: NSObject, ObservableObject,
     // RESTORATION
     func restoreByWindowNumber(
         windowNumber currentNumber: Int,
-        urlToTabIdConverter: @escaping (URL?) -> NSManagedObjectID
+        urlToTabIdConverter: @MainActor @escaping (URL?) -> NSManagedObjectID
```
BPerlakiH, line 404: This is to fix a concurrency build warning.

PR #703, opened 2024-03-21
Remove Brewfile.lock.json

### LINKED ISSUE #702: Remove Brewfile.lock from git
This file is not necessary to be included in git.
From the original discussion on HomeBrew:
https://github.com/Homebrew/homebrew-bundle/pull/552
I can conclude that this "lock" file is not intended to work the same way as in other tools,
therefore ignoring it in our case is fine.

BPerlakiH:
Fixes #702


rgaudin (approved):

kelson42:
I'm not confortable -at this stage- to merge without better understanding:
* why it is here in a first place
* why it is a problem to keep it
* why we don't want to rely on fix deps

I will look in details...

BPerlakiH:
> why it is here in a first place
It is generated by the command "brew bundle"
> why it is a problem to keep it
If a new contributor to the repository follows our instructions in Readme, checks out the main branch and run "brew bundle", that will generate a newer version of this file, and it will be most probably checked in with the PR.
> why we don't want to rely on fix deps
It is not working as in other tools, it won't "lock" our dependencies in any form. Eg. on CI this file is overwritten / changed during build... From what I have read, the only benefit of this file is to compare it with the former version to see what has changed, a kind of a "debug"...

Issue #540, opened 2023-11-20
Cmd + click to open link in new tab


BPerlakiH:
Similarly to native safari browser, it would be nice to have cmd + click opening links in new window/tab on macOS.

ishqDehlvi:
How can I reproduce this issue?

kelson42:
@BPerlakiH Can you help here please? Was under the impression it was obvious... but seems its not!

BPerlakiH:
This issue is about the fact that in the traditional macOS applications, (such as Safari browser) it is possible to use the combination of CMD + left mouse click on a link to open it in a new tab.
Currently in our application we do support to open links in a new tab, but in order to do this the user needs to right click on a link and select "open in a new window" which as a result will open the link in a new tab.
We would like to have the very same behaviour but with the shortcut of command + left mouse click.

kelson42:
Fixed by https://github.com/kiwix/apple/pull/708

PR #708, opened 2024-03-22
Detect cmd+click event to open link on new tab

quocduyvu6262:
Fixes (#540)

### DIFF
--- a/ViewModel/BrowserViewModel.swift
+++ b/ViewModel/BrowserViewModel.swift
@@ -219,6 +219,24 @@ final class BrowserViewModel: NSObject, ObservableObject,
         }
     }
 
+    // MARK: - New Tab Creation
+    
+#if os(macOS)
+    private func createNewTab(url: URL) -> Bool {
+        guard let currentWindow = NSApp.keyWindow else { return false }
+        guard let windowController = currentWindow.windowController else { return false }
+        // store the new url in a static way
+        BrowserViewModel.urlForNewTab = url
+        // this creates a new BrowserViewModel
+        windowController.newWindowForTab(self)
+        // now reset the static url to nil, as the new BrowserViewModel already has it
+        BrowserViewModel.urlForNewTab = nil
+        guard let newWindow = NSApp.keyWindow, currentWindow != newWindow else { return false }
+        currentWindow.addTabbedWindow(newWindow, ordered: .above)
+        return true
+    }
+#endif
+    
     // MARK: - WKNavigationDelegate
 
     func webView(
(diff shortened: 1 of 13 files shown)


quocduyvu6262:
@BPerlakiH Hi, I recently joined the project and really appreciate it if you could review this PR. Thanks!

ViewModel/BrowserViewModel.swift
```diff
@@ -227,6 +227,26 @@ final class BrowserViewModel: NSObject, ObservableObject,
         decisionHandler: @escaping (WKNavigationActionPolicy) -> Void
     ) {
         guard let url = navigationAction.request.url else { decisionHandler(.cancel); return }
+
```
BPerlakiH, line 230: Thank you for this, it is indeed working for macOS.
There are a couple of things, that should be addressed:
- This class is re-used for iOS as well, and for that reason this code needs to be within a ``#if os(macOS)`` block, as at the moment it won't compile for iOS.
- We do not use `;` in our codebase, for early returns after `guard` statements.
Theoretically you could do it on one line, like so:
```
... else {
    return decisionHandler(.cancel)
}
```
This is still confusing, and taking advantage of the callback function return type.
The preferred and more explicit way is eg:
```
... else {
    decisionHandler(.cancel)
    return
}
```
Please use this format.

- another issue is that most of this "new tab opening" code has been copy pasted from another function, while the return type is somewhat different, I think, the "new tab creation" can be moved to a new private function, perhaps that could return `Bool`. The advantage of this will be that the code is in 1 place, and the function itself can be wrapped into `#if os(macOS)`.

quocduyvu6262: @BPerlakiH, I have updated it by adding a condition for MacOS and moved the code for creating a new tab into a private function.

BPerlakiH (approved):
Thank you @quocduyvu6262 , it looks good to me.
One last small thing: Please remove the comments marked (suggestions). 
They are no longer needed, since the function name already contains the same.

BPerlakiH (approved):

BPerlakiH:
Look good to me.

kelson42:
@quocduyvu6262 I wanted to merge but can not because current beanch has to be rebased on main branch (no merge!).

quocduyvu6262:
> @quocduyvu6262 I wanted to merge but can not because current beanch has to be rebased on main branch (no merge!).

Hi @kelson42, I have recently rebased this branch on `main` branch. You can try to merge again. Thanks!

BPerlakiH (dismissed):

BPerlakiH (dismissed):

BPerlakiH:
@kelson42 @rgaudin I think we need more approvals for this one in order to build it, and then merge it.

PR #717, opened 2024-03-30
699 double clicking a zim opens multiple windows

### LINKED ISSUE #699: Double-clicking a ZIM opens multiple windows
When double-clicking a ZIM in the Finder (ZIM was not in catalog), the Kiwix app opens with multiple windows:
- one with the homepage of said ZIM ; that's what I expect
- Another on the Reading part, listing my local collection

BPerlakiH:
Fixes #699

### DIFF
--- a/App/App_macOS.swift
+++ b/App/App_macOS.swift
@@ -13,8 +13,15 @@ import Defaults
 import CoreKiwix
 
 #if os(macOS)
+final class AppDelegate: NSObject, NSApplicationDelegate {
+    @MainActor func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
+        true
+    }
+}
+
 @main
 struct Kiwix: App {
+    @NSApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
     @StateObject private var libraryRefreshViewModel = LibraryViewModel()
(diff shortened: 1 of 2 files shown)


App/App_macOS.swift
```diff
@@ -13,8 +13,15 @@ import Defaults
 import CoreKiwix
 
 #if os(macOS)
```
BPerlakiH, line 15: As it turned out this is the official way of closing the application when the last window is closed.

App/App_macOS.swift
```diff
@@ -171,12 +178,6 @@ struct RootView: View {
             navigation.currentItem = .reading
             browser.load(url: url)
         }
```
BPerlakiH, line 180: This was our, unofficial way to achieve the same

Views/ViewModifiers/FileImport.swift
```diff
@@ -79,7 +79,9 @@ struct OpenFileHandler: ViewModifier {
                 openedZimFileIDs.forEach { fileID in
                     guard let url = ZimFileService.shared.getMainPageURL(zimFileID: fileID) else { return }
                     #if os(macOS)
-                    NSWorkspace.shared.open(url)
```
BPerlakiH, line 82: This fixes the multiple windows situation.

kelson42 (approved):
Lgtm

PR #719, opened 2024-03-31
GPLv3 compliant headers

### LINKED ISSUE #575: Replace file License headers to match GLPv3

kelson42:
Fixes #575... and strips trailing white spaces over all files.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,13 @@
+Automactic <chris@kiwix.org>
+BPH <contact@perlakidigital.com>
+rgaudin <reg@kiwix.org>
+Kelson <kelson@kiwix.org>
+translatewiki.net <translatewiki@translatewiki.net>
+tvision251 <tvision251@gmail.com>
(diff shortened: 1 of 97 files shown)


rgaudin:
Header looks good but given how you had to consistently strip white spaces (as we do on other projects), I'm worried that hanging white space are the default or common way to do in Swift. This should be look into first.
You did not mention this change in your PR description 😉

Also, before merging it's important that @BPerlakiH sets up this template he mentioned so new files are created in accordance to this.

kelson42:
@BPerlakiH Please complete directly in this PR with your templating solution.

BPerlakiH:
Updated with the Xcode template and some post action that was required for XcodeGen.
The commenting style was updated to match the format supported by Xcode.

BPerlakiH (approved):

Issue #723, opened 2024-04-08
Unlinked ZIM file can cause invalid tab state on macOS


BPerlakiH:
## Steps to reproduce:
- have an existing and working ZIM file in the app
- open one or more tabs with the content of this ZIM file
- unlink the ZIM file
- close the app
- re-open the app

## The problem:
We are saving the tabs data separately from the ZIM files, and once the ZIM file is unlinked, we try to restore these tabs.
In most cases we will only receive a pop-up, that the content is not available, which is not user friendly, the bigger problem is that in rare cases it can also crash the app.

## Solution:
- Remove the tab data when the user closes the tab. This wasn't fully working so far on macOS, and we were storing lots of unused tabs in the DB
- make sure we delete the tab data, when unlinking the ZIM file
- make sure we do not try to restore a tab from an object that is no longer present in the DB
-

PR #724, opened 2024-04-08
Fix unlinked ZIM file tab data

BPerlakiH:
Fixes: #723

### DIFF
--- a/App/App_macOS.swift
+++ b/App/App_macOS.swift
@@ -186,7 +186,7 @@ struct RootView: View {
             navigation.currentItem = .reading
             browser.load(url: url)
         }
-        .onReceive(tabCloses) { item in
+        .onReceive(tabCloses) { _ in
             guard !navigation.isTerminating else {
                 // tab closed by app termination
                 return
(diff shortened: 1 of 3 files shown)


App/App_macOS.swift
```diff
@@ -21,7 +21,7 @@ import CoreKiwix
 
 #if os(macOS)
 final class AppDelegate: NSObject, NSApplicationDelegate {
-    @MainActor func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
+    func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
```
BPerlakiH, line 24: Here the runtime was complaining that it is not performant to execute this function on the main thread.

App/App_macOS.swift
```diff
@@ -185,7 +186,18 @@ struct RootView: View {
             navigation.currentItem = .reading
             browser.load(url: url)
         }
```
BPerlakiH, line 188: There are 2 ways to close the macOS application:
- CMD+Q  - this will terminate the application
- by closing each and every tab - clicking on X of the tab or pressing CMD+W for each tab/window.
The difference here is that if the user closes a tab, we should delete the data related to it, if we quit the app we should not delete tab data.

kelson42 (approved):
LGTM. tHX

PR #726, opened 2024-04-09
622 corrupted zim leads to crash

### LINKED ISSUE #622: Corrupted ZIM leads to crash
## Steps to repeat

1. Open Kiwix
2. Open wikipedia_en_all_maxi_2023-11.zim
3. Browse wikipedia pages
4. Search for anything
5. It crashes

## Environment

Version: 3.1 (116) installed from App Store
OS: macOS Sonoma 14.2.1 (23C71)
Hardware: 2019 intel Macbook.

## Stack trace

```
Thread 27 Crashed::  Dispatch queue: NSOperationQueue 0x7f9f1f974f00 (QOS: UNSPECIFIED)
0   libsystem_kernel.dylib        	    0x7ff8057c17a2 __pthread_kill + 10
1   libsystem_pthread.dylib       	    0x7ff8057f9f30 pthread_kill + 262
2   libsystem_c.dylib             	    0x7ff805718a49 abort + 126
3   libc++abi.dylib               	    0x7ff8057b2c72 abort_message + 241
4   libc++abi.dylib               	    0x7ff8057a4e1a demangling_terminate_handler() + 266
5   libobjc.A.dylib               	    0x7ff805443476 _objc_terminate() + 104
6   libc++abi.dylib               	    0x7ff8057b20cb std::__terminate(void (*)()) + 6
7   libc++abi.dylib               	    0x7ff8057b2086 std::terminate() + 54
8   libdispatch.dylib             	    0x7ff8056525ae _dispatch_client_callout + 28
9   libdispatch.dylib             	    0x7ff8056550c7 _dispatch_continuation_pop + 523
10  libdispatch.dylib             	    0x7ff80565473c _dispatch_async_redirect_invoke + 710
11  libdispatch.dylib             	    0x7ff805661776 _dispatch_root_queue_drain + 328
12  libdispatch.dylib             	    0x7ff805661f22 _dispatch_worker_thread2 + 152
13  libsystem_pthread.dylib       	    0x7ff8057f6c06 _pthread_wqthread + 262
14  libsystem_pthread.dylib       	    0x7ff8057f5b97 start_wqthread + 15
```

BPerlakiH:
Preparing for handling standard exception raised as part of #622 .

Fixes #622

### DIFF
--- a/SwiftUI/Model/SearchOperation/SearchOperation.mm
+++ b/SwiftUI/Model/SearchOperation/SearchOperation.mm
@@ -52,7 +52,7 @@
     // get a list of archives that are included in search
     typedef std::unordered_map<std::string, zim::Archive> archives_map;
     auto *allArchives = static_cast<archives_map *>([[ZimFileService sharedInstance] getArchives]);
-    
+
     std::vector<zim::Archive> indexSearchArchives = std::vector<zim::Archive>();
     std::vector<zim::Archive> titleSearchArchives = std::vector<zim::Archive>();
     for (NSUUID *zimFileID in self.zimFileIDs) {
@@ -63,9 +63,9 @@
                 indexSearchArchives.push_back(archive);
             }
             titleSearchArchives.push_back(archive);
-        } catch (std::out_of_range) { }
+        } catch (std::exception) { }
     }
-    
+
     // perform index and title search
     [self addIndexSearchResults:indexSearchArchives];
     if (titleSearchArchives.size() > 0) {
--- a/SwiftUI/Model/SearchOperation/SearchOperation.mm
+++ b/SwiftUI/Model/SearchOperation/SearchOperation.mm
@@ -67,7 +67,9 @@
     }
 
     // perform index and title search
-    [self addIndexSearchResults:indexSearchArchives];
+    try {
+        [self addIndexSearchResults:indexSearchArchives];
+    } catch (std::exception) { }
     if (titleSearchArchives.size() > 0) {
         int count = std::max((35 - (int)[self.results count]) / (int)titleSearchArchives.size(), 5);
         [self addTitleSearchResults:titleSearchArchives count:(int)count];


SwiftUI/Model/SearchOperation/SearchOperation.mm
```diff
@@ -49,29 +49,31 @@ - (id)initWithSearchText:(NSString *)searchText zimFileIDs:(NSSet *)zimFileIDs {
...
-        } catch (std::out_of_range) { }
-    }
-    
-    // perform index and title search
-    [self addIndexSearchResults:indexSearchArchives];
-    if (titleSearchArchives.size() > 0) {
-        int count = std::max((35 - (int)[self.results count]) / (int)titleSearchArchives.size(), 5);
-        [self addTitleSearchResults:titleSearchArchives count:(int)count];
-    }
+    try {
```
mgautierfr, line 52: You probably don't want to try/catch the whole loop adding the archive but try/catch each adding of an archive.

Else, the first corrupted zim file with stop the loop.

SwiftUI/Model/SearchOperation/SearchOperation.mm
```diff
@@ -49,29 +49,31 @@ - (id)initWithSearchText:(NSString *)searchText zimFileIDs:(NSSet *)zimFileIDs {
...
+            } catch (std::out_of_range) { }
+        }
+
+        // perform index and title search
+        [self addIndexSearchResults:indexSearchArchives];
+        if (titleSearchArchives.size() > 0) {
+            int count = std::max((35 - (int)[self.results count]) / (int)titleSearchArchives.size(), 5);
+            [self addTitleSearchResults:titleSearchArchives count:(int)count];
+        }
+    }  catch (std::exception) {}
```
mgautierfr, line 76: Sadly, xapian exception are not inheriting standard exception.
So you should also catch [Xapian::Error](https://xapian.org/docs/apidoc/html/classXapian_1_1Error.html)

kelson42: Once you have fixed libzim, it wil!?

SwiftUI/Model/SearchOperation/SearchOperation.mm
BPerlakiH: Thanks, make sense. Please see it updated.

SwiftUI/Model/SearchOperation/SearchOperation.mm
kelson42: This should be handled in https://github.com/openzim/libzim/issues/870

BPerlakiH: Do you think we could stick to the convention of throwing only `std::exception`? 
If  `Xapian::Error` could be caught by the libzim, and re-thrown as `std::exception` that would be the simplest / cleanest solution for client apps on Apple.

kelson42: Yes IMHO

kelson42:
@BPerlakiH Can younplease rebase and fix conflict?

BPerlakiH:
@kelson42 rebased and conflict solved.

SwiftUI/Model/SearchOperation/SearchOperation.mm
mgautierfr: > Do you think we could stick to the convention of throwing only std::exception?
If Xapian::Error could be caught by the libzim, and re-thrown as std::exception that would be the simplest / cleanest solution for client apps on Apple.

Yes, that's the plan with https://github.com/openzim/libzim/pull/873, we will throw a `zim::zim::ZimFileFormatError` (which inherit `std::exception`) or handle the exception internally to returns default value.
When we will do one or the other solution is still open to discussion (and evolution), but it will always be one of the two ways (or a bug in libzim code).

rgaudin:
libkiwix 13.1.0-1 has been released

kelson42:
@BPerlakiH Can you please check if if works now with the latest version of the libzim/libkiwix?

BPerlakiH:
@kelson42 After updating to latest libzim, and applying one more standard exception handler, the search is working with the corrupted wikipedia ZIM file.

PR #727, opened 2024-04-13
EPUBs - Save content of unknown mime-types

### LINKED ISSUE #193: Unable to download EPUBs in Gutenberg ZIM file
*@mikejab commented on Nov 5, 2019, 4:21 PM UTC:*

Hi

I am running IOS 13.2 and kiwix 1.1 on XR 256gb

Downloaded Project Gutenberg 57.32 gb zim.

When I try to download (red down arrow above cpu looking box) nothing happens. If I copy I get something like
kiwix://b85a521d-ac99-2d25-1282-aa7456ecf9dd/I/Pride%20and%20Prejudice.1342.epub
But this is not openable.

This is likely a substantial bug for the iOS version as I have done the following without change

Reinstalled IOS clean
Redownload zim multiple times
Tried on my wife’s XR phone.
Installed most book reader apps

I have reached out in the past but did not get assistance.

Can you assist?

Mike
5123061509

Sent from my iPhone

*This issue was moved by [kelson42](https://github.com/kelson42) from [openzim/gutenberg#92](https://github.com/openzim/gutenberg/issues/92).*

BPerlakiH:
Fixes #193 

## Before
We had the following flow for a kiwix url (eg: kiwix://42EF661C-B328-5AFF-4A1D-41F348378AFE/I/files/9791029307393.epub)
- intercept the web-browser for kiwix scheme
- get the content of this url from the ZIM file (the epub data)
- pass back the content to the web browser
- if the browser could not open it, it was triggering an alert. The same was triggered even if the content was missing

## Changes
- before sending the ZIM content to the browser, make sure it exists, and has greater than 0 size, otherwise trigger the alert immediately, and do not send anything to the browser
- one we get the content on the browser side, but before displaying anything, at this point we can determine if the mime-type of it was "understood" by the browser
- if the mime-type was understood, we can process the request as usual
- if the mime-type was not understood we offer to export the content

## Exporting the content
- update the user permissions to be able to export / write files
### macOS: 
Here it is slightly easier, once we have the content to export, we can display the "save" panel from the OS, and once the user selects where it should be saved, we can write the content there.
In terms of I/O, we write the file only once directly to the place picked by the user (eg: Desktop folder)
### iOS:
It is slightly more complicated, as in order to display / share the content we need to create a temporary file out of it.
In order not to create accidental files just by taping on links, there's an additional alert asking the user if he/she wants to export the content. Once agreed, we create the temp file, show the system dialog, where the user can save the file,
and once that is closed (either with or without success), we delete the temp file.


.gitignore
```diff
@@ -73,6 +73,7 @@ Libraries
 
 # ignore xcode project files, they are now generated by XcodeGen
 *.xcodeproj
+Support/Kiwix.entitlements
```
BPerlakiH, line 76: This file is in fact generated by xcodegen, just wasn't really changing so far.

kelson42:
@BPerlakiH What about opening the file directly with a third party app? Here, and for this kind of UX tickets, would be great to have a small screencast in the PR description so we can figure out how it works.

BPerlakiH:
@kelson42 posted the video recordings to you on slack.

kelson42:
Videos looks good. @rgaudin any remarks? I would propose to merge and test with testflight if it really work like expected?

rgaudin:
> Videos looks good. @rgaudin any remarks? I would propose to merge and test with testflight if it really work like expected?

I agree

PR #728, opened 2024-04-14
714 dimiss fixes part2

### LINKED ISSUE #714: iOS opening the main page from the ZimFile Details don't close the presented sheet
The button does open the main page of the ZIM file selected from the details, but it does so "in the background" on iOS.
This is not intuitive at all.

Solution: if we open the main page, close the presented pop-up on iOS.

Simulator Screenshot - iPhone 15 - 2024-03-25 at 01 54 07

BPerlakiH:
Fixes #714 
There were some additional cases that needed the same treatment as before.


kelson42 (approved):
LGTM

Issue #215, opened 2020-01-15
Impossible to seek on videos


kelson42:
but it works fine on other ports like android. My guess is that the content is not access directly, instead the whole content is downloaded first.

kelson42:
Basically a duplicate of https://github.com/openzim/youtube/issues/91

kelson42:
Youtube bug has been fixed but it still does not work.

kelson42:
Maybe correlated to #324

kelson42:
@automactic @rgaudin AFAIK It works now with recent ZIM files. You confirm?

rgaudin:
Not for me (1.13.6)

rgaudin:
@kelson42 why are you closing this?

kelson42:
@rgaudin With which ZIM file?

rgaudin:
Tried with Oct. 5 mali ZIM. Do you have a working ZIM for me to test?

kelson42:
@rgaudin sorry, you are right.

kelson42:
I refresh that ticket with an update using iOS 16.4.1 and Kiwix 3.0.2.

Basically it still fail always the same manner: clicking on the progress bar does nothin, the video continues at the current offset.

But it works fine with the three main browsers on iOS using http://library.kiwix.org/viewer#mali-pour-les-nuls_fr_all_2023-03/home.html, so I guess this problem has or might have probably its root cause in Kiwix itself.

BPerlakiH:
@kelson42 please see my findings here: https://github.com/kiwix/kiwix-apple/pull/738

kelson42:
My last status here is that everything works fine on iOS with seeking… but we have problems on macOS!

BPerlakiH:
@kelson42 With the latest develop, and Safari 17 (and higher), which is available from (iOS 17, iPadOS 17) macOS Sonoma, macOS Ventura, and macOS Monterey, there are no problem on macOS.
I've update the wiki page with this info as well, please see here:
https://github.com/kiwix/kiwix-apple/wiki/Known-Issue:-Videos

kelson42:
@BPerlakiH This is ticket for Kiwix

BPerlakiH:
@kelson42 from my research it seems that Safari 17.3 comes pre-installed with macOS Sonoma, therefore the WebKit version there should be able to play webm "natively", and video seeking works perfectly on that.

kelson42:
[...]

PR #738, opened 2024-04-24
Seek videos

BPerlakiH:
Relates to: #215 
This PR is a summary of my findings so far, and a minor fix for iOS, by adding the background processing and audio capabilities to the apps.

So far by using the test TED ZIM file for videos, I have found the following:

**iOS:** 
There are only small issues really: 
- when you first load a page with the video, it will start the audio but the screen will be black. If you close the video (fullscreen) and re-open the same video, everything is fine. Also if you toggle to Picture in Picture that fixes it as well.
- I couldn't make it stream to Apple TV via Airplay for some reason
Other than these 2 issues, it works perfectly, you can scrub, mute, change subtitles, use picture and picture, which properly closes if you leave the page, and even resumes when you go back. The playback (audio) can be continued even on the Lock Screen, which is nice. 
Note: strangely, if I opened the link on my phone and started the video in Safari (outside of the Kiwix app), it starts OK without the black screen (sometimes with some initial buffering hick-ups). 
Note: it doesn't work properly on the Simulator (well most audio / video usually less than perfect on simulators anyway), but works great on the device.

See a short demo:
https://github.com/kiwix/kiwix-apple/assets/6784320/089efac7-8881-4955-8a33-9d2b5734947c

**macOS:**
It is somewhat a different story here:
- It starts the video great, there's no black screen, you can scrub, mute, and even control it from the system top tab-bar.
- The problem is: I could not enter full screen mode, no matter how hard I tried.
- It is possible to change subtitles, but the pop-over is really small and cluttered.
- If you right-click on the video, you can "enable" video controls, which will show a 2nd video control behind the current one, which is mostly un-clickable, but looks more like one provided by the system (more native).
- Also in the right click menu there's an option to enter full screen, but that's not working either.
See the screenshots:

right_menu
double_controls

### DIFF
--- a/Support/Info.plist
+++ b/Support/Info.plist
@@ -36,7 +36,9 @@
 	<false/>
 	<key>UIBackgroundModes</key>
 	<array>
+		<string>audio</string>
 		<string>fetch</string>
+		<string>processing</string>
 	</array>
 	<key>UIFileSharingEnabled</key>
 	<true/>


kelson42:
@BPerlakiH Thx for the investigation and fix. I propose to just merge and please make a new tesflight so we can test ourself. Not sure we share all the same UX.

kelson42 (approved):
LGTM

Issue #747, opened 2024-05-03
Relase 3.3.0: Describing use of required reason API


BPerlakiH:
As of Apple:

> Starting May 1, 2024, when you upload a new app or app update, you must include a NSPrivacyAccessedAPITypes array in your app’s privacy manifest to provide approved reasons for these APIs used by your app’s code. For more details about this policy, including a list of required reason APIs and approved reasons for usage, visit: https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api.

PR #748, opened 2024-05-03
Release 3.3.0: describing use of required reason api

BPerlakiH:
Fixes: #747 
Fulfilling Apple requirements on Privacy Manifest file and required fields.
From the docs it seems that macOS builds require the file to be placed into another folder than for iOS.
Currently I managed to use a symlink, so we can edit both platforms from a single place.

Note: After merging this, it should be equally good for custom apps as well.


kelson42 (approved):
LGTM

PR #757, opened 2024-05-08
Target FTP forlder for release

### LINKED ISSUE #756: Fix FTP target folder for macOS releases
Currently the resulting folder is eg: `release/3.3.0/`,
whereas it should be: `release/kiwix-desktop-macos/`

rgaudin:
Fixes #756

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -99,7 +99,7 @@ jobs:
       if: github.event_name == 'release' && matrix.platform == 'macOS'
       run: |
         echo "UPLOAD_TO=ftp" >> $GITHUB_ENV
-        echo "UPLOAD_FOLDER=release/${{ env.VERSION }}" >> $GITHUB_ENV
+        echo "UPLOAD_FOLDER=release/kiwix-desktop-macos" >> $GITHUB_ENV
 
     # Post App Release: skip steps for iOS, by leaving the UPLOAD_TO empty ("")
 
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -14,7 +14,7 @@ on:
   - cron: '32 1 * * *' # NIGHTLY @ 01:32
   - cron: '00 2 * * 1' # WEEKLY @ 02:00 on Monday
   push:
-    tags: 
+    tags:
       - testflight**   # TestFlight Tag
   release:
     types: [published]   # Post App Release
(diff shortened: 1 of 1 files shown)


rgaudin:
Note that this only fixes the folder, not the filename. Awaiting @kelson42's input on that

BPerlakiH (approved):
LGTM

rgaudin:
I'll update filename as well

.github/workflows/cd.yml
```diff
@@ -214,7 +214,9 @@ jobs:
     - name: Upload DMG
       if: env.UPLOAD_TO == 'ftp' && matrix.platform == 'macOS'
       run: |
+        mv ${PWD}/kiwix-${VERSION}.dmg ${PWD}/kiwix-desktop-macos_${VERSION}.dmg
         python .github/upload_file.py --src ${PWD}/kiwix-${VERSION}.dmg --dest ci@master.download.kiwix.org:30022/data/download/${UPLOAD_FOLDER} --ssh-key ${SSH_KEY}
+        mv ${PWD}/kiwix-desktop-macos_${VERSION}.dmg ${PWD}/kiwix-${VERSION}.dmg
```
BPerlakiH, line 219: hmm I am not sure about this one...
we rename it, but the upload script still gets the `kiwix-${VERSION}.dmg` file as the `--src` input.
Maybe it would be enough to do:
```suggestion
        python .github/upload_file.py --src ${PWD}/kiwix-${VERSION}.dmg --dest ci@master.download.kiwix.org:30022/data/download/${UPLOAD_FOLDER}/kiwix-desktop-macos_${VERSION}.dmg --ssh-key ${SSH_KEY}
```

rgaudin: My bad ; fixed it. No we can't pass full destination as the script expects a folder.

BPerlakiH (changes requested):
I think it needs a slight adjustment, as above.

rgaudin:
> I think it needs a slight adjustment, as above.

Which adjustment ?

.github/workflows/cd.yml
BPerlakiH: do we need the 2nd `mv` command after the upload ?

BPerlakiH:
> > I think it needs a slight adjustment, as above.
> 
> Which adjustment ?

I think we were commenting on the same simultaneously.
Q: do we need the re-rename the file after the upload ?

BPerlakiH (approved):

.github/workflows/cd.yml
rgaudin: This filename is a requirement for this very step only. Changing expectations from this step onward seems unjustified. We dont **need** it because it's kind of the last step but it's cleaner IMO.

Issue #767, opened 2024-05-19
Support range requests for video content


BPerlakiH:
Make sure our application is capable of handling range requests for video content.

Background: as of our current investigation the currently working html5 video player (for latest 17.3+ iOS) do not create range requests. These requests might appear at some point, we want to make sure we are prepared for it.

kelson42:
Still the priority, but would like to release 3.4.0 before.

kelson42:
@BPerlakiH @rgaudin Please organise a meeting together to assess current situation and sketch a plan to move forward

rgaudin:
I don't think there's been effort on this direction recently.
To me there is still a technical assessment that must be conducted: how does it work with major browsers over HTTP? How does it work with the WKWebView over HTTP? How does it _work_ in app mode (Custom Scheme and WKWebView)?
This is a mandatory first step IMO to understand what should be changed and why.
I believe browser behavior adapt to environment/network so results might be complex and require peeking at source code.

This implies having a proper testing setup with identical videos of various sizes.

PR #768, opened 2024-05-19
Support range requests for video content

BPerlakiH:
Fixes: #767

### DIFF
--- a/Model/Entities/Entities.swift
+++ b/Model/Entities/Entities.swift
@@ -153,6 +153,22 @@ struct URLContent {
     let start: UInt
     let end: UInt
     let size: UInt
+    let lastModified: Date?
+
+    var rangeSize: UInt {
+        if start == 0 && end + 1 >= size {
+            return size
+        }
+        return end - start + 1
+    }
+
+    func contentRange(from requestedStart: UInt, requestedEnd: UInt) -> String {
+        if requestedStart == 0, requestedEnd == 0 {
+            return "bytes \(start)-\(end)/\(size)"
+        } else {
+            return "bytes \(requestedStart)-\(requestedEnd)/\(size)"
+        }
+    }
 
     var httpContentType: String {
         if mime == "text/plain" {
@@ -160,6 +176,11 @@ struct URLContent {
         } else {
             return mime
         }
+    }
+
+    var eTag: String? {
+        guard let lastModified else { return nil }
+        return "\"\(lastModified.timeIntervalSince1970)\""
     }
 }
 
--- a/Model/Utilities/Date+Extension.swift
+++ b/Model/Utilities/Date+Extension.swift
@@ -1,0 +1,26 @@
+// This file is part of Kiwix for iOS & macOS.
+//
+// Kiwix is free software; you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 3 of the License, or
+// any later version.
+//
+// Kiwix is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Kiwix; If not, see https://www.gnu.org/licenses/.
+
+import Foundation
+
+extension Date {
+    func formatAsGMT() -> String {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "EEE, dd MMM yyyy HH:mm:ss 'GMT'"
+        formatter.locale = Locale(identifier: "en_US")
+        formatter.timeZone = TimeZone(abbreviation: "GMT")
+        return formatter.string(from: self)
+    }
+}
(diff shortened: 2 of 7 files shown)


kelson42:
@BPerlakiH Codefactor should pass.

BPerlakiH:
Please see it updated.

Model/Utilities/WebKitHandler.swift
```diff
@@ -16,19 +16,14 @@
 import os
 import WebKit
 
-/// Skipping handling for HTTP 206 Partial Content
-/// For video playback, WebKit makes a large amount of requests with small byte range (e.g. 8 bytes)
-/// to retrieve content of the video.
-/// As a result of the large volume of small requests, CPU usage will be very high,
-/// which can result in app or webpage frozen.
-/// To mitigate, opting for the less "broken" behavior of ignoring Range header
-/// until WebKit behavior is changed.
```
BPerlakiH, line 25: I did found the root cause of this behaviour. In the original PR the byte range was one off, eg:
```
Byte-Range: 0-5
Content-Length: 5
```
Actually it should be either:
```
Byte-Range: 0-4
Content-Length: 5
```
OR
```
Byte-Range: 0-5
Content-Length: 6
```

Due to this one off error the browser was "kept" downloading the end of the range in chunks about ~8 bytes long.
Once the range was corrected, this side effect no longer exists.

Model/Utilities/WebKitHandler.swift
```diff
@@ -53,37 +48,64 @@ final class KiwixURLSchemeHandler: NSObject, WKURLSchemeHandler {
...
+                }
+            }
+            guard let content = ZimFileService.shared.getURLContent(url: url, start: start, end: end) else {
                 self?.sendHTTP404Response(urlSchemeTask, url: url)
                 self?.stopFor(hash)
                 return
             }
-            self?.sendHTTP200Response(urlSchemeTask, url: url, content: content)
+            if isVideo {
+                self?.sendHTTP206Response(urlSchemeTask, url: url, content: content, start: start, end: end)
```
BPerlakiH, line 95: we only apply the range responses to video type of contents

Support/injection.js
```diff
@@ -30,9 +30,12 @@ let observer = new IntersectionObserver(function(entries) {
 
 // register scroll view to handle heading on top of the page
 window.onscroll = function() {
-	if (document.documentElement.scrollTop <= 0) {
-		window.webkit.messageHandlers.headingVisible.postMessage({id: headings[0].id})
-	}
+    if (document.documentElement.scrollTop <= 0) {
```
BPerlakiH, line 33: This was throwing some errors on certain pages, so I thought won't hurt to fix it.

Model/Entities/Entities.swift
```diff
@@ -161,6 +177,11 @@ struct URLContent {
             return mime
         }
     }
+
```
BPerlakiH, line 180: This currently using the ZIM file's modification date time to create the eTag, which I think a good enough solution for our purposes.

Model/Utilities/WebKitHandler.swift
rgaudin: _Only apply to_ but also _Apply to ALL videos_…

rgaudin (changes requested):
Sorry about the long delay !! 😵‍💫

PR has no description while it introduces changes that seems unrelated to the linked ticket: all that date stuff, is it helping the range requests in anyway? Or is it just general improvements?

Otherwise Code looks OK but it doesn't seem to fix what it intends to: webm video did not trigger the many range request so I'm not looking for improvement there.
MP4 videos on the other hand was triggering the 8b requests… and still is with this branch


Screenshot 2024-05-23 at 17 52 26

You are also returning 206 responses to all videos, regarding of whether a range-request was made. While mostly harmful, this is does not respect HTTP

Model/Entities/Entities.swift
BPerlakiH: My thinking here was, that I looked at what headers are returned in Safari / Caddy when webm is played.
So I wanted to have the very same ones, to make sure it works. One of them was this eTag, which can be generated based on the date the ZIM file was changed.
We can omit this if you think it is not needed.

Model/Utilities/Date+Extension.swift
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Kiwix; If not, see https://www.gnu.org/licenses/.
+
+import Foundation
+
+extension Date {
```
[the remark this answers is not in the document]
BPerlakiH: This is related to the returned response header, eg. as on your screenshot, we have the Response:
```
Last-Modified: Thu, 16 May 2024 11:38:20 GMT
```
in order to have the date of the ZIM file formatted the same way, we need this change.
Again, if you think this is not needed, I can remove it.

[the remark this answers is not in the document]
rgaudin: No it's fine ; just wanted to make sure it was a side thing

Model/Entities/Entities.swift
rgaudin: No harm in keeping it IMO. ETag is supposed to be a uniq version ID for a resource and this implementation is OK in this regard since all resources are single-version made at the same time in a ZIM.
Still it feels like abusing the header to have the same for all resources 😀
Mandates a comment explaining this I think

kelson42:
Waiting that #784 is properly implemented to move forward.

kelson42:
ThisnPR shiukd be rebased and completed has nothing else is blocking.

BPerlakiH:
Closing in favour of https://github.com/kiwix/kiwix-apple/pull/894

Issue #744, opened 2024-05-02
Videos are pre-loading in full


BPerlakiH:
After checking it, and benchmarking, it turned out that the videos (webm) are pre-loaded, when the webpage is loaded in full ( in one http request we load from start to finish).
This is sub optimal, and with larger videos it can cause unecesary waiting time for the user.
The time it takes on macOS, using an external usb-c drive:
```
content loaded: video/webm, 6,1 MB: 173.9137916592881 ms
content loaded: video/webm, 4,5 MB: 130.82945832866244 ms
content loaded: video/webm, 5,7 MB: 157.31854166369885 ms
content loaded: video/webm, 6,6 MB: 193.12754168640822 ms
content loaded: video/webm, 15,6 MB: 422.17312499997206 ms
content loaded: video/webm, 5,7 MB: 160.3915833402425 ms
content loaded: video/webm, 18,9 MB: 525.2033333526924 ms
content loaded: video/webm, 49,3 MB: 1338.0876666633412 ms
content loaded: video/webm, 48,2 MB: 1244.5944999926724 ms
content loaded: video/webm, 49,3 MB: 1278.1917083193548 ms
content loaded: video/webm, 69,6 MB: 1844.8072916653473 ms
content loaded: video/webm, 86,2 MB: 2274.9759999860544 ms
content loaded: video/webm, 49,5 MB: 1296.8016666709445 ms
content loaded: video/webm, 64,8 MB: 1680.7168333325535 ms
content loaded: video/webm, 52,5 MB: 1368.049125012476 ms
content loaded: video/webm, 49,3 MB: 1192.994874989381 ms
content loaded: video/webm, 88,9 MB: 2392.5675000064075 ms
```
From this it is obvious that the loading time, can go over 2 seconds, if the video is larger.

benoit74:
I'm sure it can also go over 2 seconds if you have a poor HDD and/or an ("old") Intel machine.

Did you even included a hour long video in your test set?

It doesn't change anything to the issue aside from pushing emphasis on how important it is to solve this.

BPerlakiH:
# Summary

## Comparing Safari and Chrome browsers on macOS (outside of the app)
With the current html setup we have on the html video tag:
```
data-setup='{"techOrder": ["html5", "ogvjs"],
```
This makes video rendering hand over to html5 in both browsers. This mode is
more or less equal to the a pure html5 video implementation without any javascript (without videjs and videojs-ogvjs).
[...]

PR #769, opened 2024-05-19
Move video content loading off the main thread

BPerlakiH:
On top of: #767
Related to: #744

Creating a more smooth UX, for loading html content, especially large item, such as videos.

The former solution was loading the content (including video content) on the main thread of the application, as part of a synchronised queue. In practice it meant, that the UI became unresponsive while the video was loading.

To solve this, we can read the content on a utility thread, and once completed there, we can send the results back to WebKit.

**Before**: it was possible to block the UI thread with larger videos (that were not pre-loaded yet), even the "rainbow spinning wheel" was visible on macOS.
**After**: the UI remains responsive (the user can eg. scroll in the meantime), no more system spinning wheel, instead the video loading animation is visible, which is a lot better.

### DIFF
--- a/Model/Utilities/WebKitHandler.swift
+++ b/Model/Utilities/WebKitHandler.swift
@@ -18,27 +18,29 @@ import WebKit
 
 final class KiwixURLSchemeHandler: NSObject, WKURLSchemeHandler {
     static let KiwixScheme = "kiwix"
-    private let queue = DispatchQueue.main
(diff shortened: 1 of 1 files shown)


rgaudin:
It's a good thing but that needs to be applied once we've merged the _video_ code in. I'm marking this draft in the mean time so we don't accidentally merge

BPerlakiH:
Implemented separately here: https://github.com/kiwix/kiwix-apple/pull/774

PR #771, opened 2024-05-20
[457] - Immersive reading view

### LINKED ISSUE #457: iOS single tap Full screen mode
Hello and many thanks for the magnificent app! What I miss (and many others users too, no doubt) is the full screen mode activated e.g. with a single tap on the screen. It will make reading much more immersive if one could hide the top and bottom bars when they are not in use. Do not waste screen space, there is not plenty of it even on big phones. And this is a standard feature on most ebook reader apps, should be pretty simple to implement..

TheRealAnt:
Fixes #457 

- Hiding bars when scrolling down WebView & showing when scrolling up.
- Showing bars when device rotates into landscape mode.

Demo:

https://github.com/kiwix/kiwix-apple/assets/22860503/8eaa3a49-779a-4fec-93ff-8147af110526

### DIFF
--- a/Model/Utilities/TargetDevice.swift
+++ b/Model/Utilities/TargetDevice.swift
@@ -1,0 +1,44 @@
+// This file is part of Kiwix for iOS & macOS.
+//
+// Kiwix is free software; you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 3 of the License, or
+// any later version.
+//
+// Kiwix is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Kiwix; If not, see https://www.gnu.org/licenses/.
+
+import Foundation
+import UIKit
+
+enum TargetDevice {
+    case nativeMac
+    case iPad
+    case iPhone
+    case iWatch
+    
+    public static var currentDevice: Self {
+        var currentDeviceModel = UIDevice.current.model
+        #if targetEnvironment(macCatalyst)
+        currentDeviceModel = "nativeMac"
+        #elseif os(watchOS)
+        currentDeviceModel = "watchOS"
+        #endif
+        
+        if currentDeviceModel.starts(with: "iPhone") {
+            return .iPhone
+        }
+        if currentDeviceModel.starts(with: "iPad") {
+            return .iPad
+        }
+        if currentDeviceModel.starts(with: "watchOS") {
+            return .iWatch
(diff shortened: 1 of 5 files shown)


kelson42:
It‘s important to check that his fox has no unwanted side effect on macOS (iPadOS?)

kelson42:
@TheRealAnt I like that the feature is implemented with so few lines of code

Views/BuildingBlocks/WebView.swift
```diff
@@ -108,6 +119,12 @@ class WebViewController: UIViewController {
                 self?.topSafeAreaConstraint?.isActive = false
                 self?.view.topAnchor.constraint(equalTo: webView.topAnchor).isActive = true
             }
+        
+        if parent?.navigationController != nil {
+            compactViewNavigationController = parent?.navigationController
+        } else {
+            debugPrint("compactViewNavigationController not set")
```
BPerlakiH, line 126: Please change (all) ``debugPrint`` to ``os_log`` you can also consider adding a new ``subsystem`` in the `struct Log` for ``WebView``.

BPerlakiH:
@TheRealAnt Currently your changes are limited to iOS / iPad OS only, so they have no effect on macOS, which is good.
There's one thing I noticed immediately, that it also applies the bar hiding on iPad OS, which is something we did not wanted.
On iPad there are multiple things that can affect the amount of scroll needed to display the content: side bar collapsing, split screen mode and so on.
My suggestion is to keep these changes limited to iOS only for now. iPad screens are already large enough, plus as we discussed earlier, there's only 1 bar on iPad:

Simulator Screenshot - iPad (10th generation) - 2024-05-22 at 00 32 41

Views/BuildingBlocks/WebView.swift
TheRealAnt: Thanks, I've updated as per your recommendation.

Model/Utilities/TargetDevice.swift
```diff
@@ -0,0 +1,44 @@
...
+        }
+        if currentDeviceModel.starts(with: "iPad") {
+            return .iPad
+        }
+        if currentDeviceModel.starts(with: "watchOS") {
+            return .iWatch
+        }
+        return .nativeMac
+    }
+}
```
BPerlakiH, line 44: This won't compile for macOS.
We do not use macCatalyst, and UIKit cannot be imported for macOS.
We do not target watchOS at all, so that is not needed.
I would suggest something like this instead:
```suggestion
#if os(iOS)
import UIKit
#endif

enum Device {
    case mac
    case iPhone
    case iPad

    public static var current: Self {
        #if os(macOS)
        mac
        #else
        switch UIDevice.current.userInterfaceIdiom {
        case .pad: return iPad
        case .phone: return iPhone
        default:
            assertionFailure("unrecognised device type: \(UIDevice.current)")
            return iPhone
        }
        #endif
    }
}
```

PR #774, opened 2024-05-24
Move HTML content loading off the UI thread (version2)

BPerlakiH:
Related to: https://github.com/kiwix/kiwix-apple/issues/744

Version 2, without Range Requests changes.

Make UX more smooth, when loading html content, especially large item, such as videos.

The former solution was loading the content (including video content) on the main thread of the application, as part of a synchronised queue. In practice it meant, that the UI became unresponsive while the video was loading.

To solve this, we can read the content on a utility thread, and once completed there, we can send the results back to WebKit.

Before: it was possible to block the UI thread with larger videos (that were not pre-loaded yet), even the "rainbow spinning wheel" was visible on macOS.
After: the UI remains responsive (the user can eg. scroll in the meantime), no more system spinning wheel, instead the video loading animation is visible, which is a lot better.

### DIFF
--- a/Model/Utilities/WebKitHandler.swift
+++ b/Model/Utilities/WebKitHandler.swift
@@ -121,11 +121,12 @@ final class KiwixURLSchemeHandler: NSObject, WKURLSchemeHandler {
 
     @MainActor
     private func sendHTTP404Response(_ urlSchemeTask: WKURLSchemeTask, url: URL) {
-        guard isStartedFor(urlSchemeTask.hash) else { return }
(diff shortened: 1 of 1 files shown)


kelson42:
@rgaudin @BPerlakiH Sorry, but not sure anymore what has been decided here, can we review/merge straight without #744?

BPerlakiH:
@kelson42 yes it can be merged separately as an improvement for loading the html content more smooth, if there is video content within.

rgaudin (approved):
LGTM

Issue #773, opened 2024-05-23
Streaming uncompressed data (with direct access)


rgaudin:
What we do on Android for videos doesn't seem possible directly. It's a combination of tricks that are missing on iOS (at least I did not find them easily).
First the `AssetFileDescriptor` allows to mimic an asset (so a clearly bound piece of data) from a filepath, an offset and a size.
Then, there is the integration between Android `MediaPlayer` and the `WebChromeClient` renderer (with some hacks to transition smoothly).

I found some other people inquiring for `AVPlayer` and `WKWebvierw` integration but without answer. Maybe the wording or the keywords were wrong…

Despite this, I had some success with an easy tweak to the code: streaming the data to the client. Currently, we request the requested content from libzim, store that in a variable and put that on the response. On the renderer side, this content is read (sometimes not completely) and used.

This is OK for relatively small content, or maybe even large ones that are consumed entirely by the client (RAM will be required) but in the video use case, we know that the client wont even try to read the whole thing and will display it piece by piece.

Doing so is as easy as repeatedly calling `urlSchemeTask.didReceive(additionalData)`…

```swift
let response = HTTPURLResponse(
    url: url,
    statusCode: statusCode,
    httpVersion: "HTTP/1.1",
    headerFields: headers)
urlSchemeTask.didReceive(response!)

...

for i in 1...nbStreams {
    partEnd = partStart + streamThreshold
    content = ZimFileService.shared.getURLContent(url: url, start: partStart, end: partEnd)
    urlSchemeTask.didReceive(content!.data)
    partStart = partEnd
}
if (finalBytes > 0) {
    content = ZimFileService.shared.getURLContent(url: url, start: partStart, end: partStart + finalBytes)
    urlSchemeTask.didReceive(content!.data)
}
urlSchemeTask.didFinish()
```

This is very efficient in keeping RAM usage under control on very large videos.

I don't know exactly how it works internally but simply looping on writing 2MB chunks does the trick so I suppose renderer-reading is synced somehow.

---

Another improvement that is independent from this is reading video files directly from the filesystem. Leveraging `item.getDirectAccessInformation()` which returns the ZIM path on the fs and the offset at which the content start, we can easily read the video data from it (we already know its size).

[...]

PR #778, opened 2024-05-29
773 streaming uncompressed data

BPerlakiH:
Fixes: #773 

Reading the content in "chunks".
Both direct and indirect / ZIM streaming data solution is added.
Currently the ZIM streaming is not in use, until the offset issue is resolved on the libZIM side (see https://github.com/openzim/libzim/issues/886).

There are still issues with syncing the urlScheme closing, and maybe it's worth to add a Task cancelation, when the user leaves the currently loading page.

### DIFF
--- a/Model/Entities/Entities.swift
+++ b/Model/Entities/Entities.swift
@@ -30,6 +30,7 @@ class Bookmark: NSManagedObject, Identifiable {
     class func fetchRequest(
         predicate: NSPredicate? = nil, sortDescriptors: [NSSortDescriptor] = []
     ) -> NSFetchRequest<Bookmark> {
+        // swiftlint:disable:next force_cast
         let request = super.fetchRequest() as! NSFetchRequest<Bookmark>
         request.predicate = predicate
         request.sortDescriptors = sortDescriptors
@@ -50,12 +51,14 @@ class DownloadTask: NSManagedObject, Identifiable {
     @NSManaged var zimFile: ZimFile?
 
     class func fetchRequest(predicate: NSPredicate? = nil) -> NSFetchRequest<DownloadTask> {
+        // swiftlint:disable:next force_cast
         let request = super.fetchRequest() as! NSFetchRequest<DownloadTask>
         request.predicate = predicate
         return request
     }
 
     class func fetchRequest(fileID: UUID) -> NSFetchRequest<DownloadTask> {
+        // swiftlint:disable:next force_cast
         let request = super.fetchRequest() as! NSFetchRequest<DownloadTask>
         request.predicate = NSPredicate(format: "fileID == %@", fileID as CVarArg)
         return request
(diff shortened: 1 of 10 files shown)


Model/Entities/Entities.swift
```diff
@@ -134,33 +137,46 @@ class Tab: NSManagedObject, Identifiable {
...
     var httpContentType: String {
         if mime == "text/plain" {
             return "text/plain;charset=UTf-8"
         } else {
             return mime
         }
     }
+
+    var isVideo: Bool {
+        mime.contains("video")
```
rgaudin, line 167: Should test if it starts with `video/`

Model/Utilities/WebKitHandler.swift
```diff
@@ -81,52 +81,112 @@ final class KiwixURLSchemeHandler: NSObject, WKURLSchemeHandler {
...
-        guard let content = await readContent(for: url) else {
+        guard let metaData = await contentMetaData(for: url) else {
             sendHTTP404Response(urlSchemeTask, url: url)
             return
         }
-        sendHTTP200Response(urlSchemeTask, url: url, content: content)
 
+        let dataProvider: any DataProvider<URLContent>
+        let rangeSize: UInt // The size of the data chunks used for streaming data
+        if metaData.isVideo, let directAccess = await directAccessInfo(for: url) {
```
rgaudin, line 91: There is no fallback should directAccessInfo return an empty pair. You wisely handled the nul pair in `getDirectAccessInfo` but you're not leveraging it here.
Here we could gracefully fallback to the `ZimContentProvider` should the directAccessInfo be `nil`

rgaudin (changes requested):
> There are still issues with syncing the urlScheme closing, and maybe it's worth to add a Task cancelation, when the user leaves the currently loading page.

Well, better do this now because it will get more complex once you had range requests handling…

Model/Utilities/WebKitHandler.swift
BPerlakiH: Hmm, I am not sure if I understand your point fully, I think what you are suggesting is already handled.

In swift: `if let directAccess = <some optional>` will only be `true` when the optional is not `nil`, if it's true, it will assign the value to the variable: `directAccess`  at the same time.
Whereas `if condition == true, otherCondition == true` is the same as: `if condition == true && otherCondition == true`.
If it is `nil` (or it's not a video), it will go into the `else` branch, where we have the `ZimContentProvider` (and in this `else` scope the variable `directAccess` is not available).

After thinking about this a little more, I made some changes.
I moved the validation to the optional (and only) init of the `DirectAccessInfo`. This makes sure we cannot create an invalid instance of it.
On the other hand, I think validating only on the `path` should be enough.
If we have a valid path with offset=0, that should be OK.
On the other-hand if we have an empty path regardless of the offset value, it will be invalid, as we cannot open the file.

Model/Entities/Entities.swift
BPerlakiH: Please see it updated.

rgaudin (approved):
LGTM now. Let's discuss tomorrow about what to do regarding the libzim issue

BPerlakiH:
With the latest changes we are reading directly for media type (both audio and video),
whereas for uncompressed content we use the full range, so no reading "in chunks".

kelson42:
@BPerlakiH CI does not pass, and would be better to have a git log cleaner

kelson42:
@BPerlakiH Does it fix #744 too?

rgaudin (approved):
LGTM ; albeit the CI and the git history

BPerlakiH:
Yes, it is fixing #744 too.
I have update the unit tests as well, and squashed down the logs.

Issue #780, opened 2024-05-30
`mailto:` links are not opening the OS-configured mail application


benoit74:
ZIM: https://mirror.download.kiwix.org/zim/.hidden/dev/fas-military-medicine_en_2024-05.zim
Kiwix app: 3.4.0 (161) - Testflight
OS: MacOS Sonoma 14.5

When clicking on `Steve Aftergood` at the bottom of the front page, we should open a `mailto:` link. This does not happen (while it does work on Firefox on same machine, so handling of mailto links seems to be properly configured at OS level (is it an OS-wide configuration?).

kelson42:
Tested with latested tesflight, and once i have accepted to open external link, the whole app crash

PR #792, opened 2024-06-07
Change "mailto:" urls to be external

BPerlakiH:
Fixes: #780

### DIFF
--- a/App/App_iOS.swift
+++ b/App/App_iOS.swift
@@ -50,7 +50,7 @@ struct Kiwix: App {
                 .onOpenURL { url in
                     if url.isFileURL {
                         NotificationCenter.openFiles([url], context: .file)
-                    } else if url.scheme == "kiwix" {
+                    } else if url.isKiwixURL {
                         NotificationCenter.openURL(url)
                     }
                 }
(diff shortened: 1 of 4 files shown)


kelson42:
@BPerlakiH Is that the right logic!? Everything which is not zim:// should be external IMHO.

BPerlakiH:
> @BPerlakiH Is that the right logic!? Everything which is not zim:// should be external IMHO.
> 
> We have a few other schemes, for example, tel: geo: ftp:...

I did found that we handle "geo" scheme separately. Should we say that everything apart from "kiwix" and "geo" is external then ?

BPerlakiH:
@kelson42 Please see it updated.

kelson42:
@BPerlakiH This looks more what I was expacted. From a user perspective, its important the OS tries to open the link with the appropriate app, in the case of mailto: the default mail app.

Here a few remarks which are related to the naming of things:
* We should have a zim:// protocol, not a kiwix:// protocol, can we change that in a dedicated issue/PR?
* I challenge the "external" naming, probably "unsupported" would be better?

Issue #793, opened 2024-06-07
Supporting entries ending in `/`


rgaudin:
There is nothing in the [ZIM spec](https://wiki.openzim.org/wiki/ZIM_file_format#Paths) that prevents using entries ending in `/` and thus we ended up in #779 with a ZIM (quartierschinois) that has a `/`-ending main page.

Nothing prevents [URLs](https://en.wikipedia.org/wiki/URL) to end in slash neither.

One popular convention though is that the path is composed of segments separated by `/` and the last part is the document. As such, a `/`-ending path can be considered equivalent to the same path without that trailing slash.
This convention is still implemented in some web servers, browsers, and libraries.

A swift [`URL`](https://developer.apple.com/documentation/foundation/url) has a `path` property that we use all over the code base to extract the ZIM entry path.

In the swift version we use, `URL.path` removes that trailing slash, breaking lookup of ZIM entries with `/`-ending path.

In later version of swift (macOS 13+), Apple added an `URL.path()` that keeps the trailing slash and works as intended with those entries.

What should we do?

1. Consider `/`-ending ZIM entries incorrect and fix ZIM files to not use them?
1. Fix by using `URL.path()` and bump our support from macOS12+ to macOS13+ and iOS15+ to iOS16+ ?
1. Fix for all versions by _normalizing_ path ourselves (`URL.hasDirectoryPath` tells us if URL ends in `/`) ?

I'm in favor of the latter, until we drop those versions support at the end of the year.

kelson42:
Fix 1 if possible in good conditions, not because this is wrong, but because this can be a source of trouble without bringing clear advantage.

AND

Fix 2 because Kiwix should work properly and macOS15 will be released within a few month (so we can just bump the version of Swift... but maybe better in 3.5.0.

benoit74:
Fix 1 is not easy in good conditions in warc2zim. It would means removing the trailing `/` in all entries which have one (quite possible) but it means we have a risk of conflict, probably rare but mostly impossible to solve. It is perfectly legit for a website to have two different resources at two different URLs differing only by a trailing `/` (e.g. `www.example.com/myresource` and `www.example.com/myresource/`).

Since ZIM entries are just a "string", mostly mapped to a URL, and URLs ending with a slash are legit, I don't get why we would impose more constraints that necessary on ZIM entries.

benoit74:
[...]

PR #797, opened 2024-06-08
793 supporting entries ending in "/"

BPerlakiH:
Fixes: #793 

Made sure that we suport urls to contentPath ending with "/",
we drop if there's any "#" symbol with additional arguments there (which was also failing).
Fixed the issue with the error pop-up, now only a failing "main" document should trigger the pop-up.

Tested with "Mes quartiers chinois":

Screenshot 2024-06-08 at 15 06 08

### DIFF
--- a/Model/Utilities/String+Extension.swift
+++ b/Model/Utilities/String+Extension.swift
@@ -57,6 +57,11 @@ extension String {
         )
     }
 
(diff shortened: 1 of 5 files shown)


kelson42:
I guess this PR is blocked / depends on #796 ?!

rgaudin (dismissed):

Issue #772, opened 2024-05-22
After resuming app (iOS) video displayed black


kelson42:
- Watching a video
- Pausing
- Starting another app
- Coming back to Kiwix and getting

<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/a655f8cd-fa4f-4088-bd9b-f56f5192f08e">

Video should display proper frame.

BPerlakiH:
I assume it's on an iPad right?

I could not recreate this issue on iPad OS 17.4.1, with the current main branch (as of 30 May).
I did tried to background the app, open another one, I even tried split view mode, but the video is playing as expected.

kelson42:
> I assume it's on an iPad right?

Yes, iPadOS 17.5.1

BPerlakiH:
Ok, I was able to reproduce it, by opening another high consumption app.

kelson42:
@BPerlakiH This issue is the next one to fox and then #768 should be completed

kelson42:
Still buggy, not even after hinübernating, just after going shortly in background:
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/0fc7f505-ae34-4bcd-89fd-03388131fcea">

then going breifly to background
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/16b8eee3-f62d-4318-a66c-f09dfa02f8bf">

BPerlakiH:
@kelson42 I could not re-create this issue on my iPad (with the current main branch), what was the video that you were using ?

kelson42:
> @kelson42 I could not re-create this issue on my iPad (with the current main branch), what was the video that you were using ?

https://mega.nz/file/XRE2VBrI#U0sPZQ6dttL4alTnuQz6rc3e7G2lEr588hMENer-XUE

PR #801, opened 2024-06-09
772 after resuming app ios video displayed black

BPerlakiH:
Fixes: #772 

Note: So far I only found this "hacky" javascript solution to fix this issue.
The traditional "low memory" warning triggers are not called in this case, therefore this black screen side effect is not due to that.
By playing and pausing the video (one after another) fixes this.
The fix is applied if:
- we are not in picture in picture mode - here we do not have this problem
- we are not at the beginning of the video - here we also do not have this problem, plus if we apply it as we come back to the tab., it will change the "poster" of the video to the first frame of the video (which might be black)

### DIFF
--- a/Support/injection.js
+++ b/Support/injection.js
@@ -69,7 +69,7 @@ function refreshVideoState() {
     if (document.pictureInPictureElement != null) {
         return;
     }
-    $("video").each(function(i) {
+    $("video").each(function(_) {
         if (this.paused && this.currentTime > 0) {
             this.play();
             this.pause();
--- a/Support/injection.js
+++ b/Support/injection.js
@@ -69,7 +69,7 @@ function refreshVideoState() {
     if (document.pictureInPictureElement != null) {
         return;
     }
-    $("video").each(function(_) {
+    $("video").each(function() {
         if (this.paused && this.currentTime > 0) {
             this.play();
             this.pause();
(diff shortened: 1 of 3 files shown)


BPerlakiH:
User impact: it is not fixing it immediately, there's a slight delay after re-opening the application, so the black screen is visible for a moment, after that it fixes the issue.

kelson42:
@BPerlakiH There is way to trigger this earlier than ondisplay()? For example when the app gets out of hibernation?

BPerlakiH:
Actually, I had to wrap the call ińto a "Task" so it executes slightly later (next run loop), as the "black screen" problem appears somewhat later than "onAppear".

kelson42:
Hmmm, Imguess we can still do better. Video.js has a feature to set a "poster" overlay. Would that not be a solution to capture the frame a put it on the top, so, even if video goes black, this will be hidden in the background. See https://stackoverflow.com/questions/27583937/how-can-i-make-a-css-glass-blur-effect-work-for-an-overlay/27592975#27592975

BPerlakiH:
Yes, possibly we need further investigation why this is happening in the first place...
We have poster set at the start of the video, maybe we can use something similar.
I did checked the console output, but there are no errors no output whatsoever ever, when the video stops (eg. due to another streaming app).
Maybe there's a way to capture that moment in JS, and then we could do some fallback / overlay solution.

kelson42:
I'm a bit surprised about this, but it seems there is no way tomdo better. Merging.

Issue #809, opened 2024-06-15
macOS: Side tab should update after unlinking a ZIM file


BPerlakiH:
On macOS we have a 3 column layout for the "Opened" ZIM files.
Once we unlink a ZIM file all columns should update.
Currently the 3rd column remains in an undesired state, after a ZIM file is removed.
It gets even more dangerous after the last ZIM file is removed, in certain rare cases it can even crash the app (as we are still referencing the ZIM file which is no longer available).

https://github.com/kiwix/kiwix-apple/assets/6784320/805959e8-3040-45da-b69b-8d14c8ad92d5

Apart from this the message should be in plural (`detail` => `details`):

> Select a ZIM file to see detail**s**

PR #810, opened 2024-06-15
Fix macOS unlink last ZIM file issue

BPerlakiH:
Fixes: #809 

https://github.com/kiwix/kiwix-apple/assets/6784320/3e3aed36-c656-4fc2-b1d1-a6d242485035

### DIFF
--- a/Views/Library/ZimFilesOpened.swift
+++ b/Views/Library/ZimFilesOpened.swift
@@ -25,6 +25,7 @@ struct ZimFilesOpened: View {
         animation: .easeInOut
     ) private var zimFiles: FetchedResults<ZimFile>
     @State private var isFileImporterPresented = false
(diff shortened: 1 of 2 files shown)


Views/Library/ZimFilesOpened.swift
```diff
@@ -46,6 +47,9 @@ struct ZimFilesOpened: View {
                 Message(text: "zim_file_opened.overlay.no-opened.message".localized)
             }
         }
+        .onChange(of: zimFiles.count) { _ in
+            viewModel.selectedZimFile = zimFiles.first // makes sure we also nil out, if all ZIMs were unlinked 
+        }
```
BPerlakiH, line 52: Explanation: `zimFiles.first` returns an optional, so it covers both cases, when there's still files to pick from, and when there's nothing left.

Issue #789, opened 2024-06-01
ZIM metadata „illustration“ is not read properly


kelson42:
Here we have a proper logo (with Kiwix Server)
https://dev.library.kiwix.org/#lang=&q=Mali+mp4

But the same file in the Kiwix local library displays the defaukt logo placeholder
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/6f0c83e5-76b7-45b7-be0c-e5934d21abbb">

kelson42:
Not sure how this is done, but there is a primitive in libkiwix which shiukd be used:
https://libkiwix.readthedocs.io/en/latest/api/classkiwix_1_1Book_1_1Illustration.html?highlight=Illustration

I suspect that because the logo of the ZIM is accessed directly and because we have changed in the path the URL, then something is wrong. Please read https://wiki.openzim.org/wiki/Metadata

rgaudin:
The reader is using its [`url`](https://libkiwix.readthedocs.io/en/latest/api/classkiwix_1_1Book_1_1Illustration.html#class-documentation) property

```cpp
book->getIllustrations().at(0)->url
```

Documentation is of no help here but a quick glimpse at the code have me believe that `url` is only set when read from XML or from OPDS so in apple reader's case, only if downloaded from the catalog. I could confirm this with a test.

@BPerlakiH, we should use `getData()` indeed.
A quick hack could be returning a base64 image/png url as this is used as an URL 😝

kelson42:
Makes sense, metadata should be retrieved from online catalogue AFAP, but if no entry there, ALL metadata (illustration included) should come from the ZIM file.

kelson42:
Still not working:
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/acc3880f-0cc0-4a50-8d77-e33119e43509">

kelson42:
After iOS update to version 18... it works!
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/5350c2c6-f416-4539-bc53-83fe80b619ca">

I'm a bit puzzled.

rgaudin:
Given you use this device to test the app, I'm not sure it's a wise choice to be on a beta system… but there's no downgrade possible so it's too late now 😀

kelson42:
[...]

PR #811, opened 2024-06-16
Fix favIcons of opened ZIM files (not added via catalog download)

BPerlakiH:
Fixes: #789


kelson42:
@BPerlakiH CI does not pass

kelson42:
@BPerlakiH What isnthenlogic of usage between the remote URL and the local illustration?

BPerlakiH:
@kelson42 I have updated the code, it seems that empty string to empty data was not handled properly.
What isnthenlogic of usage between the remote URL and the local illustration?

kelson42:
AFAP usenonline catalogue metadata (even in local library) and otherwise the local (from ZIM file).

BPerlakiH:
If the url is set, that will be displayed, if not it will fallback to the image data, so that covers the scenario you mentioned.

Issue #814, opened 2024-06-17
Export of a PDF fails


kelson42:
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/d89b093e-7bf8-441b-ac61-866b73488160">

If I display a PDF on iOS and then want to export it to the fs, I get a white page PDF.

kelson42:
instrongly suspect that the current system works only for HTML content... but actually if anything else is displayed, instead of generating a PDF, the file should probably be saved/exported straight in its own format.

kelson42:
This is still buggy IMHO, trying with a video:
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/360d662b-be22-4274-8c2b-3d8d0f18edd5">

then exporting to fs....
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/8a278692-8530-4828-b91a-52490a8d9439">

Like you see it tries to save as PDF althouh it should save as WEBM

kelson42:
@BPerlakiH Please fix bugs in generic manner.

PR #820, opened 2024-06-19
Fix share/export content depending on it's type

BPerlakiH:
Fixes: #814

### DIFF
--- a/ViewModel/BrowserViewModel.swift
+++ b/ViewModel/BrowserViewModel.swift
@@ -95,6 +95,7 @@ final class BrowserViewModel: NSObject, ObservableObject,
     // MARK: - Lifecycle
 
     @MainActor
+    // swiftlint:disable:next function_body_length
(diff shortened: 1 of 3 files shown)


ViewModel/BrowserViewModel.swift
```diff
@@ -158,15 +159,25 @@ final class BrowserViewModel: NSObject, ObservableObject,
         }
     }
 
+    func pdfData() async -> Data? {
+        if metaData?.isTextType == true {
```
rgaudin, line 163: This mandates a comment. I believe this triggers PDF generation of the displayed page if the entry's type is text (HTML, etc). Right? Otherwise it returns the data from libzim so this returns binary anyway I guess

rgaudin (dismissed):
LGTM ; with a comment :)

kelson42:
@BPerlakiH LGTM, but please fix CodeFactor error

kelson42 (approved):
LGTM

Issue #816, opened 2024-06-17
Losing capacity to focus in a ZIM


kelson42:
Downloading this file:
https://library.kiwix.org/#lang=&q=water+preppers&category=

Opening homepage and I'm able to entern a search pattern in the top right filter box. But once the focus leaves the input box, I'm not able anymore to focus anything!

It is not a problem with the ZIM because it works fine on library.kiwix.org.

kelson42:
@BPerlakiH This is a very critical issue as it makes a full ZIM almost not usable

Issue #815, opened 2024-06-17
ZIM is not vertically alligned properly


kelson42:
Downloading this file:
https://library.kiwix.org/#lang=&q=water+preppers&category=

Loading in kiwix serve:
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/7d4a3456-6dc4-4926-8a85-9c5dc13496ff">

Loading in Kiwix for iOS:
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/95cc7012-4a7a-4a90-bea7-031c7f1902b8">

There is somekind of top white margin appearing only on Kiwix for iOS bit not in kiwix serve. we should not have this top margin

kelson42:
Same problem with an other ZIM
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/24753758-4dd4-400f-a332-b505bfc79e6b">

PR #823, opened 2024-06-20
Fix offset and animation issues

BPerlakiH:
Fixes: #815 
Fixes: #816 

It also fixes some other scroll issues.

## it was not possible to reveal the bars in certain cases:

https://github.com/kiwix/kiwix-apple/assets/6784320/9a36f06c-5ce5-457c-b8d7-a72da21fbf4e

## it was not possible to get out of a screen (forever flickering):

https://github.com/kiwix/kiwix-apple/assets/6784320/392015f2-3bf4-4113-b3bc-a00e059ce35d

### DIFF
--- a/Views/BuildingBlocks/WebView.swift
+++ b/Views/BuildingBlocks/WebView.swift
@@ -68,9 +68,11 @@ final class WebViewController: UIViewController {
     private let webView: WKWebView
     private let pageZoomObserver: Defaults.Observation
     private var webViewURLObserver: NSKeyValueObservation?
-    private var lastOFfset: CGFloat = 0.0
+    private var lastOffset: CGFloat = 0.0
     private var topConstraint: NSLayoutConstraint?
+    private var bottomConstraint: NSLayoutConstraint?
     private var urlObserver: NSKeyValueObservation?
+    private var animStart: Double = 0
 
     init(webView: WKWebView) {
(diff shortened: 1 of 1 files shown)


kelson42:
@BPerlakiH Thank you for the patch. Could you please explain briefly why the bug? Why only on iPad? How the solution looks like, because all of this is not clear to me.

BPerlakiH:
Briefly it is connected to the scroll down to hide bars functionality: https://github.com/kiwix/kiwix-apple/pull/771
Which we applied only for iOS, but it meant to be really fully applicable only for iPhones. Whereas the implementation is shared between iPhone and iPads, it had some side effects for iPads, that is now fixed.

The offset of the initial screen appeared, due to this extra call we had:
```
webView.setValue(true, forKey: "_haveSetObscuredInsets")
```
which is in collision with hiding / showing the bars.

The other issue was that hiding / showing the bars in itself can cause a scroll event, which in return might toggle the bars again, and we can end up in an endless loop (as on the video).
In short solution to that: the hide / show of the bars is animated. We won't apply showing / hiding the bars until the previous animation is still running.

Another small fix is that, when the bottom bar is displayed, that is taken into consideration for laying out the web-view content (which wasn't the case previously).

rgaudin (approved):
LGTM

PR #824, opened 2024-06-21
Fix custom js

BPerlakiH:
Fixes: #816

As a matter of fact the base PR is fixing the real issue:
https://github.com/kiwix/kiwix-apple/pull/823

since we are on this, it's better to fix the custom JS code as well.


rgaudin (dismissed):
LGTM ; please rename commit so we don't loose the why we've added this

BPerlakiH:
Replaced with: https://github.com/kiwix/kiwix-apple/pull/825

Issue #817, opened 2024-06-17
Bookmarking a PDF fails


kelson42:
See:
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/d2015058-6a27-44f3-9ae9-93b453f47fd5">

The bookmarkmtitle is wrong, if no entry title is available it should be the entry path...

kelson42:
@BPerlakiH You should not need a PDF parser if you don't need an HTML parser. This is also not what os requested in the issue.

PR #826, opened 2024-06-22
Fix bookmarking of PDFs

BPerlakiH:
Fixes: #817 

Currently this is how it looks with all the Water docs bookmarked on macOS:

Screenshot 2024-06-22 at 14 25 39

### DIFF
--- a/ViewModel/BrowserViewModel.swift
+++ b/ViewModel/BrowserViewModel.swift
@@ -608,15 +608,20 @@ final class BrowserViewModel: NSObject, ObservableObject,
             let bookmark = Bookmark(context: context)
             bookmark.articleURL = url
             bookmark.created = Date()
-            if let parser = try? HTMLParser(url: url) {
-                bookmark.title = parser.title ?? ""
-                bookmark.snippet = parser.getFirstSentence(languageCode: nil)?.string
-                guard let zimFileID = UUID(uuidString: url.host ?? ""),
-                      let zimFile = try? context.fetch(ZimFile.fetchRequest(fileID: zimFileID)).first else { return }
-                bookmark.zimFile = zimFile
+            guard let zimFileID = UUID(uuidString: url.host ?? ""),
+                  let zimFile = try? context.fetch(ZimFile.fetchRequest(fileID: zimFileID)).first,
+                  let metaData = ZimFileService.shared.getContentMetaData(url: url) else { return }
+
+            bookmark.zimFile = zimFile
+            if metaData.isTextType,
+               let parser = try? HTMLParser(url: url) {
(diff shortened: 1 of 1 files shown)


kelson42:
@BPerlakiH You should not need a PDF parser if you don't need an HTML parser. This is also not what os requested in the issue. I really don't get it.

BPerlakiH:
@kelson42 The bookmarks are currently containing these fields we do display:
- title 
- (optional) short snippet
- (optional) image url

So far we were getting these fields only via an HTML parser. In case of a PDF file it was not working as expected. Therefore I thought we might use a PDF parser to get the same fields if possible.

BPerlakiH:
In case of a non html (text) document, I can simplify it to be:
- title replaced with the path
- no snippet
- no image

kelson42:
@BPerlakiH Indeed, but not sure what you mean?

kelson42:
yes, this is what is requested in the issue and this is appropriate to me.

So, as far as I understand you wanted to get a better bookmark by allowing to get a better bookmark. Can you please:
* open a dedicated issue to explain the problem and proposal
* move the appropriate code in a draft PR

benoit74:
As discussed for nautilus and zimit, I think it makes much more sense to enhance scrapers so that they populate properly the title and provide proper indexing data for search. This would allow all readers to benefit from such an enhancement at once. I don't mind if this is implement in apple reader, but it might soon be "obsolete" once all pdfs have a proper title due to implementation in the scraper. I intend to add this to the python-scraperlib in the coming weeks.

kelson42:
@benoit74 At the core of this issue is a lack of PDF support at scraper size. But, even if this has to be fixed there, it has to be handled properly at reader level if not title metadat. For PDF or any other supported mime-type.

BPerlakiH:
I've updated this PR, and created a new issue for the more detailed solution: https://github.com/kiwix/kiwix-apple/issues/827

kelson42 (approved):

rgaudin:
I'm a bit worried by this PR. I feel there's no room for discussion and everything is rushed.

My first opinion when reading the PR description was the one of @kelson42 but reading the implementation, I think it was the appropriate way to go: PDF is a natively supported format on apple system. Hence the PDF Parser is builtin.
The PR had the appropriate fallback so ZIM entry so it was respecting the same concept as for HTML: if document itself has a title, use it (as in a regular browser or pdf reader) and if not fallback to ZIM.

kelson42:
@rgaudin See [my comment](https://github.com/kiwix/kiwix-apple/issues/827#issuecomment-2184074829) to the dedicated ticket.

Issue #818, opened 2024-06-17
Under certain condition opening ZIM (homepage) gives a black screen


kelson42:
See for example:
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/4b62ab74-2ff2-403c-8202-89dfbec068d9">

No reproduction steps so far

kelson42:
Actually I have a reproduction case, it happens if I try to reopen (after closing its tab) a ZIM file if Kiwix went on hibernation.

kelson42:
@BPerlakiH @mgautierfr @rgaudin To me here we face a similar root cause like #772. I believe the core of the problem might be that if the app is hibernated and then resumed, the filedescriptors are lost. Could we try to confirmnthis theory?

BPerlakiH:
So far I have found that if the Kiwix app is in background, and there's a genuine low memory error on device, after which foregrounding the Kiwix app, it will have this error, when we try to re-open a tab.

kelson42:
> So far I have found that if the Kiwix app is in background, and there's a genuine low memory error on device, after which foregrounding the Kiwix app, it will have this error, when we try to re-open a tab. 

Which error? What would be your proposal?

PR #834, opened 2024-06-28
Partial revert of 78b5f00, to fix webview black screen

BPerlakiH:
Fixes: #818 

After many experiments I haven't found anything useful, as the UI layer was crashing, and XCode could not debug it anymore...
So I investigated with earlier versions of the app, and after some git bisect it turned out that the commit 78b5f00 was the one to blame.
I still think it is not ideal, in terms of the code, but at the moment this is the very best version, that has no side effects.

I have tested it on all devices, and seems to work as expected.


kelson42:
I can really not judge the nature of the patch. It womders me that we need a "hack" but I will merge and if it works this is the most important.

PR #840, opened 2024-06-29
Fix file export extension and make it more generic

BPerlakiH:
Fixes: #814 

Renaming PDF specific parts to be more generic.
Figure out the file extension from the URL or from the mime-type of the content.


kelson42 (changes requested):
Please be coherent in the naming between "share" and " export"... I don't understand why we have a mix

Issue #839, opened 2024-06-29
Content default positioning is not always correct


kelson42:
See 
<img width="1080" alt="image" src="https://github.com/kiwix/kiwix-apple/assets/1029718/3d39ab6a-503b-4e15-bd11-19a1b9bfe1da">

PR #841, opened 2024-06-30
Remove default zoomScale, that causes wrong content adjustment

BPerlakiH:
Fixes: #839


kelson42:
@BPerlakiH Why was this zoom added in a first place? If it there I guess there is a reason so removing it will create a bug somewhere else!?

BPerlakiH:
> @BPerlakiH Why was this zoom added in a first place? If it there I guess there is a reason so removing it will create a bug somewhere else!?

According to the comment it meant to be fixing a state after screen resize, but I cannot reproduce anymore this specific problem. I tried with orientation changes, and also with resizing the view via iPad split view.
So it seems to me this specific problem it meant to fix no longer exists (maybe not a problem from iOS 16).

Issue #79, opened 2017-11-07
Implement „Find in page“


Popolechien:
"Search in Android not only finds pages, but matching terms within a page as well. Very helpful in longer articles."
(copied from beta tester feedback)

kelson42:
@automactic Please keep one ticket per issue. If you want to gather tickets together you have many systems, like tags, milestones, projects.

automactic:
I disagree. If you want to keep issue count down or wish to keep issues from becoming stale, the content of the tickets needs to be small and actionable items. IMO, we need to decide what can and should appear in the issue section of the project. 

If you want to keep feature request in the issue section, this is fine by me. But please also don't expect me to be able to quickly resolve issues that needs significant engineering efforts and planning. (i.e. expect issues to go stale)

kelson42:
@automactic I for sure don't want to keep **artificially** ticket count down and avoid stalled. Tickets should reflect the reality. This is mandatory to understand the status of the project and I do my best to secure a high quality ticket portfolio. Here are the guidelines to report bugs https://github.com/kiwix/overview/blob/master/REPORT_BUG.md, please let me know if something is unclear.

nk9:
It looks like this is the single issue being used for both iOS and Mac. Will the UI for this really be the same between the two? For example, one of the duped issues asked for Cmd+F to open this. That affordance makes sense on Mac but not iOS. I appreciate that the code base is the same for both but I'm wondering if it wouldn't be clearer to separate the issues by platform to account for differences like this. I suppose there could be a third bug for common code, but that feels like overkill.

Either way, we need to define how this UI will work on both platforms so that it can be implemented.

I am happy to lend a hand on this issue. I am new to the project, however. Is this a straightforward first issue for someone with a bit of Swift experience?

kelson42:
@nk9 Both versions, macOS and iOS use now the dame code base. Obviously, the keyboard shortcuts are not available on iOS. Regarding the implementation, such a feature has to use the HTML render builtin for the core of the feature. Regarding the UI, it's true that the situation is quite open but I woukd try to do like in Web browsers. How this has been solved in Safari/Chrome/Firefox?

nk9:
We're using a `WKWebView`, so it should be as easy as invoking the [web view's `find` method](https://developer.apple.com/documentation/webkit/wkwebview/3650493-find).

I've downloaded the source, will give it a shot and see if I can make a PR. If someone else wants to do it first, be my guest!

[...]

PR #849, opened 2024-07-07
79 implement find in page

BPerlakiH:
Fixes: #79 

## iPhone:

https://github.com/kiwix/kiwix-apple/assets/6784320/f83a6201-bc79-4ea7-aa1f-e54f0ed5c53b

## iPad:

https://github.com/kiwix/kiwix-apple/assets/6784320/0369c904-3253-4413-a954-617ad1532981

## macOS: 

https://github.com/kiwix/kiwix-apple/assets/6784320/f7d6ce3b-2025-4b03-ae99-55926af5d6a6


kelson42:
@BPerlakiH To me looks good, but actually I can not figure out where do you exactly click to raise this "find in page" feature?!

BPerlakiH:
@kelson42 I had to do a couple implementations, to adjust to each device:
- on iOS I had to add a new icon in the top bar (the one on the right):
Screenshot 2024-07-10 at 11 55 03

and on iPad (full screen mode) it's a new icon (left to the "home" icon):


whereas on macOS it's a floating button in the bottom right corner:
Screenshot 2024-07-10 at 11 57 16
or you can press "CMD+F" to bring it up.

kelson42:
@BPerlakiH Thank you for the clarification. Two remarks:
* For iPhone, I believe this feature is not proeminent enough to have its own logo. Any chance to make it available through two clicks somewhere? Idependently of this, I don't understand why we have so much space between icons?! Is that some kind of iOS/usability rule?
* For iPadOS, basically same remark as for iPhone, but far less problematic as we have a lot of space
* For macOS, the logo is not OK: far too small and not centered properly

BPerlakiH:
I think double tab/click might conflict with other actions reserved for that, like selecting / copying text and such within the web "area". If we do a long tap, or something more alternate it won't be found by users. By looking into our app reviews, most complaints we had regarding this behaviour for the catalog / settings icon (where you had to long press to bring it up).

The spacing is the default OS spacing for tab navigation items, so this is the accepted default by Apple for accessibility.

kelson42:
@BPerlakiH With two clicks, I don't mean a double click, I mean somewhere through a menu.

kelson42:
I will merge and UI improvement should be handled in separate issue

kelson42 (approved):
LGTM

VitoVan:
Waiting for the new release with excitement...

🎉 🎉 🎉

Issue #854, opened 2024-07-10
Crash 3.4.0 - Levenshtein.calculate(_:_:) +72


BPerlakiH:
<img width="685" alt="Screenshot 2024-07-10 at 11 31 37" src="https://github.com/kiwix/kiwix-apple/assets/6784320/1de9ee50-42ca-4f33-9637-9f26c57c4dcf">

seems like an infinite loop by the first sight.

PR #862, opened 2024-07-13
Fix search crash, and improve efficiency

BPerlakiH:
Fixes: #854 

Since the Levenshtein class has its own cache. The instance of the class and its cache was re-used for each search result, regardless of the searched term, this can cause undefined behaviour. 

I moved the instance creation within the loop, so it's starts with a fresh cache.
Measured the performance, and there's practically no difference before/after.

As for the performance measured earlier, that is for short strings, see my comment below for more complex cases.


kelson42:
@BPerlakiH don't merge main in feature branch

kelson42:
> this can cause undefined behaviour.

How that?

BPerlakiH:
@kelson42 I went deeper with this one:
The former algorithm wasn't really efficient.
With longer input values (in rage of ~1000 chars) it eating up to 4GB of memory, and run for 29 seconds...
Once it runs out of memory it will crash as described.
I tried to optimise the former algorithm, but got it down only to 400MB...

So I searched and adopted a new one.

The new algorithm produces the same outputs, but using only up to 9MB, and with the same test values it is only taking 0.5 sec to finish ;)

So it is not only fixing the crash, but also makes the searching quicker.

Issue #853, opened 2024-07-10
Crash 3.4.0 - libxml2.2.dylib: xmlMallocZero +48


BPerlakiH:
<img width="547" alt="Screenshot 2024-07-10 at 11 30 28" src="https://github.com/kiwix/kiwix-apple/assets/6784320/932480b8-f71e-4d35-bc02-0c221ed64d5c">

BPerlakiH:
@kelson42 I had a deeper look at this.
We have a dependency called `Fuzi`, which is a swift wrapper around `libxml2`.
It is linking the system provided version of `libxml2` [at build time](https://github.com/cezheng/Fuzi/blob/master/Package.swift#L14).

There's a couple of issues, it won't fail properly if initialised with invalid xml/html content data, there's already [a long standing issue on that on Fuzi's side](https://github.com/cezheng/Fuzi/pull/128).

I [implemented the proposed changes for that](https://github.com/cezheng/Fuzi/pull/128), and linked my version of Fuzi to the project, but it's still not 100%, as it is not failing properly under all conditions. Additional to that some issues have been already fixed in the underlying libxml2 as well. comparing my local version (2.09) and [the latest libxml2 version](https://gitlab.gnome.org/GNOME/libxml2/-/releases).

We have a couple of options here, bearing in mind that **only 2 devices crashed**:
- **A)** try to enforce the latest version of libxml2 by manually adding it and linking it, and use a fork of Fuzi for that
- **B)** switch Fuzi and libxml2 to something else (quality of 3rd party packages might vary, so there's no guarantee it will be an actual improvement)
- **C)** we use the html parsing in 2 places, the search and the bookmark. For the bookmarks there's already [a proposition to remove the html parsing](https://github.com/kiwix/kiwix-apple/pull/830). We can do the same for search. The difference is:

## Search results with default html parsing option (first sentence):
<img width="823" alt="Screenshot 2024-07-13 at 13 01 35" src="https://github.com/user-attachments/assets/1f713906-d1ef-4cb8-ace1-2dd8cb9a9174">

## Search results with html parsing disabled:
<img width="862" alt="Screenshot 2024-07-13 at 13 01 20" src="https://github.com/user-attachments/assets/137f9484-604f-4c42-a06a-b75b085a71ad">

We can be more subtle here, either: 
- **C1)** disable this functionality in the next release, so the html parsing won't be used, if we find a better replacement for it we can re-enable it. 
[...]

PR #865, opened 2024-07-13
Remove Fuzi, libxml2, Search snippets are using libzim only

BPerlakiH:
Fixes: #853 

Note: first Bookmark fixing PR should be merged: https://github.com/kiwix/kiwix-apple/pull/830 , as this one is based on that.


SwiftUI/Model/DefaultKeys.swift
```diff
@@ -24,7 +24,7 @@ extension Defaults.Keys {
         "externalLinkLoadingPolicy", default: Brand.defaultExternalLinkPolicy
     )
     static let searchResultSnippetMode = Key<SearchResultSnippetMode>(
-        "searchResultSnippetMode", default: Brand.defaultSearchSnippetMode
+        "searchResultSnippetModeV2", default: Brand.defaultSearchSnippetMode
```
BPerlakiH, line 27: Since we reduced the amount of options, we don't want former users to use their previous values in settings.
By changing the key, we successfully override former user defaults.

kelson42: I don't think this is the right approach. Instead, keep the same setting name, and create a compatibility and migration path for old values.

BPerlakiH: Thank you. Good point.
I have updated it, and checked how the values are mapped for migrating users. 
Due to the underlying library and the enum values, it works now as expected:

| Former user value | Migrated to |
| ---- | ---- |
| disabled | disabled |
| first paragraph | matches |
| first sentence | matches |
| matches | matches |

kelson42: Great please add a clear comment in the code explaining what is going on (with link to this issue) and then a todo, to remove all of this in a few year

BPerlakiH: I have added my explanation in code as well.

Issue #869, opened 2024-07-15
No book found line for "lowtech"


kelson42:
<img width="1080" alt="image" src="https://github.com/user-attachments/assets/698ec2ac-dba3-4c9a-9326-07a44fabd090">

But if I search on library.kiwix.org
<img width="1080" alt="image" src="https://github.com/user-attachments/assets/e240ef47-2822-4e43-b748-1c7fcf9c6da7">

PR #870, opened 2024-07-17
869 no book found line for lowtech

BPerlakiH:
Fixes: #869

It turned out that the low-tech ZIM is available in multiple languages.
We do support multiple user languages, according to which we filter the content.

Matching these 2 lists were not fully working. Multiple user selected content languages could match a ZIM if it was on the list of user languages. It wasn't working the other way around. Eg. user has "fra" only selected, and the ZIM has multiple languages according to the feed: "eng,fra,deu,nld,spa,ita,por,pol,ara,vie,kor", the former selection could not find it.

With the updated query now we can create multi to multi matching queries, and the missing ZIM items will show up.

The remanning changes were needed to make things testable.

### DIFF
--- a/Model/Entities/ZimFileMetaData/ZimFileMetaData.h
+++ b/Model/Entities/ZimFileMetaData/ZimFileMetaData.h
@@ -46,4 +46,25 @@
 // methods
 - (nullable instancetype)initWithBook:(nonnull void *)book;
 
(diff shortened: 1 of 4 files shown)


kelson42:
@BPerlakiH Like always, we need a proper PR description!

BPerlakiH:
> @BPerlakiH Like always, we need a proper PR description!

Sorry, please see it updated.

Issue #871, opened 2024-07-18
"Search Snippet" setting is outdated


kelson42:
<img width="1080" alt="image" src="https://github.com/user-attachments/assets/e156f808-29ad-426e-a470-21d0fbcdd9c0">

AFAIK it should be now a simple boolean switch

PR #873, opened 2024-07-20
Change search snippet in setting to be a toggle

BPerlakiH:
Fixes: #871 

Search snippet in settings is now a toggle instead of a drop down.


kelson42:
@BPerlakiH I wonder why there is no changes in the string files?! Please secure that all the code base is 100% clean of this old feature.

BPerlakiH:
From the string files we still use the title for the toggle, the remaining ones were removed here: https://github.com/kiwix/kiwix-apple/pull/865/files

Issue #874, opened 2024-07-20
OPDSParse with invalid data


BPerlakiH:
Make sure that inputing invalid data into the OPDSParse function is handled correctly.

We already have a unit test case for this, but so far it has been disabled, with the comment:
"Requires work in dependency to resolve the issue."

As [libkiwix is always returning true](https://github.com/kiwix/libkiwix/issues/1099), regardless of the input, this might cause some issues, as  invalid input handling is not in place.

On the reader side:

https://github.com/kiwix/kiwix-apple/blob/86a037c4b25c2954b7cabe4fd04dc1af7f5f361a/Model/OPDSParser/OPDSParser.swift#L27-L31

Whereas underneath that we have:
https://github.com/kiwix/kiwix-apple/blob/86a037c4b25c2954b7cabe4fd04dc1af7f5f361a/Model/OPDSParser/OPDSParser.mm#L42-L51

Related libkiwix ticket: https://github.com/kiwix/libkiwix/issues/1099

kelson42:
@BPerlakiH Please secure that whatever the libkiwix returns true or false, this is handled properly here and the app does not crash. For the rest, https://github.com/kiwix/libkiwix/issues/1099 should be fixed.

BPerlakiH:
Please, see the related PR here: https://github.com/kiwix/kiwix-apple/pull/884,
securing no crashes on the reader side.

PR #875, opened 2024-07-20
Enable OPDS invalid data input test-case

BPerlakiH:
This enables the test case for invalid OPDS data input, which initially will fail.
Once libkiwix will return false, for invalid data input, this testcase should pass.
See: #874 and https://github.com/kiwix/libkiwix/issues/1099


kelson42:
@BPerlakiH CI does not pass, please fix the PR

Issue #877, opened 2024-07-21
Crash 3.4.0: DataBase.mergeChanges()


BPerlakiH:
```
#1	(null) in -[NSPersistentHistoryChangeRequest initWithToken:delete:percentageOfDB:] ()
#2	(null) in +[NSPersistentHistoryChangeRequest fetchHistoryAfterToken:] ()
#3	0x000000010097c1e0 in closure #1 in Database.mergeChanges() at 
```
``SwiftUI/Model/Database.swift:116``

<img width="544" alt="Screenshot 2024-07-21 at 14 06 31" src="https://github.com/user-attachments/assets/488b83b8-3639-4921-823b-add1d16e913c">

PR #879, opened 2024-07-25
Fix Crash with Single DB background context

BPerlakiH:
Fixes: #877

We had multiple background contexts from which the DB queries were operating, without much of synchronisation, but at the end they were forced to merge there changes into a single place.
Additionally to that, we had a token mechanism to synchronise these contexts, which was crashing from time to time.

Solution:
- Use Apple's built in merge context solution, which is easier in our scenario (the token based approach is more appropriate for App Extensions such as Lock Screen widgets and so on, which we do not have).
- Make sure we use a single background context the queries of which are also queued up properly, so there's less of a chance for conflicts.

Tested on macOS, iPad and iPhone. Both the download and the category fetching / parsing, and other DB related actions, such as tab management.

### DIFF
--- a/App/App_iOS.swift
+++ b/App/App_iOS.swift
@@ -36,7 +36,7 @@ struct Kiwix: App {
         WindowGroup {
             RootView()
                 .ignoresSafeArea()
-                .environment(\.managedObjectContext, Database.viewContext)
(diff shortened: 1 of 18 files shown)


kelson42:
It's a really a huge PR and I have no clue if this is the right approach to solve the problem. 100% trusting you on that one.

Issue #881, opened 2024-07-26
Debug flag shows: Database concurrency issues


BPerlakiH:
With the debugger flag added to launch options:
-com.apple.CoreData.ConcurrencyDebug 1

The macOS app can crashed. 

Initially it seems, that: the list of downloaded files, referenced by the view are constantly updated (with each download progress change), and the indexing of the list is crashing in SwiftUI.

PR #882, opened 2024-07-26
Fix DB leading to UI crash on for each download

BPerlakiH:
Fixes: #881 
Changing the way we handle download progress.
The former implementation was writing the downloaded progress directly into the DB, as a background operation, additionally to that, it was using a timer with 0.25 sec updates to do so, which was causing the UI to update.
The UI is done in a for-each loop for each download task, but with the background operations going on so often, it was capable of crashing the app.
It was 100% re-producable, when enabling the:
"-com.apple.CoreData.ConcurrencyDebug 1"

So formerly we had:
download start -> start timer -> at each 0.25s write the progress into DB

Now we have:
-> download start -> on each progress -> inform the subscribed UI that there's a change

I also moved the restartToken out of the DB into UserDefaults (local storage), which allows to restart the download after the app closes, and re-opened by the user.

I did tested the downloading / pause / resume / restart app + restart download on all devices: iPhone, iPad, macOS.

It is no longer crashing with the above debug flag.

### DIFF
--- a/Model/DownloadService.swift
+++ b/Model/DownloadService.swift
@@ -64,7 +64,6 @@ final class DownloadTasksPublisher {
         }
     }
 
-
     func updateFor(uuid: UUID, downloaded: Int64, total: Int64) {
         if let state = states[uuid] {
             states[uuid] = state.updatedWith(downloaded: downloaded, total: total)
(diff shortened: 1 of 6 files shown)


kelson42:
@BPerlakiH Should we remove the draft?

PR #884, opened 2024-07-27
Make sure only utf8 compatible data can be passed into OPDS Parsing

BPerlakiH:
Fixes: #874 

It turned out that if the data passed in is UTF8 compatible we won't have a crash, no matter how invalid the content of the data is (eg. not an xml format).

Whereas it was possible to crash the app with non uff-8 compatible input, like so:
Screenshot 2024-07-27 at 12 53 46

Luckily we can handle this check on the swift / reader side.

Related unit tests were added, making sure we throw an error before the data with invalid encodings goes any deeper.

### DIFF
--- a/Model/OPDSParser/OPDSParser.mm
+++ b/Model/OPDSParser/OPDSParser.mm
@@ -42,6 +42,9 @@
 - (BOOL)parseData:(nonnull NSData *)data {
     try {
         NSString *content = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+        if (content == nil) {
+            return false;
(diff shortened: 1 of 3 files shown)


kelson42:
Where is the app crashing? Is the app crashing in libkiwix or in Swift?It seems you are just building an expensive workardound. This is NOK, we are building aolutions.

BPerlakiH:
@kelson42 Thank you, you are right. The crash occurs within libkiwix, as the content (encoded) will be nil. I updated it to catch this at the objective-c level, so there's no double encoding involved.

Issue #883, opened 2024-07-27
Increased text size setting in iOS cuts off items in the bottom menu


ronaldl29:
Here is a screenshot of the issue. The back button is cut off. 
![IMG_1508](https://github.com/user-attachments/assets/96a88728-5454-4ff4-8194-b4ea566f0375)

This happens when the text size setting is increased in iOS. 
![IMG_1509](https://github.com/user-attachments/assets/8071dd57-3971-4afd-8866-b39140b8489a)

kelson42:
@ronaldl29 Thank you for yor bug report. Can you please share with us both versions of iOS and Kiwix running on your device?

ronaldl29:
Thank you for the quick response!

Kiwix v3.4.0
iOS 17.5.1

@kelson42

kelson42:
@ronaldl29 Perfect, I have put this bug in the next milestone 3.5.0 and hopefuly this will be fixed within a few weeks.

ronaldl29:
You guys are awesome. Thank you!

BPerlakiH:
Please see the fix here with screenshots: https://github.com/kiwix/kiwix-apple/pull/886

PR #886, opened 2024-07-27
Fix Accessibility of icons when large text is set

BPerlakiH:
Fixes: #883 

Screenshot 2024-07-27 at 15 15 45
Screenshot 2024-07-27 at 15 13 29
Screenshot 2024-07-27 at 15 13 18


kelson42:
@BPerlakiH Thank you very much for your quick fix... but out of curiosity: why your search box is such inflated?

BPerlakiH:
I had the "Larger Accessibility Sizes" turned on in Settings.
Without that it looks better:
IMG_3027

PR #887, opened 2024-07-28
Add extra padding for zoomed in web pages

### LINKED ISSUE #885: Content margin are removed when no sidebar (zooming problem)
With sidebar
image

... and without sodebar
image

Like you can see the content as no left and right margin anymore and the margin should be there.

Not aure this is Kiwix related, might be related to the ZIM itself. It seems to happen only if I put a high zoom in the settinga

BPerlakiH:
Fixes: #885 

## Origin

I have found the zoom functionality is applying a JS callback to adjust the CSS values of the page:
https://github.com/kiwix/kiwix-apple/blob/c0539cc83bc7c2afb95ed95c6e5cc8bc94d01fc5/Views/BuildingBlocks/WebView.swift#L213-L220

## Proposed solution
I tried adjustments on the the swift view level, but that did not look good.

Therefore, I am proposing **to adjust the webkit padding** in a similar fashion as the text size increase is done for zooming in. 

As of my testing it looks good on all devices (iPhone, iPad, macOS).

### DIFF
--- a/Model/Utilities/Javascript.swift
+++ b/Model/Utilities/Javascript.swift
@@ -1,0 +1,29 @@
+// This file is part of Kiwix for iOS & macOS.
+//
+// Kiwix is free software; you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 3 of the License, or
+// any later version.
+//
+// Kiwix is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Kiwix; If not, see https://www.gnu.org/licenses/.
+
+import Foundation
+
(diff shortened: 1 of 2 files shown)


kelson42:
I don't think this is a good idea at all to modify the css or dom. Why do we do that?

BPerlakiH:
This was the implementation of the "page zoom" functionality.

kelson42:
@BPerlakiH There should be some kind of zoom provided by the html widget itself

BPerlakiH:
We have the default pinch gesture to zoom:

https://github.com/user-attachments/assets/11211905-f0d1-485b-b522-5a9ef091a2aa

kelson42:
pinch is not what we need or talk about here!

BPerlakiH:
We have the "magnify" option for the webView, but that zooms in a similar fashion as the pinch zoom, making the letters relatively larger, but you need to scroll horizontally as well to read it. 
With the current solution the margins remain nicely in place, and the text wraps at the side of the screen.

BPerlakiH:
Converted to draft, as it's a stale PR / issue.

BPerlakiH:
Since the GH issue is closed, we no longer need this.

Issue #889, opened 2024-07-29
Clicking on a fullscreen video results in forever black screen


benoit74:
Scraper: youtube (but probably other are impacted the same)
App version: 3.4.1 (170) via TestFlight
OS: Mac OS Sonoma 14.5

ZIM tested:
- https://download.kiwix.org/zim/videos/avanti-3dimensional-geometry_2024-03.zim (old one, where sound does not work)
- https://download.kiwix.org/zim/videos/avanti-3dimensional-geometry_2024-06.zim (new one, with VP9 videos, sound working, ...)

Steps to reproduce the bug:
- start a video
- go into fullscreen
- click with the mouse in the fullscreen (not on the control bars, but anywhere in the video) => the video stops
- click again in the full screen => the sound resume but the video stays black in fullscreen

Note that from the moment the bug appears, the fullscreen is always black but if we exit fullscreen video plays well again

Note also that close the Kiwix window (not the app, only the window) is sufficient to restore fullscreen working properly.

BPerlakiH:
Thank you @benoit74 for reporting this.

I am able to fully reproduce it with the provided videos.
I have retested this case with the test videos provided by @rgaudin:
<img width="275" alt="Screenshot 2024-07-30 at 11 30 29" src="https://github.com/user-attachments/assets/fb9a9053-1e7c-4bec-a4a8-6efb8434a21b">

This is, what I have found so far:

The reported ZIM is using:
- webm + Video.js 8.6.1 + videojs-ogvjs @version 2.0.0

This issue won't occur in the test videos, which are using:
- webm + no video js (pure html video tag)
- mp4 + video js "8.12.0"; + ogv V = "1.9.0-20240423190513-22f6924"


Do you think we can fix it on the content side?

kelson42:
> Do you think we can fix it on the content side?

No we can't. I don't understand how we have missed this?! It has clearly been stated that this was working at 3.4.0 release, so what happened?

BPerlakiH:
@kelson42 I think we tested some of the combinations of video js and no js versions, and I suppose based on that we assumed it will always work. The reported ZIM file is using video js release from "Oct 12, 2023". I don't remember any discussion around specific versions of videojs (or ogv.js) that we meant to support.

The good news is that it works without those JS files, and since the content cannot be changed on the input size, that lead me to the following:

[...]

PR #892, opened 2024-07-30
Blocking any JS files, that contains video or ogv in the name, to fix video issues

BPerlakiH:
Fixes: #889 

It's a bit hack... since the content cannot be changed, and it works better without those JS files, it is possible solution to our problem.


rgaudin:
What's the explanation?

kelson42:
This "solution" is not acceptable obviously.

PR #903, opened 2024-08-03
Fix incorrect language selector list in library settings

HiroyasuNishiyama:
The language list in the library settings panel seems to have the following problems, as shown in the following figure:

- Some language names are displayed in internal strings (e.g. `eng,por,spa,...`) ,
- ZIM files count is not correct.

スクリーンショット 2024-08-03 21 24 43

This seems to be due to the fact that if multiple languages are set in the language attribute of the ZIM file, they are not handled correctly.

This PR attempts to correct that.

### DIFF
--- a/Views/Settings/LanguageSelector.swift
+++ b/Views/Settings/LanguageSelector.swift
@@ -164,22 +164,36 @@ class Languages {
         fetchRequest.propertiesToGroupBy = ["languageCode"]
         fetchRequest.resultType = .dictionaryResultType
 
-        let languages: [Language] = await withCheckedContinuation { continuation in
+        let languages: [(String, Int)] = await withCheckedContinuation { continuation in
             Database.shared.performBackgroundTask { context in
                 guard let results = try? context.fetch(fetchRequest) else {
                     continuation.resume(returning: [])
                     return
                 }
-                let language: [Language] = results.compactMap { result in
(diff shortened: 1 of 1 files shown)


kelson42:
@HiroyasuNishiyama Thank you for this new PR around language mgmt. Not being able to read swift this is how it should work if we have a multilang ZIM: counter should be incremented of one for each language listed (and of course no multilang item in the list, so only one language item).

BPerlakiH:
Thank you @HiroyasuNishiyama for bringing this up. Can you please do a step by step how to reproduce the state on your screenshot?

I did clean and build from the main branch, I also launched the Kiwix macOS app from the latest TestFlight build and cannot reproduce this issue.
In both cases I see this:
Screenshot 2024-08-04 at 13 11 51

BPerlakiH:
Thank you @HiroyasuNishiyama to pointing this out. I did manage to recreate this issue.
I opted for a slightly different implementation, and moved out the "collecting the language codes" into a new class, which can be tested for various edge cases, please have a look here:
https://github.com/kiwix/kiwix-apple/pull/906

HiroyasuNishiyama:
@BPerlakiH  Thank you.
I think your fix looks better.

kelson42:
Superseeded by #906. Closing.

@HiroyasuNishiyama Thanks again for your effort!

PR #908, opened 2024-08-04
Change to release version 3.5.0

BPerlakiH:
Before we do the release we need to align the version number to our updated milestone.
We can merge this after the weekly builds run, as the App Store version needs to be changed as well.


BPerlakiH:
@kelson42 I have updated the AppStore versions to be 3.5.0.
It can be merged now.

Issue #900, opened 2024-08-02
Download error feedback


rgaudin:
I have two distinct bugs in this short screencast:

- My library is outdated (2 months old), despite having checked to update automatically.
- When trying to download a ZIM that's missing (proxyman tells me it results in a `404` for https://download.kiwix.org/zim/wikipedia/wikipedia_en_golf_maxi_2024-04.zim), I have zero feedback that there has been an error.

https://github.com/user-attachments/assets/d7a11a44-0689-4862-9a39-5bea39ef2839

BPerlakiH:
Ok, I think we need to split this issue into 2 parts:
1) we need to find out why you have 2 months old library, when you have auto-refresh on. It might be related some older debug version of the app, that you might "stuck" with, without clearing the app install folder. It might be only local to you, but might be a "real" user update issue.
2) For now let's focus on the download issue, since you are stuck in this phase, that's the ideal "test environment" for us. I will have a look at how the download errors are handled, and if we could possibly apply a user facing alert at least.

BPerlakiH:
The error handling part is fixed here: https://github.com/kiwix/kiwix-apple/pull/912

BPerlakiH:
I moved the outdate library issue here: https://github.com/kiwix/kiwix-apple/issues/922
and renamed this one to be related to the download error only.

PR #912, opened 2024-08-05
Handle download status code errors

BPerlakiH:
Fixes: #900 

It is fixing actually only part of the ticket, which is how we handle the download errors:

https://github.com/user-attachments/assets/89dad530-7053-478e-85d1-889f7183f848


rgaudin (approved):
LGTM ; can we open a different ticket for the other part so the linked ticket can be closed

Issue #920, opened 2024-08-06
On fresh install opening a ZIM file directly after download crashes the app on iPad


BPerlakiH:
When downloading a ZIM file and trying to open the main page immediately, can crash on the iPad.

After some investigation, it seem that my former commit: #7e706f32ad19fb353fb28209814c1bd07db7be94

is responsible for this.

BPerlakiH:
I have found the real issue, please see the comments in my PR: https://github.com/kiwix/kiwix-apple/pull/921

PR #921, opened 2024-08-07
Fix crash: opening ZIM directly after download crashes the app on iPad

BPerlakiH:
Fix: #920

After my investigation it turned out that the real problem lied somewhere deeper.
My former fix in 7e706f32ad19fb353fb28209814c1bd07db7be94 did solved only part of the issue, and postponed the problem to the point where the download is finished.

## Background:
We select a ZIMFile, for that we create a DownloadTask which keeps track of our progress, and can resume after app restart. So far so good.

The DownloadDetails view was relying on the DownloadTask itself, since that had a back reference to the ZIMFile, it could display all necessary details.

The problem is, that as it turns out SwiftUI has a bug, that if you delete the underlying managed object in our case the DownloadTask, which is deleted after the download completes, well... SwiftUI will just hold on to the old reference of that object, and will crash.
See the related problems reported by others here: 
https://stackoverflow.com/questions/59123439/swiftui-holding-reference-to-deleted-core-data-object-causing-crash

## The solution
Since we start the whole process from the ZIMFile, which also has an optional reference to the DownloadTask, we can base our views on the ZIMFile instead, which might or might not have a DownloadTask. Since we operate on the optional value of DownloadTask, it won't crash.

### TLDR:
We replace all of the ``downloadTask`` strong references in those views with ``downloadZimFile.downloadTask?`` and it stops crashing.

### DIFF
--- a/Model/DownloadService.swift
+++ b/Model/DownloadService.swift
@@ -221,8 +221,8 @@ final class DownloadService: NSObject, URLSessionDelegate, URLSessionTaskDelegat
         Task { @MainActor [weak progress] in
             progress?.resetFor(uuid: zimFileID)
         }
-        let context = Database.shared.viewContext
-        context.perform {
+        Database.shared.performBackgroundTask { context in
+            context.mergePolicy = NSMergePolicy.mergeByPropertyObjectTrump
             do {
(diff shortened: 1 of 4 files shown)


rgaudin (approved):
LGTM

Issue #922, opened 2024-08-07
My library is outdated (2 months old), despite having checked to update automatically.


BPerlakiH:
My library is outdated (2 months old), despite having checked to update automatically.

https://github.com/user-attachments/assets/d7a11a44-0689-4862-9a39-5bea39ef2839

BPerlakiH:
@rgaudin could you please send me your macOS folder of Kiwix? 
Maybe that could help to reproduce your local current state, and see where the problem is.

rgaudin:
https://tmp.kiwix.org/ci/self.Kiwix_rgaudin-922.tgz

I see no outgoing network request so I think it's not a network issue but a condition one.

BPerlakiH:
@rgaudin @kelson42 
Ok, I looked into this one: 
**on iOS** we schedule a background task when the AutoRefresh is set, which will kick in 24 hours after the last refresh date.

**On macOS**, the only point it triggers auto-refresh is when you visit the "New" section...
So the question is: how AutoRefresh should really work ? Do we want the same on macOS as on iOS ?

BPerlakiH:
On the same note, I think it is very confusing for a fresh user, that a freshly installed application is pretty much "useless", until the "Fetch Catalog" button is not pressed.
So ideally we should probably start the app with an auto-refresh, since it is enabled by default anyways.

rgaudin:
I agree with both: we should auto-fetch on start and fetch periodically on macOS. That's what I'd expect as a user from that option.

BPerlakiH:
So far I have found: 
- In short there's no "background task" equivalent for macOS. 
We could do an auto fetch on start, currently the result of the fetch takes the user to the catalog, which looks strange on start.
I also discovered an issue with this catalog navigation on iOS, I will open a new issue for that.
Since the user can visit one of the categories, while we are fetching, the current message "No ZIM file in this category" could also be updated, if we are still "fetching".

BPerlakiH:
Related iPhone issue is here: https://github.com/kiwix/kiwix-apple/issues/923

kelson42:
> I agree with both: we should auto-fetch on start and fetch periodically on macOS. That's what I'd expect as a user from that option.

@BPerlakiH Whatever how this is done, this is what we need to do... and it has to provide the same feature in iOS and macOS. Like always, try to keep as far as possible the most same code base between both OSes.

BPerlakiH:
ok, to sum up, the plan is:
[...]

PR #926, opened 2024-08-11
922 my library is outdated 2 months old despite having checked to update automatically

BPerlakiH:
Fixes: #922 

Summary of changes:
- remove background processing / background fetching for iOS
- since we no longer scheduling background task, those permissions were also removed

- auto update works in the following way: 
- If enabled we will check if the library should be re-fetched (meaning it's older than 1 day), in the following cases:
- the user foregrounds the app (either when starting the app, or background + foreground), on macOS it's the moment of the app starting
- the user visits either the "New" or "Categories" sections
- additionally we update the displayed messages while fetching, so instead of "No new ZIM Files" we indicate that we are still "Fetching..."

I've added some additional safeguard for the app start / foregrounding, as on Custom apps we do not need to fetch categories if they are not included, the sections "New" and "Categories" are turned off on those apps, so that's OK.

### DIFF
--- a/App/App_iOS.swift
+++ b/App/App_iOS.swift
@@ -28,7 +28,6 @@ struct Kiwix: App {
 
     init() {
         fileMonitor = DirectoryMonitor(url: URL.documentDirectory) { LibraryOperations.scanDirectory($0) }
-        LibraryOperations.registerBackgroundTask()
         UNUserNotificationCenter.current().delegate = appDelegate
     }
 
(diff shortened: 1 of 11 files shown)


rgaudin (approved):
LGTM appart from the failing tests

BPerlakiH:
I've updated the unit tests. It is somewhat complex case, but here's a short rundown:
- for historic reasons (probably testing some DB things) we run the full application as the host for the unit tests (although for most unit-tests it is not necessary).
- Since we have a singleton instance behind the LibraryProcess, reflecting the state of the fetching, that fetching is also started when the app run by unit-tests is launched.
- Current shortest solution is to have a way to inject a non-shared instance for unit tests.

Issue #933, opened 2024-08-16
iPhone video starts in full screen


BPerlakiH:
## Environment:
iPhone iOS 17.5 | current develop branch
**It is only affecting iPhone iOS 17**

## Content:
[TED collective (31 Jul 2024)](https://download.kiwix.org/zim/ted/ted_mul_collective_2024-07.zim.meta4)
[TED humanities (02 Aug 2024)](https://download.kiwix.org/zim/ted/ted_mul_humanities_2024-08.zim.meta4)

## Reproduce steps:
Starts playing the video content.
**Expected behaviour**: the video content should start playing within the "small" player.
**Actual behaviour**: the video plays in full screen mode.

https://github.com/user-attachments/assets/1fa6cecb-879b-4f3f-ad15-42050ca03524

PR #924, opened 2024-08-08
Fix iPhone video starting with a black screen on iOS 17 iPhone

BPerlakiH:
Fixes: #916 
It seems it is an iOS 17 iPhone bug:
if we have a poster attribute set, the video starts with a black screen.

## Content used:
Currently in the "New" section:
- [TED collective (31 Jul 2024)](https://download.kiwix.org/zim/ted/ted_mul_collective_2024-07.zim.meta4)
- [TED humanities (02 Aug 2024)](https://download.kiwix.org/zim/ted/ted_mul_humanities_2024-08.zim.meta4)

## Solution:
TLDR: Remove poster attribute on iPhone iOS 17.

## Remove the poster attribute and having a poster displayed:
Looking deeper inside the HTML content we are using, we have the following structures:

## TED "I hope" | iPhone 17.5 | BEFORE poster removal
```html
<div id="video-wrapper">
    <div id="vjs_video_3"
        data-setup="{&quot;techOrder&quot;: [&quot;html5&quot;, &quot;ogvjs&quot;], &quot;ogvjs&quot;: {&quot;base&quot;: &quot;assets/ogvjs&quot;}, &quot;autoplay&quot;: false, &quot;preload&quot;: true, &quot;controls&quot;: true, &quot;controlBar&quot;: {&quot;pictureInPictureToggle&quot;:false}}"
        poster="videos/87396/thumbnail.webp" playsinline="true" preload="auto"
        class="video-js vjs-default-skin vjs-fill vjs-paused vjs_video_3-dimensions vjs-controls-enabled vjs-touch-enabled vjs-v7 vjs-user-active"
        tabindex="-1" lang="en-gb" role="region" aria-label="Video Player"><video class="vjs-tech" preload="true"
            playsinline="playsinline" poster="videos/87396/thumbnail.webp"
            data-setup="{&quot;techOrder&quot;: [&quot;html5&quot;, &quot;ogvjs&quot;], &quot;ogvjs&quot;: {&quot;base&quot;: &quot;assets/ogvjs&quot;}, &quot;autoplay&quot;: false, &quot;preload&quot;: true, &quot;controls&quot;: true, &quot;controlBar&quot;: {&quot;pictureInPictureToggle&quot;:false}}"
            id="vjs_video_3_html5_api" tabindex="-1" src="videos/87396/video.webm">
            <source src="videos/87396/video.webm" type="video/webm">
        </video>
        <div class="vjs-poster" aria-disabled="false"
[...]

PR #936, opened 2024-08-18
933 iphone video starts in full screen

BPerlakiH:
Fixes: #933 

Note it is based on the PR: https://github.com/kiwix/kiwix-apple/pull/924

## Solution:

As of [Apple's doc](https://developer.apple.com/documentation/webkit/wkwebviewconfiguration/1614793-allowsinlinemediaplayback), we need to set:
``allowsinlinemediaplayback``
and additionally set:
``mediaTypesRequiringUserActionForPlayback`` - otherwise it's not working
and we need to enforce that there's a ``playsinline`` attribute on the html `<video>`.

### DIFF
--- a/Support/injection.js
+++ b/Support/injection.js
@@ -95,6 +95,7 @@ function fixVideoElements() {
             if(attributes.getNamedItem('poster')) {
                 attributes.removeNamedItem('poster');
             }
+            video.setAttribute('playsinline', '');
         });
     }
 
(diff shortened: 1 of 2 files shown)


kelson42 (commented):
@BPerlakiH With the two files reported in the issue, `playsinline` is available so this is not the root cause of this issue. As such and because we should not modify the DOM, I believe this part of the patch is not wanted. @rgaudin What do you think?

rgaudin:
Noticed this as well but since it's a requirement for it to work, I think it's good to keep it.

PR #943, opened 2024-08-19
Fix Blank item in search results

BPerlakiH:
Fixes: #940 

The issue was mainly that if we had the same url for the search result, but with different snippets, that created the same id for both, and one of them could not be rendered, because of this conflict.
It's safer to use the hash of full object, rather than the url's hash only, plus it's better to reduce the search results by the same id, before we process them any further.

I did simplify some of the loops around the search results processing, which made the search teeny-tiny faster.

Since we no longer do html parsing, the dispatchGroup solution is no longer needed, it just creates a lot of context switches between the threads, and it's actually measurably slower than without it.

### DIFF
--- a/SwiftUI/Model/SearchOperation/SearchOperation.swift
+++ b/SwiftUI/Model/SearchOperation/SearchOperation.swift
@@ -16,21 +16,23 @@
 import Defaults
 
 extension SearchOperation {
-    var results: [SearchResult] { get { __results.array as? [SearchResult] ?? [] } }
+    var results: [SearchResult] { __results.array as? [SearchResult] ?? [] }
 
     open override func main() {
         // perform index and title search
         guard !searchText.isEmpty else { return }
         performSearch()
 
-        // parse and extract search result snippet
         guard !isCancelled else { return }
-        let snippetMode = Defaults[.searchResultSnippetMode]
-        for result in results {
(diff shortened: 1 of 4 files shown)


SwiftUI/Model/SearchOperation/SearchOperation.swift
```diff
@@ -16,47 +16,38 @@
...
-                        data: data,
-                        options: [.documentType: NSAttributedString.DocumentType.html,
-                                  .characterEncoding: String.Encoding.utf8.rawValue],
-                        documentAttributes: nil
-                    )
-                case .disabled:
-                    break
-                }
+        // parse and extract search result snippet
+        if case .matches = Defaults[.searchResultSnippetMode] {
```
BPerlakiH, line 32: I moved this `if` outside of the loop. It's more efficient this way.

SwiftUI/Model/SearchOperation/SearchOperation.swift
```diff
@@ -16,47 +16,38 @@
 import Defaults
 
 extension SearchOperation {
-    var results: [SearchResult] { get { __results.array as? [SearchResult] ?? [] } }
```
BPerlakiH, line 19: This is only syntax change, we do not need to add `get`, it is going to be the same without it.

Issue #946, opened 2024-08-20
Search-bar is inconsistent on iPhone


BPerlakiH:
Used iPhone 15 Pro Max simulator,
Content: https://download.kiwix.org/zim/wikipedia/wikipedia_en_geography_nopic_2024-06.zim.meta4

Comparing when the user comes from the bookmark link to a page, vs. coming from a search result:
![Simulator Screenshot - iPhone 15 Pro Max - 2024-08-20 at 13 50 43](https://github.com/user-attachments/assets/a4216245-654d-4507-a439-827ce9ebf54f)

![Simulator Screenshot - iPhone 15 Pro Max - 2024-08-20 at 13 51 43](https://github.com/user-attachments/assets/a3bfa61e-fa02-4ed3-bcb3-dbe2de901864)

PR #947, opened 2024-08-20
Fix search bar inconsistencies on iPhone

BPerlakiH:
Fixes: #946 

## Root cause:
We store the button tab items when the search starts, and restore them when the search is cancelled. While the problem is that we should also restore them if the search succeeds, and we navigate away.

### DIFF
--- a/ViewModel/SearchViewModel.swift
+++ b/ViewModel/SearchViewModel.swift
@@ -20,7 +20,6 @@ import Defaults
 
 class SearchViewModel: NSObject, ObservableObject, NSFetchedResultsControllerDelegate {
     @Published var searchText: String = ""  // text in the search field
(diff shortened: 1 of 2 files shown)


ViewModel/SearchViewModel.swift
```diff
@@ -18,9 +18,8 @@ import CoreData
 
 import Defaults
 
-class SearchViewModel: NSObject, ObservableObject, NSFetchedResultsControllerDelegate {
+final class SearchViewModel: NSObject, ObservableObject, NSFetchedResultsControllerDelegate {
     @Published var searchText: String = ""  // text in the search field
-    @Published var isSearching = false  // for iOS 15 & 14
```
BPerlakiH, line 23: no longer in use

Issue #954, opened 2024-08-25
Crash on clearing recent search 


HiroyasuNishiyama:
Clicking clear recent search button crashes kiwix application on iPadOS 17.5.
Running on Xcode, it shows the following uncaught exception:

```
*** Terminating app due to uncaught exception 'NSGenericException', reason: 'Your application has presented a UIAlertController (<SwiftUI.PlatformAlertController: 0x110263200>) of style UIAlertControllerStyleActionSheet from _TtGC7SwiftUI19UIHostingControllerGVS_15ModifiedContentVS_7AnyViewVS_12RootModifier__ (<_TtGC7SwiftUI19UIHostingControllerGVS_15ModifiedContentVS_7AnyViewVS_12RootModifier__: 0x112027800>). The modalPresentationStyle of a UIAlertController with this style is UIModalPresentationPopover. You must provide location information for this popover through the alert controller's popoverPresentationController. You must provide either a sourceView and sourceRect or a barButtonItem.  If this information is not known when you present the alert controller, you may provide it in the UIPopoverPresentationControllerDelegate method -prepareForPopoverPresentation.'
```

It works well on iPadOS 17.0, iOS, and macOS.

BPerlakiH:
Thank you @HiroyasuNishiyama for reporting this.

PR #955, opened 2024-08-25
Remove search clear dialog, as it's crashing on iPadOS 17.5

BPerlakiH:
Fixes: #954 

It seems to be an issue with SwiftUI itself, therefore the quickest fix we can do is not to have this confirmation dialog, but clearing the search history directly.


kelson42:
@HiroyasuNishiyama  Does that fix the issue properly to your experience/opinion?

HiroyasuNishiyama:
@kelson42 Using the updated source code, the crash no longer occurs on iPad.
Personally, I have no issue with changing the specification to clear recent search without confirmation.
However, other users might be concerned about the specification change, so it might be better to make this change work only for iPadOS 17.5.

kelson42:
Lets go with this patch and see if someone complains.

Issue #798, opened 2024-06-08
Change url scheme from "kiwix://" to "zim://"


BPerlakiH:
We want to update all in app urls to have a scheme "zim://" instead of "kiwix://".
Note: we might need to migrate the user bookmarks, to make things back-ward compatible as well.

kelson42:
@BPerlakiH Here is the small [documentation about the URLs](https://github.com/kiwix/libkiwix/wiki/ZIM-URLs), obviously use relative URL when needed.

kelson42:
@BPerlakiH Please first make a PR to handle in a generic fashion this kind of data migration challenge.

PR #956, opened 2024-08-25
Migration services added

BPerlakiH:
Initial generic migration service with tests.
Related to: #798 

We will yet to see in practice if `async` is needed for this, but for start let's keep it simple.

### DIFF
--- a/Model/Migration/MigrationService.swift
+++ b/Model/Migration/MigrationService.swift
@@ -41,10 +41,8 @@ struct MigrationService {
 
     func migrateAll(using userDefaults: UserDefaulting = UserDefaults.standard) -> Bool {
         var allSucceeded = true
-        for migration in migrations {
-            if migration.migrate(userDefaults) == false {
-                allSucceeded = false
-            }
+        for migration in migrations where migration.migrate(userDefaults) == false {
+            allSucceeded = false
         }
         return allSucceeded
     }
--- a/Model/Migration/MigrationService.swift
+++ b/Model/Migration/MigrationService.swift
@@ -1,0 +1,51 @@
+// This file is part of Kiwix for iOS & macOS.
+//
+// Kiwix is free software; you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 3 of the License, or
+// any later version.
+//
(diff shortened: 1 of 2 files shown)


kelson42:
I don't see any modularity in this. So here are the things:
* You should be able to have X migration modules
* Each migration module should have a part to detect if the migration should apply, and the function to apply
* Each migration module is typically a class
* You need a solution to secure that module are applied in a certain order
* Adding a migration scenario should not be more complex than creating the class with the proper interface

BPerlakiH:
> I don't see any modularity in this. So here are the things:
> 
> * You should be able to have X migration modules
> * Each migration module should have a part to detect if the migration should apply, and the function to apply
> * Each migration module is typically a class
> * You need a solution to secure that module are applied in a certain order
> * Adding a migration scenario should not be more complex than creating the class with the proper interface

I see, maybe from this PR it is not looking like what it really is.

> * You should be able to have X migration modules
Yes, it is possible to add any number of migrations, the following way:
```swift
            let migrations = MigrationService(migrations: [
                Migrations.schemeToZIM(using: Database.shared.viewContext)
            ])
            _ = migrations.migrateAll()
```
Let's say we need more migrations:
```swift
            let migrations = MigrationService(migrations: [
                Migrations.schemeToZIM(using: Database.shared.viewContext),
                Migration(userDefaultsKey: "migrate_user_settings", migration: UserSettings.migration),
            ])
            _ = migrations.migrateAll()
```
and so on.

> * Each migration module is typically a class
In this case each migration is an instance of ``struct Migration``.
To construct such, we need 2 things:
- a user default key: this is used to store if the migration was completed or not (so we only run it once)
- a function: this is doing the actual work of the migration
We then pass in a list of these ``Migration`` structs to the ``MigrationManager`` and that will execute those that are not yet completed, in that specific order.

> * You need a solution to secure that module are applied in a certain order
The order of execution is the order of that ``Migration`` struct list passed into the ``MigrationManager``.

> * Adding a migration scenario should not be more complex than creating the class with the proper interface
I agree, it's even simpler, since you only need to create an instance of the ``struct Migration`` and pass it into the ``MigrationManager``.
What we can do better, is probably merge together the point that is used for both iOS and macOS, so it's not duplicated.

Issue #959, opened 2024-09-01
Show user error when libkiwix.org is not available


BPerlakiH:
Currently there's no user facing error, that would indicate that the library service is not reachable, especially for a newly installed app:

https://github.com/user-attachments/assets/0c0bcb01-d2ae-461d-9053-02be77c799ac

PR #960, opened 2024-09-01
Error handling for catalog not available 503

BPerlakiH:
Fixes: #959

### Solution: 
Additional enum case for error, and re-use existing translation key for displaying the error.

### DIFF
--- a/Tests/LibraryRefreshViewModelTest.swift
+++ b/Tests/LibraryRefreshViewModelTest.swift
@@ -45,7 +45,7 @@ final class LibraryRefreshViewModelTest: XCTestCase {
 
         HTTPTestingURLProtocol.handler = { urlProtocol in
             let response = HTTPURLResponse(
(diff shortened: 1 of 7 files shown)


kelson42 (changes requested):
@BPerlakiH IMHO you should have a library URL.builder which should be totally disconnected from the libraryView... and the reuse it in your observer

BPerlakiH:
> @BPerlakiH IMHO you should have a library URL.builder which should be totally disconnected from the libraryView... and then reuse it in your observer. Havin to hardcode URL wherever you need it is not a good sign.

I think it's not that bad as it seems. Formerly we had the url hard coded in a single place in the LibraryViewModel (which is not a view). The only change is that I moved it up higher to the top of the file, so it's easier to find. That's about it, when it comes to the production code.

In tests, now I updated to use a mock url, and it works equally good with that as well. The responses are also mocks, which is they key part of those tests.

PR #961, opened 2024-09-01
798 change url scheme from kiwix to zim

BPerlakiH:
Fixes: #798 

**Note**: changes applied on top of the #956 PR.

### Update the code to create "zim://" urls
Whenever we re-constructing urls (eg. to create bookmarks), the code has been updated.

### Migrating bookmarks:
This was the easy part, once the app starts we do a one-off migration, selecting the bookmarks that start with the url "kiwix://" and update those to be "zim://"

### Migrating tabs:
This was the hard part. We store the tabs together with the so called "interactiveState" of the webView loaded. Which is a serialised data provided by Apple. This contains the loaded url, and the browsing history (eg. former urls).

Formerly we used a "trick" to update such urls, for the [migration of ZIM files](https://github.com/kiwix/kiwix-apple/pull/700/files), which is in short utf8 decode the data, and from that regex replace the urls.

Now the same trick is not working to change the url scheme, possibly because the length of the string is changing. Formerly we updated the zim id of the urls, which is the same length. (The scheme change from "kiwix" to "zim" is not the same length).
**Editing the data as utf8 decoded string (as before), breaks the tabs, and they won’t load.**

This "interactiveState" of the webView is not properly modifiable, or decodable in a reliable fashion, [details of serialisation here](https://github.com/WebKit/WebKit/blob/debb3b092dee3eb99fc49e79423f7a627b72dbff/Source/WebKit/Shared/SessionState.serialization.in).

It seems to be plist file, but cannot be decoded neither with python (``plistlib``), nor with swift (``PropertyListDecoder``), nor with macOS (``plutil``), without the detailed serialisation knowledge, it just won’t work.

**TLDR: For these reasons, it should not be modified directly.**

### Solution for tab urls:
Since the tab urls, and interactive state is "saved" once we leave a tab (or the application), the solution is to migrate those urls "on demand", meaning once we want to load such a URL, we check if it's an old "kiwix" one, and modify it to be a new "zim" url, and load that instead.
Once loaded as a new "zim" URL, it will be saved correctly.

### Possible followup:
Since we have a new DWDS content, in a separate PR, I can remove this "utf-8" decoding "trick" introduced by [this older PR](https://github.com/kiwix/kiwix-apple/pull/700/files), and replace it with a similar "on demand" migration solution to change the hosts (ZIM file id migration from old to new).


Tests/BookmarkMigrationTests.swift
```diff
@@ -42,4 +42,9 @@ final class BookmarkMigrationTests: XCTestCase {
...0F}\u{10}\u{11}\u{12}\u{13}\u{14}\u{15}_\u{10}\u{17}SessionHistoryEntryData_\u{10}\u{18}SessionHistoryEntryTitle_\u{10}2SessionHistoryEntryShouldOpenExternalURLsPolicyKey_\u{10}\u{16}SessionHistoryEntryURL_\u{10}\u{1E}SessionHistoryEntryOriginalURLO\u{10}P\0\0\0\0\0\0\0\0\u{02}\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0�\u{07}�@�\u{13}\u{06}\0\0\0\0\0\0\0\0\0����\0\0\0\0�\u{07}�@�\u{13}\u{06}\0����\0\0\0\0�\u{03}\0\0\0\0�?\0\0\0\0����TDWDS\u{10}\u{01}_\u{10}2kiwix://A992BF76-CA94-6B60-A762-9B5BC89B5BBF/index_\u{10}2kiwix://A992BF76-CA94-6B60-A762-9B5BC89B5BBF/index�\u{0C}\r\u{0E}\u{0F}\u{10}\u{17}\u{18}\u{19}\u{1A}\u{1B}O\u{10}P\0\0\0\0\0\0\0\0\u{02}\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0V<�A�\u{13}\u{06}\0\0\0\0\0\0\0\0\0����\0\0\0\0U<�A�\u{13}\u{06}\0����\0\0\0\0\0\0\0\0\0\0�?\0\0\0\0����[hier und da\u{10}\u{02}_\u{10}<kiwix://A992BF76-CA94-6B60-A762-9B5BC89B5BBF/wb/hie_.und_.da_\u{10}<kiwix://A992BF76-CA94-6B60-A762-9B5BC89B5BBF/wb/hie_.und_.da\u{10}\u{01}\0\u{08}\0\u{0F}\0\u{1E}\0-\0<\0>\0?\0F\0^\0{\0�\0�\0�\0�\0�\u{01}\u{0B}\u{01}$\u{01}E\u{01}�\u{01}�\u{01}�\u{01}�\u{02}\t\u{02}\u{14}\u{02}g\u{02}s\u{02}u\u{02}�\u{02}�\0\0\0\0\0\0\u{02}\u{01}\0\0\0\0\0\0\0\u{1D}\0\0\0\0\0\0\0\0\0\0\0\0\0\0\u{02}�"
```
BPerlakiH, line 42: This is how the "interactiveState" looks like in pure data.

rgaudin:
I see no change to a plist file. Is the new scheme properly announced to the OS?

BPerlakiH:
> I see no change to a plist file. Is the new scheme properly announced to the OS?
Good catch, thank you @rgaudin.

I have updated the plist file.

rgaudin (approved):
LGTM now.

PR #963, opened 2024-09-03
Fix search snippet color

HiroyasuNishiyama:
In dark mode, search result snippets are difficult to read because they are displayed in black on a grey background as shown below.

Simulator Screenshot - iPad Air (4th generation) - 2024-09-03 at 10 28 17

This happens on iPads and iPhones.
This PR tries to resolve the issue.

### DIFF
--- a/SwiftUI/Model/SearchOperation/SearchOperation.swift
+++ b/SwiftUI/Model/SearchOperation/SearchOperation.swift
@@ -14,7 +14,9 @@
 // along with Kiwix; If not, see https://www.gnu.org/licenses/.
 
 import Defaults
-import SwiftUI
+#if os(iOS)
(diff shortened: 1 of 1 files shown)


BPerlakiH:
@HiroyasuNishiyama Thank you for pointing this out. I think fixing it on the view level is less complex, please have a look here:
https://github.com/kiwix/kiwix-apple/pull/965

BPerlakiH (changes requested):
I am suggesting to go with the simpler approach here:
https://github.com/kiwix/kiwix-apple/pull/965

HiroyasuNishiyama:
@BPerlakiH 
Thank you for your suggested fix.
Your simpler fix seems fine.

Issue #962, opened 2024-09-02
[REGRESSION] Opening app with ZIM doesnt open ZIM


rgaudin:
When opening Kiwix by double-cliking a ZIM file or using `open` command, the App starts and displays the last viewed ZIM.

I am expecting the clicked ZIM to be opened instead

rgaudin:
macOS 14.6.1 / Kiwix 3.5.1

kelson42:
@rgaudin @BPerlakiH Do we know if this is a regression?

rgaudin:
Well we had #699 and #850

kelson42:
OK, so this is a regression. To be fixed in priority!

BPerlakiH:
I managed to reliably reproduce it.
There are 2 ways to exit the macOS application:
1) closing all windows (one by one), either via the red X system button (top left corner of the window) or pressing CMD+W on each window. Once the last window is closed, the app will exit.
2) Regardless of the amount of windows open, pressing CMD+Q, will exit the app immediatelly.

So what is the difference?
In case 1) we close the window, and delete the tab / window data, as the user closed it, so it's no longer needed.
In case 2) we save data from the windows, in order to restore it on next launch. This mechanism of restoring the windows is not fully working.
There's a built in SwiftUI way to handle this specific feature, but so far it was working only for so called "[Document based app](https://developer.apple.com/documentation/swiftui/building-a-document-based-app-with-swiftui)".

kelson42:
@BPerlakiH We should secure that both way of quitting trigger the same quit() procedure where we do our stuff.

BPerlakiH:
So the way I see it is:
We can have multiple windows, therefore closing a window, and deleting (not storing) any data related to that, is the correct behaviour.
- closing the last window, closes the application, this is also correct.
- saving the state of the windows and restoring it could be a nice to feature, but it's not working properly on macOS, and I checked it wasn't fully working (not even in the oldest available TestFlight build: 3.4.0 (163)). This implementation is restoring only 1 window, and not all of them, which leads to this bug.

For these reasons, I am proposing, to remove this semi-working window restore solution we have, which solves this specific bug. Then create a new issue, to implement the windows management for macOS. It's really an edge case, since the only way to use this feature is to use CMD+Q or from the App Menu choose Quit Kiwix.

kelson42:
@BPerlakiH Top Prio. Would be really great if you find a way to ensure this is the last time we reopen the issue?
[...]

PR #968, opened 2024-09-07
Fix opening app with ZIM doesn't open ZIM

BPerlakiH:
Fixes: #962 

As mentioned in the ticket description, the easiest way to fix this issue is to disable the non-working window restore solution on macOS.

**Testing: Make sure you have only 1 version of Kiwix installed**. It is possible to have  multiple of them (eg from: AppStore, TestFlight, local build).
and as of [Apple docs](https://developer.apple.com/documentation/xcode/defining-a-custom-url-scheme-for-your-app):

> If multiple apps register the same scheme, the app the system targets is undefined.

meaning that the clicked ZIM file might open in another version of Kiwix, if you have more than 1 installed.

### DIFF
--- a/App/App_macOS.swift
+++ b/App/App_macOS.swift
@@ -136,8 +136,13 @@ struct RootView: View {
             case .reading:
                 BrowserTab().environmentObject(browser)
                     .withHostingWindow { window in
-                        if FeatureFlags.hasLibrary == false {
-                            browser.loadMainArticle()
+                        if let windowNumber = window?.windowNumber {
+                            browser.restoreByWindowNumber(windowNumber: windowNumber,
+                                                          urlToTabIdConverter: navigation.tabIDFor(url:))
+                        } else {
+                            if FeatureFlags.hasLibrary == false {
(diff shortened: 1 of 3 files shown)


kelson42:
@rgaudin Are you able to review/test before merging?

rgaudin:
Tested, works like a charm: opening a ZIM if Kiwix if not open opens Kiwix with single window of said ZIM. When opening a ZIM while Kiwix is open, it add a window with said ZIM.

kelson42:
We should have only one window. It ahould not open a new window, it should open a new tab in existing window.

rgaudin:
> We should have only one window. It ahould not open a new window, it should open a new tab in existing window.

This is following Apple's UI guidelines

kelson42:
This multi windows is going to endlessly create problems. I will open an issue to stop this.

rgaudin:
> This multi windows is going to endlessly create problems. I will open an issue to stop this.

This is how macOS apps behave and are expected to behave. We surely can force a single window but that's at the expense of the user experience.

Issue #861, opened 2024-07-11
Library loading is blocking the main thread


BPerlakiH:
When opening the application, the library loading is still doing some heavy work on the main thread.
If we open the application it is needed to load the library straight away, but there are cases when this is not immediately needed. Eg. when the Kiwix app is not open, and we open it by clicking on a ZIM file in Finder. If no other windows were saved earlier, it will open the content page, therefore the library loading can be either postponed, or done in the background.

The more ZIM files we have in our library the longer we will be stuck on this screen:
<img width="1012" alt="Screenshot 2024-07-12 at 00 32 33" src="https://github.com/user-attachments/assets/fa9848de-a620-4073-9f00-216dc7bdb3dc">

BPerlakiH:
After some investigation, this is not going to be an easy change, as the underlying Objective-C code is not thread safe, so that needs to be replaced with swift code or split into smaller chunks, so we can do the file open / reading part on a different thread. For this reason I changed the milestone for this to be 3.5.0.

PR #974, opened 2024-09-12
861 library loading is blocking the main thread

BPerlakiH:
Fixes: #861 

## Summary:
I measured that the longest operation we have, when loading the library is:
[``[Archive](zim::Archive archive = zim::Archive([url fileSystemRepresentation]);``](https://github.com/kiwix/kiwix-apple/blob/91f8161eb4fb101686d1d63def9fc8df88f94c6c/Model/ZimFileService/ZimFileService.mm#L86)

## The way it worked so far:
We opened all the former files that were added to the local library, and passed through this Archive reading and stored it in memory.

After that was done, opening and displaying (any content really from) that ZIM file was quite quick.

## The problem:
When we have relatively small files, or not much files added to the library, there's no visible issue.
Problems start when we have large files. Eg. a 100 GB wiki ZIM file, is taking time to open with the aforementioned  ``Archive``.

## The solution:
Open the ``Archive`` on demand. Meaning, if we want a content from a ZIM file, only at that point open the ``Archive``, and store it at that point, as opposed to app start. Subsequent requests to the get content (or metadata) from the same ZIM file will be equally quick as before. This makes the app start a lot faster, since we do not need to open those archives anymore on start.

## Handle the side effects:
The side effect is that when we press on a button (eg. to open the home page of a local library ZIM file), won't open it right away, so we need some UI loading indicator that we are "working" on it.

For this I have added 2 new classes, ``AsyncButton`` and ``AsyncButtonView``, which can handle alternative loading states, while the underlying task is working.

## Handle concurrency:
Formerly most of the ZIM file operations were running on the main thread, which was safe in a way that no other threads were accessing the same: list of ``Archive`` or other IO operations outcomes.
The solution for this is a new ``@globalActor``, that is: single instance, and only allows async access to the ``ZimFileService`` instance, and this way guarantees that the stored data within is only accessed in a thread safe manner.

## Preview on mac:
https://github.com/user-attachments/assets/07a4e79b-337c-411b-af43-7a8eaf3eb89a

## Adjusting search:
The other point to solve was searching. By default we can search in any local ZIM file, or to be precise the ones that were selected by the user to search in. Since the Search is initialised on start, it made sense to do the opening of the ZIM files at that point. This way we don't have a race between app start opening the ZIM archives and the user starting to search. This has also other benefits, as if the user excludes a ZIM file from search results (which is saved to the next app session), we do not need to open that archive. As soon as the user is (re)including that file in search results, the archive will be opened. The asynchronous part of this is handled during the search result loader / spinner:

Screenshot 2024-09-14 at 18 00 03

Screenshot 2024-09-14 at 17 58 52


BPerlakiH:
This is how it looks on mac:
https://github.com/user-attachments/assets/07a4e79b-337c-411b-af43-7a8eaf3eb89a

rgaudin (dismissed):
LGTM ; very useful improvement I'd say.

My only concern being on the UI. Now that it works well, could we imagine the visual spinner to be bound by the button's dimensions?
In the library card example, it works well because the loader is contained inside the card.
On the open homepage button, it doesn't because it's floating on top with the same (larger than the button) dimensions.

Notye: this comment is based on the screencast.

BPerlakiH:
@rgaudin the progress bar issue has addressed:


https://github.com/user-attachments/assets/04ac9d63-50d0-408c-9c24-36b79b8b5082

rgaudin:
Yep, looks good now

kelson42:
@BPerlakiH Can you please rebase?

PR #979, opened 2024-09-15
Remove duplicate entry from search results

HiroyasuNishiyama:
The search results sometimes contain two identical entries, one with the snippet and another without it.
This is because index search results and title search results are treated separately.
This PR attempts to remove the duplicate results.


BPerlakiH:
Thank you @HiroyasuNishiyama for bringing this up. I am looking into it, how it could be handled the most performant way.

BPerlakiH:
@HiroyasuNishiyama Please see my proposition of solving this here: https://github.com/kiwix/kiwix-apple/pull/981

kelson42:
Superseeded by #981

Issue #980, opened 2024-09-15
Duplicate entries in search results


BPerlakiH:
Based on the findings from: https://github.com/kiwix/kiwix-apple/pull/979

We can have multiple search results, as indexed search and title search can return the same entries.
Additionally some entries will have duplicate URLs, one with and one without a trailing slash, eg:
``
zim://B2A69C3D-7852-400F-9A07-8986875DF683/solar.lowtechmagazine.com/tags/speed/
zim://B2A69C3D-7852-400F-9A07-8986875DF683/solar.lowtechmagazine.com/tags/speed
``
Whereas both will result in the same search result, and page linked.

Example on macOS:

<img width="1552" alt="Screenshot 2024-09-15 at 13 52 05" src="https://github.com/user-attachments/assets/eddb01db-ae75-4ba4-8eca-6ef8d2e540f1">

kelson42:
Here a general remark. The libzim provides two searches:
* Title suggestions
* Fulltext searches

Usually it is either one or the other. Only the Apple reader does somehow a mix. I'm not super found of this approach honestly as it creates a lot of new challenges.

Not against fixes this one obviously, but just want to share the information that this approach might disappear in the future.

BPerlakiH:
@kelson42 I did some further investigation on this. There are more issues discovered, we have this:
https://github.com/kiwix/kiwix-apple/blob/90931345a9ad8c8bd61530f7d86891b15e2e0fda/SwiftUI/Model/SearchOperation/SearchOperation.mm#L62-L65

Now this indeed means we do search in both places, as you wrote.

Currently, with the wikipedia copy I have, it throws an exception on indexed search:
``DatabaseCorruptError: dir_end invalid in block 28240``
Which has the following consequences, if I do change it as you suggested: to be either indexed or title search:
```
if (archive.hasFulltextIndex()) {
    indexSearchArchives.push_back(archive);
} else {
    titleSearchArchives.push_back(archive);
}
```
it won't give any results, since the indexed search fails, and we won't do the title search at all.

Additionally to this, I did found that we do the search on a set of archives, which is also not perfect: 
- if it throws an exception on 1 archive from the set, we loose the results from the whole set!

I am updating the PR to do it one try / catch per archive. That way we can continue to get results even if one of the archives is "bad".

PR #981, opened 2024-09-15
Remove duplicate search entries

BPerlakiH:
Fixes: #980 
Superseeds #979

What we really want is unique search results, which is in fact unique URLs that the user can click on and navigate further. If 2 search results are linking to the same URL, that's a duplicate.

## Solution
Store a set of ``foundURLs`` as we are processing the search results (both indexed and title search), and look for duplicates. Also make sure we adjust the URL endings to always contain the trailing "/" slash, but this is only for our comparison, the returned search result URLs should not be modified.

## After the fix on macOS:

Screenshot 2024-09-15 at 13 47 36

### DIFF
--- a/SwiftUI/Model/SearchOperation/SearchOperation.h
+++ b/SwiftUI/Model/SearchOperation/SearchOperation.h
@@ -21,7 +21,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, strong) NSString *searchText;
 @property (nonatomic, assign) BOOL extractMatchingSnippet;
-
+@property (nonatomic, strong) NSMutableSet *foundURLs;
 @property (nonatomic, strong) NSMutableOrderedSet *results NS_REFINED_FOR_SWIFT;
 
 - (id)initWithSearchText:(NSString *)searchText zimFileIDs:(NSSet *)zimFileIDs;
--- a/SwiftUI/Model/SearchOperation/SearchOperation.mm
+++ b/SwiftUI/Model/SearchOperation/SearchOperation.mm
@@ -76,10 +76,10 @@
         std::string zimFileID_C = [[[zimFileID UUIDString] lowercaseString] cStringUsingEncoding:NSUTF8StringEncoding];
         try {
             auto archive = allArchives->at(zimFileID_C);
-            [self addTitleSearchResults:archive count: 25];
             if (archive.hasFulltextIndex()) {
                 [self addIndexSearchResults:archive count: 25];
             }
+            [self addTitleSearchResults:archive count: 25];
         } catch (std::exception &e) {
             NSLog(@"perform search exception: %s", e.what());
         }
--- a/SwiftUI/Model/SearchOperation/SearchOperation.mm
+++ b/SwiftUI/Model/SearchOperation/SearchOperation.mm
@@ -26,6 +26,24 @@
 #import "SearchResult.h"
 #import "ZimFileService.h"
 
+@interface NSURL (PathManipulation)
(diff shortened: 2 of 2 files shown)


HiroyasuNishiyama:
@BPerlakiH Thank you for your fix.

I think one issue is how to handle the score of search results when there are duplicate entries. 
In the original code, title search results are ranked higher than index search results and are displayed higher in the result list. 
Since the title search result score seems more natural, I tried to reflect that rather than simply dropping it in my fix.

BPerlakiH:
Thank you @HiroyasuNishiyama, that is a good point.
I did addressed 2 concerns in my latest changes, one described in the [issue itself, about failing archives, and loosing results on those](https://github.com/kiwix/kiwix-apple/issues/980#issuecomment-2351699038).

I also changed the order of execution, so we do the title search first, and then the indexed search, this way if we drop duplicates we do that from the indexed ones, leaving the ``probability: null``, which I think would be equivalent to your solution.

Please note that we do another sorting on these results here:
https://github.com/kiwix/kiwix-apple/issues/980#issuecomment-2351699038

HiroyasuNishiyama:
@BPerlakiH 
I think simply changing the order to do title search first introduces another problem.
The results of title search do not include snippets, so if there is a duplicate entry, the snippet will be dropped from the result.

And I realized that my fix had a misunderstanding.
To match the current search scores, I think the `probability` value should be set to the larger of 0.75 (i.e. the solution of log(7.5576 - 6.4524 * x)=1) or the title search result probability rather than setting it to `nil`. 
Though, I know it's ugly to introduce such a constant into this code.

rgaudin:
I think it would be good to have @kelson42's opinion here, in regard to how this should compare with other readers.

kelson42:
At this stage, I would recommend to stop investing more time on this PR. Like I suspected, the solution is not straight.

We should think out of the box about this. IMO whatever we decide do, this is not the role of the reader to do this kind of "smart" computations.

HiroyasuNishiyama:
As a user, I like the current Apple Kiwix search interface, where I do not have to worry about the indexing method.
And I would prefer that redundant entries be excluded from the results from the search results.
Of course, it is up to your team to decide what design choices to make.

For reference, I would like to leave my current thoughts on this PR.
My first PR and this PR removes redundant search results in the Objective C++ module (`SearchOperation.mm`). 
Since this causes issues in handling the deletion results as mentioned above, I think it would be better to delete identical entries so that the first entry remains after the sorting process in swift (`SearchOperation.swift`) shown by @BPerlakiH . 
As an addition, we need to add a process to leave the snippet of the removed entry.

BPerlakiH:
Ok, to sum up what's in this PR already:
- removing duplicate entries from search results (I reversed the title/index now, so snippets are there)
- try / catching errors on searches by each ZIM file (and search type), this way we avoid no results (or less results) if one of the ZIM files is invalid.

For future discussion:
- we could improve to use only one way of searching, if there are absolutely no results from that, maybe only in that case fall back to another type of search
- possible improvement on the OpenZIM side, is to have snippets for title searches as well, eg. returning the first sentence / paragraph of the article (currently we get back the title as the snippet, which we can not / do not use).

IMHO: this PR is already an improvement from what we had. I also do agree with @kelson42 that this is still not an ideal place to be, and a broader / more universal approach is needed for search handling, that would unify (cross platform) and simplify the reader side.

kelson42:
@BPerlakiH @rgaudin @HiroyasuNishiyama I will merge this PR because it seems to fix the bug reported originaly. For the long standing problem/improvement, I have created [a dedicated issue](https://github.com/kiwix/kiwix-apple/issues/983).

Issue #988, opened 2024-09-21
Remove OrderedCollections dependency


BPerlakiH:
We have 2 swift package (external) dependencies.
One of them is OrderedCollections, which is a large amount of extension from Apple, out of that we only use 1 file, and not even to it's full extent, as the most interesting part of it is used, when we run out of memory.

Proposition: instead of using a whole library, just do a simple class that does the same work.

kelson42:
@BPerlakiH From your description, it is not clear to me what is the problem with current situation. Can you please describe it? Having a lot of dependencies is not the best situation, but recoding on your own part of the functions is a questionable approach. What feature exactly is concerned? What do you need tomre ode?

kelson42:
The inconveniences of current solutions are the usual ones, but here the additional code necessary to replace this library is very small. Therefore this might make sense.

PR #989, opened 2024-09-21
Remove OrderedCollections dependency

BPerlakiH:
Fixes: #988 

One of our dependencies is OrderedCollections, which is a large amount of extension from Apple, out of that we only use 1 file, and not even to it's full extent, as the most interesting part of it is used, when we run out of memory, and try to remove the "older" entries from the cache.

Proposition: instead of using a whole library, just do a simple class that does the same work.


rgaudin (approved):
Seems simple enough indeed

kelson42:
@BPerlakiH CI does not pass!

Issue #994, opened 2024-09-26
Top bar should be masked / disappearing during reading


kelson42:
Like on browsers (Safari and Firefox). For the moment this is always visible
<img width="1080" alt="image" src="https://github.com/user-attachments/assets/9bcd6098-573b-4484-86c6-4f6f9bdb48ef">

PR #995, opened 2024-09-30
994 - Extend 'immerse reading' for iPad as well

BPerlakiH:
Fixes: #994

Extending the `immerse reading` that was done for iPhone for iPad as well.

https://github.com/user-attachments/assets/6559c470-07a8-438a-aa72-e7b36472cfe5


Views/BuildingBlocks/WebView.swift
```diff
@@ -130,15 +130,11 @@ final class WebViewController: UIViewController {
 // MARK: - UIScrollViewDelegate
 extension WebViewController: UIScrollViewDelegate {
     func scrollViewDidScroll(_ scrollView: UIScrollView) {
```
BPerlakiH, line 132: This whole block of this file is already under ``#if os(iOS)`` so we are good in terms of macOS.

Issue #993, opened 2024-09-26
Deleting a book should delete related tabs


kelson42:
But it does not!
![IMG_2716](https://github.com/user-attachments/assets/cd5315fd-e326-4b3a-8eac-9f9d034d0997)

BPerlakiH:
I couldn't recreate this on iPad 17. The tabs related to the ZIM file that was deleted are closing properly. 
Can you provide more specific steps please?

kelson42:
I don't remember, but might that be that this is related to "unlink" vs "delete" book?

BPerlakiH:
Ok, with "unlink" it's also working on iPad 17, will switch to iPad 18, and see how that works.

PR #999, opened 2024-10-05
Avoid crash on iOS 18

BPerlakiH:
Related to: #993 

On iPad 18.1 the following webView function is deprecated, and leads to a crash:
Screenshot 2024-10-05 at 21 25 58

Whereas deleting the book does work on 18.1 (simulator):

https://github.com/user-attachments/assets/bdd5f1a3-c8c5-4604-b7d6-3c5104b73d3b

and it seems that unlinking the book also works on 18.1 (simulator):

https://github.com/user-attachments/assets/690a6112-e375-4880-97c8-5e0f1f8d75ac

Will test on device as well..


BPerlakiH:
Tested on iPadOS 18.0.1, and both delete and unlink works as expected.

Issue #1003, opened 2024-10-07
Deleting ZIM file should stop playing pip video


kelson42:
If I play a video in pip and delete the source ZIM from the library, then the video continues. It should be stopped.

kelson42:
@BPerlakiH Is that doable?

BPerlakiH:
@kelson42 I am looking into it. It seems we need to close the tabs related to the ZIM content, which eg. in case of macOS is not happening at the moment.

kelson42:
@BPerlakiH This was impacting my iPad

BPerlakiH:
Thank you, I did found a way to handle this on the iPad. I do see related issues on the iPhone and macOS which behave differently, so I will investigate those further.

kelson42:
I obviously don't understand all the details but deleting/unlinking a ZIM file should delete all related items: bookmarks, library entries, tabs... etc.

I would add that should be better be implemented via constraints (for example a tab can not exist if the related library entry does not exist) than imperatively (when I delete a tab, I should check if there is a pip running). Is that doable?

BPerlakiH:
> I obviously don't understand all the details but deleting/unlinking a ZIM file should delete all related items: bookmarks, library entries, tabs... etc.
> 
> I would add that should be better be implemented via constraints (for example a tab can not exist if the related library entry does not exist) than imperatively (when I delete a tab, I should check if there is a pip running). Is that doable?

I will have a look. The things that are reflected in the local DB eg: a Tab, is more closer to those constraints, which you described, whereas some other parts (BrowserViewModel, WKWebView, Picture in Picture) that are more loosely coupled. Some of these are returning pain-points to be honest, so I will think about them, how to do it better for the future.

PR #1005, opened 2024-10-12
Close Picture in Picture if the ZIM file is deleted/unlinked

BPerlakiH:
Fixes: #1003

## What has changed:
- NavigationView updated to NavigationStack, as per [Apple's recommendations](https://developer.apple.com/documentation/swiftui/migrating-to-new-navigation-types)
- `class` functions changed to `static`. They are essentially the same, it's more of a lint issue
- SwiftUI ``@State var`` were changed to private scope, it's also recommended (not to accidentally change it from the outside)
- `NavigationItem`'s `.onBoarding` renamed to `.welcomeScreen` to be inline with the name of the view where it navigates to
- Fixed a couple of `@MainActor` (concurrency) issues

## iOS:
- Fixed issues with deleting tabs, there was a possibility to end up without any tabs on iPad (if the deletion of the last tab happened while something else was selected):

https://github.com/user-attachments/assets/87ce46cc-1c62-49fb-8927-8fcd673b166a

- On unlinking / deleting a ZIM file, all the related tabs are closed, the `Tabs` entities are deleted, and the cached `BrowserViewModels` are removed, which removes the relevant WKWebViews. The Picture in Picture video started from the closing tab is suspended and stoped.
- Note: iOS (both iPhone and iPad) works in a way that we have the "illusion" of tabs, we only really display 1 tab at a time. The surrounding views of the WKWebView are re-used / re-drawn when switching tabs. SwiftUI is designed this way, which in itself is not bad. The problem is that re-creating the WKWebView is very slow (I tested it), and noticeably bad for the user. For this reason we do cache the WKWebView itself, and some of its baggage in the form of `BrowserViewModel`. Taking the whole view out of cache and inserting into our redrawn `Content` is way more quicker.
- There's another issue with WKWebView, it's almost impossible to "reset" it, and get it back to its original state. Once I was close to achieve it, but the web history (back forward buttons) cannot be edited (we met this problem before).
- So the solution here is to delete the tab, and the cached BrowserViewModel, and create a new one (if needed).

## MacOS
This was a harder case. Here we do not have the "illusion of tabs", we do have them in full instance named windows. In practice we can display them even side by side. From the perspective of the code: all the views and related models are duplicated (almost a full copy of the application) for each window / tab. Anything that is sent via Notification centre (eg. open url), will be sent to all windows. Solution here is to either identify who sent it, or filter the receiver. Anything that is static instance, will be shared amongst the windows / tabs.
Unfortunately SwiftUI is not giving easy access to the current window, or from the window the individual views or view models.
- The former implementation was a bit of mix, we had the "reading" state for the current window, but the WKWebView was tied to the window, meaning it was created once, never to be released. This made the iOS tab solution only partially working, since there was no way to either "recreate" the `WKWebView`, or to clear it out completely.
- Solution: I reused the same method of `BrowserViewModel` caching from iOS, with a bit of extra `static` workaround for the case when we want to open a `url` in a new window / tab. Since in this case we need to "link together" the newly created tab's id (which is also used a cache key for BrowserViewModel), and the window itself, so we know what tab id we should use in the new window (and application instance) for getting the `BrowserViewModel`.

## Follow ups: 
- [the macOS window restoration never really worked](https://github.com/kiwix/kiwix-apple/issues/1015), and that is still not working at the moment, although after all these changes (and knowledge gained) it will be easier to achieve it, if needed. It's a rather low priority issue, as the only way to achieve it is to have something open and CMD+Q the app. If all the windows will be closed, one by one, we get back to the "initial" one window mode after relaunching the app.
- even if the tab is closed/deleted and the related BrowserViewModel is cleared from the cache, [it stays around in the memory (a leak)](https://github.com/kiwix/kiwix-apple/issues/1016). One of the reasons, I have found so far, is that it's easy to put a class into Swift UI's `@Environment`, but there's no easy way to take it out from there (so far I haven't found a solution for that).


kelson42:
@BPerlakiH Any news here?

Issue #971, opened 2024-09-08
Release tag seems wrong


kelson42:
For some reason it is `testflight-3.5.2` and it should be `3.5.2`

Unclear - at least right now - to me why we don't have a proper tag.

BPerlakiH:
It seems to me that we are re-using the tags that were used to created the TestFlight releases (and which was approved by Apple), and we do not create a new tag.

kelson42:
To me the workflow should be as such:
* Tag `testflight` when you want to do Testflight release. If this tag already exists, then remove/superseeds older version.
* If we make a release tag `x.y.z` and release in GitHub.
* Released version trigger the publishing in testflight but only such a version can make it through production.
* If there is a problem with Apple validation - post release - then a patch release has to be made.

BPerlakiH:
I've updated the readme and the CD workflow to reflect the above requirements.

Issue #753, opened 2024-05-07
Implement Apple Pay and use the “Donate with Apple Pay” button for "Support Kiwix"


BPerlakiH:
Following Apple's recommendations for non-profit organisations:
https://developer.apple.com/apple-pay/nonprofits/
We should implement Apple Pay and use the "Donate with Apple Pay" button.

This should be done smartly so it works fine for custom apps as well

kelson42:
Only Appke store pay should work with Applepay, other build (dmg) should have the traditional link.

Popolechien:
I have applied with Benevity/Apple Pay to get us approved.

rgaudin:
@Popolechien any news on this? Should we move that to 3.5.0?

kelson42:
@Popolechien We need an update here before implementing the feature.

Popolechien:
![Screenshot 2024-09-18 at 12 07 36](https://github.com/user-attachments/assets/b0b59410-e19e-4d16-9093-9a7c7635ebd5)
No news. I have resubmitted, made a note to ping them on Monday 23rd, but seeing as a bunch of contact links are wrong/deprecated/broken I am cautiously optimistic.

Popolechien:
From Apple:
> Hi team,    
> We are excited to let you know that your organization has been approved to accept donations with Apple Pay. 
> You can now submit your iOS app to the App Store for review or deploy your website’s integration of Apple Pay. If you’re in the process of integrating Apple Pay into your app or website, make sure to review the guidelines and resources on the Apple Developer website. 
> If you have any questions on implementing Apple Pay, you can talk to other developers and Apple engineers on the Apple Developer forums or contact us. 
> Best Regards, 
> Apple Developer Relations

kelson42:
@BPerlakiH You need anything else to move forward with the code?

BPerlakiH:
Thank you @kelson42, I think initially it should be good, if anything arise while implementing it, I will let you know.

BPerlakiH:
I've found the following things we need:
https://causeshelp.benevity.org/hc/en-us/articles/360001149486-Benevity-and-Apple-Pay

> To accept donations with Apple Pay, your organization will need the following:
> Payment provider
> Apple Developer Program membership
> Approval to accept donations with Apple Pay
> Apple Pay implemented in your app or website

My question is do we have a payment provider? 
[...]

PR #1010, opened 2024-10-17
971 release tag seems wrong

BPerlakiH:
Fixes: #971

I've updated the Readme file and the CD workflow to have our new approach documented and enforced.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -2,8 +2,8 @@ name: CD
 
 # | Platform        | Nightly | Weekly (if diff HEAD)  | TestFlight Tag | Release          |
 # |-----------------|---------|------------------------|----------------|------------------|
-# | macOS           | ftp     | app-store              | app-store      | ftp              |
-# | iOS             | ftp     | app-store              | app-store      | -                |
+# | macOS           | ftp     | app-store              | app-store      | app-store, ftp   |
+# | iOS             | ftp     | app-store              | app-store      | app-store        |
 # |-----------------|---------|------------------------|----------------|------------------|
 # | VERSION         | Date    | get from project.yml   | project.yml    | project.yml      |
 # |-----------------|---------|------------------------|----------------|------------------|
@@ -29,7 +29,7 @@ env:
   # conditionally updated later:
   EXPORT_METHOD: "app-store"
   EXTRA_XCODEBUILD: ""
-  UPLOAD_TO: "" # !important
+  UPLOAD_TO: "" # !important this filters down the matrix combinations
   VERSION: ""
   SIGNING_CERTIFICATE: ""
   SIGNING_CERTIFICATE_P12_PASSWORD: ""
@@ -40,6 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [iOS, macOS]
+        destination: [app-store, ftp]
     runs-on: macos-13
 
     steps:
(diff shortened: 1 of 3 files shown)


kelson42 (changes requested):
@BPerlakiH I believe this was still not clear to you how I want it to work.

I have fixed the `README.md` but:
* The section "Last Step" is probably still wrong
* The worflow is still wrong for the release.

There is no something like a "post release step", this is one release process which is triggered by the GitHub release and then go to testflight. It's not testflight and then github release.

BPerlakiH:
Ohh, it is still not clear for me...
This is my understanding of it:
We need to build and upload a binary build for Apple to review.
We do this via CD (either weekly or with the TestFlight tag).
For the AppStore review we select (one of the) binaries already uploaded (usually the latest).
Once the binary is approved, we cannot change it to something else, eg. triggering the Github release -> build new binary -> upload to App Store Connect.
So my understanding is that once the binary (previously built) is approved by Apple, we need to tag it on Github as a release. This would be sufficient enough, but we want a dmg file as well. So we rebuild it from the same commit but only for FTP .dmg file.

kelson42:
> For the AppStore review we select (one of the) binaries already uploaded (usually the latest).

No, this has to be one we have released (so we a a clear release number).

BPerlakiH:
> > For the AppStore review we select (one of the) binaries already uploaded (usually the latest).
> 
> No, this has to be one we have released through the usual process. The validation of Apple does not decide what we release. We decide what we release and Apple decides if it goes through to the audience of its App store.

Ok, I see your point now. This is how I understand it:
1) we can create testflight builds as many we like, it won't affect either the public release to the app store or the GitHub release, or the FTP release
2) for each App Store release, we trigger an appropriate Github release, so for example:
- we Github release 3.6.0 that creates a github tag, builds everything and uploads it to Apple, Apple reviews it the next day or so
- if 3.6.0 is rejected 
- we change some code
- create Github release 3.6.1, that builds everything uploads it to Apple, Apple reviews it, and approves the next day
- we press the public release button on App Store Connect, so it's available for everyone publicly
- at this point, we still need a "post App Store release step" to create and upload the .dmg binary to FTP, to make that publicly available and the very same that is available in the App Store.

BPerlakiH:
@kelson42 After re-reading your changes I think we are missing some pieces, to get a common understanding of this process.

We can create a release version, eg: 3.6.0. This is stored in git within the code. As time passes we weekly / or on demand create builds for this, that is available to try out. For clarity let's call them release candidates. This possibility of trying them out is available via TestFlight. The idea is that it should be tested on real devices before the public sees it.
Under each release version, we have several builds, eg this is for 3.6.0:

Screenshot 2024-10-20 at 00 59 18

Those builds are uploaded there via our TestFlight CD process. Any of those builds can be picked for a version to be released. This is what I meant that we usually pick the latest one.

So far we were following the safest path, that the build (release candidate) we try out is the one we send to Apple. If it's approved, we want to mark it as a "final" release, and as such it should be reflected in GitHub release and as a tag as well. At the same time, we do want to upload this to FTP so it's available there as well. These 2 last steps is automated, by the formerly called "post Release step": 
1) "visible" as a Github release
2) uploaded to FTP
maybe we should call it 'post App Store step'.

From all of this, what we haven't really practiced so far is creating the proper GitHub release tag, and created too many "testflight" tags with various suffixes.


Now we can turn this around, and do a new build and tag as a GitHub release, which will produce a new build on that list, and retest that, and pick it to be published to the App Store. If all is good, it still requires us to do the FTP step, once it's approved.
With this approach though we might end up with more GitHub release versions, than App Store versions, since we will GitHub release and tag even the ones that were not approved. 

Eg: version 1.6.0 is not approved by Apple:

| App Store:| GitHub:|
|-------|-------|
| 1.6.1 | 1.6.1 |
| 1.5.0 | 1.6.0 |
|       | 1.5.0 |

BPerlakiH:
In this whole Apple process of getting from code to public App Store, the release version is fixed from the very beginning. The build number varies though.
We can go against that, and say if Apple rejects a specific version we start all over and create a new one with a patch number higher, that's OK with me. I would still prefer to have matching App Store released versions with GitHub released versions.

kelson42:
> So that the "release" means it is publicly available both on App Store

We primarely release a code (version), all the rest is secondary.

rgaudin (approved):
LGTM

rgaudin (approved):
👍

kelson42:
I don't know if this is ready to review but at a first look I don't see how the release event would here push on testflight.

README.md
```diff
@@ -146,25 +146,50 @@ https://developer.apple.com/documentation/safari-developer-tools/inspecting-ios
...
+Once satisfied with the quality of the app in TestFlight, we can
+proceed with the Kiwix release process.
+
+The release process is triggered by a [GitHub
+Release](https://github.com/kiwix/kiwix-apple/releases). Only the
+version released (to Testflight) that way can be then sent to
+approval to Apple. Once approved by Apple, we can release them to the
+AppStore. At the same time, will publish the macOS DMG to our [file
```
rgaudin, line 181: This *At the same time* is confusing. It should be clear that this is triggered by the Github release event.

rgaudin:
> I don't know if this is ready to review but at a first look I don't see how the release event would here push on testflight.... Or it was already doing it in the past?

Indeed, I thought this was a doc-only PR. What is described in this PR's readme is not what is happening in the workflow. We are setting UPLOAD_TO differently. This will require refactoring of the workflow as it was previously either one or the other

kelson42:
@BPerlakiH Can you please focus on fixing this? This is a very old PR and I want this working fine before releasing.

README.md
BPerlakiH: Thanks @rgaudin. I agree, please see it updated.

.github/workflows/cd.yml
```diff
@@ -50,6 +52,10 @@ jobs:
     - name: Install python dependencies
       run: pip install pyyaml==6.0.1
 
+    - name: Remove Apple Pay capability for macOS
```
BPerlakiH, line 55: This is to fix: #753 builds

.github/workflows/ci.yml
```diff
@@ -41,6 +41,10 @@ jobs:
         if: matrix.platform == 'iOS'
         run: echo "EXTRA_XCODEBUILD=-sdk iphoneos ${{ env.APPLE_AUTH_PARAMS }}" >> $GITHUB_ENV
```
BPerlakiH, line 43: Also to fix: #753 builds

rgaudin (dismissed):
LGTM but I find it a tad less readable.

I'm not enjoying the use of bash array for `UPLOAD_TO`. We've been using the GITHUB_ENV solely as a mean to store data between steps. It was dumb on purpose.

I dont know if the [`contains()`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#contains) expression is matching on it as an array or as strings.

@BPerlakiH could you please carify? If this matches as string, then we can simplify the syntax and use `UPLOAD_TO="app-store ftp"` instead of `UPLOAD_TO=("app-store" "ftp")`.

BPerlakiH:
> LGTM but I find it a tad less readable.
> 
> I'm not enjoying the use of bash array for `UPLOAD_TO`. We've been using the GITHUB_ENV solely as a mean to store data between steps. It was dumb on purpose.
> 
> I dont know if the [`contains()`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#contains) expression is matching on it as an array or as strings.
> 
> @BPerlakiH could you please carify? If this matches as string, then we can simplify the syntax and use `UPLOAD_TO="app-store ftp"` instead of `UPLOAD_TO=("app-store" "ftp")`.

I also think it's less readable than it was before... This was the least destructive way to change it. We have both EXPORT_METHOD and the Signing Certificate different when it comes to macOS ftp vs macOS App-Store, so without changing the XCodeBuild step as well we need to run those in "parallel".

I checked the ``contains`` method, and it is in fact does string matching, so I changed it accordingly.

rgaudin (approved):
All good now

Issue #1018, opened 2024-11-01
Get the logo size from Assets, instead of hard coding it


BPerlakiH:
Related to: https://github.com/kiwix/kiwix-apple-custom/pull/73

PR #1019, opened 2024-11-01
Get image size from assets, instead of hard coding it

BPerlakiH:
Fixes: #1018


kelson42:
@BPerlakiH Can you please rebase?

PR #1022, opened 2024-11-03
753 donate with apple pay button for iOS to support Kiwix

BPerlakiH:
Fixes: #753

Should be tested with real payments to see if it works end to end.

### DIFF
--- a/App/App_macOS.swift
+++ b/App/App_macOS.swift
@@ -19,11 +19,15 @@ import Combine
 import Defaults
 import CoreKiwix
 import PassKit
+import StripeApplePay
 
 #if os(macOS)
 final class AppDelegate: NSObject, NSApplicationDelegate {
     func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
         true
+    }
+    func applicationDidFinishLaunching(_ notification: Notification) {
+        StripeAPI.defaultPublishableKey = Payment.stripePublicKey
     }
 }
 
--- a/App/App_macOS.swift
+++ b/App/App_macOS.swift
@@ -182,11 +182,11 @@ struct RootView: View {
                 }
             }
             .frame(minWidth: 160)
-            .safeAreaInset(edge: .bottom) {
-                SupportKiwixButton {
-                    openWindow(id: "donation")
-                }
-            }
+//            .safeAreaInset(edge: .bottom) {
+//                SupportKiwixButton {
+//                    openWindow(id: "donation")
+//                }
+//            }
         } detail: {
             switch navigation.currentItem {
             case .loading:
--- a/App/App_macOS.swift
+++ b/App/App_macOS.swift
@@ -32,6 +32,8 @@ struct Kiwix: App {
     @NSApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
     @StateObject private var libraryRefreshViewModel = LibraryViewModel()
     private let notificationCenterDelegate = NotificationCenterDelegate()
+    private var amountSelected = PassthroughSubject<SelectedAmount?, Never>()
+    @StateObject var formReset = FormReset()
 
     init() {
         UNUserNotificationCenter.current().delegate = notificationCenterDelegate
@@ -80,6 +82,31 @@ struct Kiwix: App {
             }
             .frame(width: 550, height: 400)
         }
+        Window("Donate", id: "donation") {
+            PaymentForm(amountSelected: amountSelected)
+                .frame(width: 320, height: 320)
+                .onReceive(NotificationCenter.default.publisher(for: NSWindow.willCloseNotification)) { notification in
+                    if let window = notification.object as? NSWindow,
+                       window.identifier?.rawValue == "donation" {
+                        debugPrint("closing donation")
+                        formReset.reset()
+                    }
+                }
+                .environmentObject(formReset)
+                .onReceive(amountSelected) { amount in
+                    debugPrint("amountSelected: \(amount)")
+                    // after upgrading to macOS 14, use:
+                    // @Environment(\.dismissWindow) var dismissWindow
+                    // and call:
+                    // dismissWindow(id: "donation")
+                    NSApplication.shared.windows.first { window in
+                        window.identifier?.rawValue == "donation"
+                    }?.close()
+                }
+        }
+        .windowResizability(.contentSize)
+        .windowStyle(.titleBar)
+        .commandsRemoved()
     }
 
     private class NotificationCenterDelegate: NSObject, UNUserNotificationCenterDelegate {
@@ -99,6 +126,7 @@ struct Kiwix: App {
 }
 
 struct RootView: View {
+    @Environment(\.openWindow) var openWindow
     @Environment(\.controlActiveState) var controlActiveState
     @StateObject private var navigation = NavigationViewModel()
     @StateObject private var windowTracker = WindowTracker()
@@ -134,17 +162,20 @@ struct RootView: View {
                     usingNetworks: Payment.supportedNetworks,
                     capabilities: Payment.capabilities
                 ) {
-                    PayWithApplePayButton(
-                        .donate,
-                        request: payment.donationRequest(),
-                        onPaymentAuthorizationChange: payment.onPaymentAuthPhase(phase:),
-                        onMerchantSessionRequested: payment.onMerchantSessionUpdate
-                    )
-                    .frame(width: 200, height: 30, alignment: .center)
-                    .padding()
+                    SupportKiwixButton {
+                        openWindow(id: "donation")
+                    }
+//                    PayWithApplePayButton(
+//                        .donate,
+//                        request: payment.donationRequest(),
+//                        onPaymentAuthorizationChange: payment.onPaymentAuthPhase(phase:),
+//                        onMerchantSessionRequested: payment.onMerchantSessionUpdate
+//                    )
+//                    .frame(width: 200, height: 30, alignment: .center)
+//                    .padding()
                 }
             }
-            .frame(minWidth: 150)
+            .frame(minWidth: 160)
         } detail: {
             switch navigation.currentItem {
             case .loading:
--- a/App/App_macOS.swift
+++ b/App/App_macOS.swift
@@ -18,6 +18,7 @@ import UserNotifications
 import Combine
 import Defaults
 import CoreKiwix
+import PassKit
 
 #if os(macOS)
 final class AppDelegate: NSObject, NSApplicationDelegate {
@@ -109,6 +110,7 @@ struct RootView: View {
     private let tabCloses = NotificationCenter.default.publisher(for: NSWindow.willCloseNotification)
     /// Close other tabs then the ones received
     private let keepOnlyTabs = NotificationCenter.default.publisher(for: .keepOnlyTabs)
+    private let payment = Payment()
 
     var body: some View {
         NavigationSplitView {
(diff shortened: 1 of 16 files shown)


rgaudin:
This is a UI PR ; please include a screencast or screenshots.

BPerlakiH:
https://github.com/user-attachments/assets/89fe0ba8-b3a3-49a0-9bde-7fd0d363cf48


https://github.com/user-attachments/assets/bb1a1e1a-8ccc-44ee-b437-200959accdde


https://github.com/user-attachments/assets/e863485d-5ccc-4909-9f24-8ebb7734fcd4

kelson42:
@BPerlakiH Nice, but CI is red!

BPerlakiH:
From what I suspect, this might require more integration into Stripe, to get the available payment options from there, and possibly make the payments appear on Stripe dashboard, but let's try it without it first.

kelson42:
@BPerlakiH I can not merge like this, what is going on with the CI must be properly understood and treated.

rgaudin:
@BPerlakiH your marchant ID commit 3bee29b fixed the iOS build but not the macOS one. Indeed, it's not clear what's happening and is definitely specific to this PR's changes.

For clarity, I reran the CI workflow on `main` and [it succeeded](https://github.com/kiwix/kiwix-apple/actions/runs/11681859210)

Model/Payment.swift
```diff
@@ -0,0 +1,111 @@
...
+import SwiftUI
+import Combine
+
+struct Payment {
+
+    let completeSubject = PassthroughSubject<Bool, Never>()
+
+    static let merchantId = "merchant.org.kiwix.apple"
+    static let paymentSubscriptionManagingURL = "https://www.kiwix.org"
+    static let supportedNetworks: [PKPaymentNetwork] = [
```
rgaudin, line 27: Where is this from? Is this from Stripe documentation? The fact that we are using a third party (Stripe) makes me think we shouldn't need those things

BPerlakiH: Stripe is not yet added, this form comes from Apple's PassKit framework.

BPerlakiH:
> @BPerlakiH your marchant ID commit [3bee29b](https://github.com/kiwix/kiwix-apple/commit/3bee29ba11ce25285775290a56924117bc57245d) fixed the iOS build but not the macOS one. Indeed, it's not clear what's happening and is definitely specific to this PR's changes.
> 
> For clarity, I reran the CI workflow on `main` and [it succeeded](https://github.com/kiwix/kiwix-apple/actions/runs/11681859210)

@rgaudin 
I think the reason behind it is that the merchant certificate for Apple Pay is iOS only, can we either extend it to be all platform or create another one for macOS ?

rgaudin:
OK, I think we're not aligned with what's in this PR. My understanding (@kelson42 please speak up if I'm mistaken) is that this PR should introduce donations on the App (via Apple Pay).

So when merging this, we would have a working donation UI that results in actual donated money on our Stripe account.

Unless I misunderstood, that's not where we're going.

- I believe we must follow [Stripe's documentation](https://docs.stripe.com/apple-pay) which entails creating a PaymentIntent on Stripe before asking ApplePay to process it.
- That PaymentIntent will require a Stripe API Key. Here's the test-mode one: `pk_test_oglo2v3Wc7ibH2oQe5oUDkhi`. You can use [Stripe test cards](https://docs.stripe.com/testing) in Stripe-mode but for ApplePay, you must use a real card (wont be charged as long as using the test API key)
- It seems quite clear to me that Stripe+ApplePay is only available on iOS. For macOS, the only option is web via Safari AFAICT. In this regard, we should simply open the donation URL on the browser.
- Stripe doc suggest the paymentIntent to be created from a secure location (a new endpoint on one of our server) so that we can control the amount/currency requested. Given this is not a purchase but a donation and we allow users to select the amount, it sounds like unnecessary complications ATM

Obviously this is from reading Stripe's doc and I have no prior ApplePay experience so there might be other ways.

@BPerlakiH please let me know what you think of this

rgaudin:
From what I've seen online, we should be able to remove the entitlement/capability from project and add the merchantID to the Stripe dep config. Have you tried?

kelson42:
@BPerlakiH Unfortunately CI is still red!

BPerlakiH:
@kelson42 @rgaudin with modified CI builds (taking out Apple Pay for macOS) this is passing OK. See here: https://github.com/kiwix/kiwix-apple/actions/runs/11999162541

Therefore I am proposing to merge first: 
https://github.com/kiwix/kiwix-apple/pull/1010
which contains the CI fix for this as well.

Revert notes added to: https://github.com/kiwix/kiwix-apple/issues/1033

Model/Payment.swift
```diff
@@ -0,0 +1,166 @@
...
+        case .willAuthorize:
+            os_log("onPaymentAuthPhase: .willAuthorize")
+            break
+        case .didAuthorize(let payment, let resultHandler):
+            os_log("onPaymentAuthPhase: .didAuthorize")
+            // call our server to get payment / setup intent and return the client.secret
+            // async http call...
+            Task { [resultHandler] in
+
+                let paymentServer = StripeKiwix(endPoint: URL(string: "http://192.168.100.7:4242")!,
```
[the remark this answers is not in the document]
kelson42: I guess this is fixed now

rgaudin (changes requested):
- There are 3 codefactor issues (the JSON one might need an exception)
- Payment was failing on device as I mentioned the other. Was in fact due to currency check on backend that wan't normalized. Fixed now.
- In Settings view, the Support Kiwix _line_ is not clickable (unlike the others) and one must click the _Support Kiwix_ text to trigger the popup.
- Monthly tab is still visible. We should remove all code that is related to recurring. We dont known when or if that will happen so no need to carry this complexity.
- There is still no post-donation feedback for neither success or failure. We clearly decided on a toast-like Thank you notice.
- Why the `supportedNetworks`? Why the `capabilities`? This should all be unknown to us because we are using Stripe. It looks like we never get into into anyway
- If there's a documentation you followed, can you add it as a comment in Payment.swift since we're not using the Stripe SDK (and its documentation)

FYI I switched the config to use production and successfully completed an actual donation. I received notification from my bank and it's visible on Stripe 👍 
**I reverted to test API now**

rgaudin:
https://github.com/user-attachments/assets/8dbc243c-e36f-4dc6-a9f8-63d7e2f66399

Model/Payment.swift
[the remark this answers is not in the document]
BPerlakiH: Yes this is updated to the real one.

BPerlakiH:
> * Payment was failing on device as I mentioned the other. Was in fact due to currency check on backend that wan't normalized. Fixed now.

Thank you!

> * In Settings view, the Support Kiwix _line_ is not clickable (unlike the others) and one must click the _Support Kiwix_ text to trigger the popup.

I have fixed this.

> * There are 3 codefactor issues (the JSON one might need an exception)
> * Monthly tab is still visible. We should remove all code that is related to recurring. We dont known when or if that will happen so no need to carry this complexity.
> * There is still no post-donation feedback for neither success or failure. We clearly decided on a toast-like Thank you notice.

Code factor issues are/were solved, and the thank you pop-up is/was also there, we just lost a good chunk of commits on the last rebase.
I've restored them now.

The monthly options are also turned off for the end-user. I would recommend to leave it in, since the UI part is ready and tested for that. We can experiment with it in terms of back-end integration, and if we decide not to have it, it's going to be an easy PR to take it out.

> * Why the `supportedNetworks`? Why the `capabilities`? This should all be unknown to us because we are using Stripe. It looks like we never get into into anyway.

We could possibly get those from the Stripe API, but I will have a look if they are really required to be defined upfront.

> * If there's a documentation you followed, can you add it as a comment in Payment.swift since we're not using the Stripe SDK (and its documentation).

I've added all the relavant links and docs into Payment.swift.

BPerlakiH:
https://github.com/user-attachments/assets/90cc9022-7cdd-44f5-85ab-5005734e5687


https://github.com/user-attachments/assets/497a2019-84d1-446b-8007-6df5cfbbbd74


Obviously we can tweak the messages to display something else.

rgaudin:
Thank you ; this now works as expected!

- I'm OK with keeping the recurring code (not accessed)
- I'd like you to sort out quickly the paymentMethod/capability thing. I can't see how that would be useful but I can imagine it being problematic.
- @kelson42 @Popolechien  we need your feedback **ASAP** on the thank-you message
- We have no feedback if there's been an error. I believe we should. @kelson42 ? @Popolechien ? At least if the error is when communicating with our backend (so the user is not lead into thinking it's a bank issue on his end if the problem is on our side)


https://github.com/user-attachments/assets/6f6fbe90-7461-4728-9cd5-284ec962e82c


IMG_E83621458087-1

Popolechien:
Thank you message looks good to me. 
Error message could be along the lines of "`Well that's awkward. There has been an issue with your payment. Please try again?`"
Displaying the reason (unless it is a wrong password kind of thing) for the failure probably won't help much as these technical messages are usually pretty absconse.

BPerlakiH:
I've double checked these `supportedNetworks` and `merchantCapabilities`. They work like a filter what should be displayed to the end user, and cannot be empty, eg: if `supportedNetworks` is not set, there's a runtime warning:
`` [General] Payment request is invalid: Error Domain=PKPassKitErrorDomain Code=1 "Invalid in-app payment request" UserInfo={NSLocalizedDescription=Invalid in-app payment request, NSUnderlyingError=0x600000060810 {Error Domain=PKPassKitErrorDomain Code=1 "PKPaymentRequest must contain an NSArray property 'supportedNetworks' of objects of type 'NSArray' within '(`` 
So for now I've added all the possible payment methods, we do not want to limit anyone's generosity, I suppose.

BPerlakiH:
> Have you ever seen the button with the Setup label? I did not.
Yes, I did see the setup button, it is for the case when the user not yet fully configured Apple Pay eg: no cards were added to the user Wallet. In this scenario Apple Pay can be set-up and the user can continue to use our application, and finish the donation.

rgaudin:
> > Have you ever seen the button with the Setup label? I did not.
> 
> Yes, I did see the setup button, it is for the case when the user not yet fully configured Apple Pay eg: no cards were added to the user Wallet. In this scenario Apple Pay can be set-up and the user can continue to use our application, and finish the donation.

OK then that makes sense

BPerlakiH:
I've added the error message pop-up:

Simulator Screenshot - iPhone 8 - 2024-11-30 at 14 04 25

I also merged in the changes for macOS, but the UI for that is still disabled.
Once we figure out the certificate / CI part of it, it can be enabled.

App/App_macOS.swift
```diff
@@ -127,7 +193,12 @@ struct RootView: View {
                     }
                 }
             }
-            .frame(minWidth: 150)
+            .frame(minWidth: 160)
```
BPerlakiH, line 196: Here we disable the support button for macOS

Model/StripeKiwix.swift
```diff
@@ -0,0 +1,118 @@
...
+        // A positive integer representing how much to charge in the smallest currency unit
+        // (e.g., 100 cents to charge $1.00 or 100 to charge ¥100, a zero-decimal currency).
+        // The minimum amount is $0.50 US or equivalent in charge currency.
+        amount = Int(selectedAmount.value * 100.0)
+        currency = selectedAmount.currency
+        assert(Payment.currencyCodes.contains(currency))
+    }
+}
+private struct SessionParams: Encodable {
+    let validationUrl = "apple-pay-gateway-cert.apple.com"
```
BPerlakiH, line 117: @rgaudin: Maybe it would be safer not to have this payload as part of the POST request.
What do you think?

rgaudin (approved):
LGTM (tested)

Model/StripeKiwix.swift
rgaudin: No, it's OK, that's part of the process. It's used to toggle between prod/test and global/china servers.

PR #1025, opened 2024-11-08
Switch to macos-latest GitHub runners

BPerlakiH:
In an attempt to try to fix the macOS build issues from: https://github.com/kiwix/kiwix-apple/pull/1022

Switching to macOS-latest GitHub runners.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -40,7 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [iOS, macOS]
-    runs-on: macos-13
+    runs-on: macos-latest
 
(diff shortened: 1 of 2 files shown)


rgaudin:
We dont want to ever rely on latest, as it's a moving target. ATM `macos-latest` is `macos-14` so please use that.

BPerlakiH:
Well, I have tried with the latest to see if it helps, but there's no difference really. The CI is failing the same way after adding Apple Pay, see here: https://github.com/kiwix/kiwix-apple/pull/1026.

For this reason, I am closing these 2 PRs, as they are not helping at the moment.

Issue #1007, opened 2024-10-15
There seems to be no protection against fetching again an again the same version of the catalog


kelson42:
If I click twice on "fetch catalog" it takes many seconds each time to complete. Which tends to let me think there is no check if the version of the online catalogue is the same as the copy we have locally. If the version would be checked and proven to be the same (via Etag), then Kiwix should not redownload everything and it should be over in 0.1s.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match

kelson42:
@BPerlakiH Can you please confirm/explain how it works?

BPerlakiH:
@kelson42 I can see in the traffic that the catalog is refetched on each button press:
https://library.kiwix.org/catalog/v2/entries?count=-1
It seems to have the correct headers though:
```
Cache-Control: max-age=0, must-revalidate
ETag: "1726580544242481858.2213904/z"
```
The amount of time it takes is divided though. From what I can see the request (on a good wifi) takes only 900ms, whereas parsing and (maybe inserting the data) takes longer, around 2 more seconds.

The libkiwix library is also attempting to download the favicons, which I think is not really needed as it just simply not working, so we download those "on demand", when displaying the list of categories. It is also creating a lot of log outputs:
```
Cannot download favicon from https://library.kiwix.org/catalog/v2/illustration/da7fb796-e4ee-641f-41ff-6d7da49a28cb/?size=48
Cannot download favicon from https://library.kiwix.org/catalog/v2/illustration/4f489a41-d2dd-8066-7136-d21277420e45/?size=48
Cannot download favicon from https://library.kiwix.org/catalog/v2/illustration/f73302be-60bc-b740-53ba-50b1624eb4b9/?size=48
Cannot download favicon from https://library.kiwix.org/catalog/v2/illustration/e2c51a75-6700-cb90-69bd-f22a9de13482/?size=48
Cannot download favicon from https://library.kiwix.org/catalog/v2/illustration/646f6c7f-60da-6a29-ebf6-47384265edce/?size=48
Cannot download favicon from https://library.kiwix.org/catalog/v2/illustration/c9470bd8-571d-4815-76d3-fb4086036613/?size=48
Cannot download favicon from https://library.kiwix.org/catalog/v2/illustration/9f6dcc7e-a1dc-4868-4a8f-30d01a38ba29/?size=48
[...]

PR #1034, opened 2024-11-24
Use eTag for http caching

BPerlakiH:
Fixes: #1007

### DIFF
--- a/ViewModel/LibraryViewModel.swift
+++ b/ViewModel/LibraryViewModel.swift
@@ -213,7 +213,7 @@ final class LibraryViewModel: ObservableObject {
     }
 
     private func process(parser: Parser) async throws {
-        try await withCheckedThrowingContinuation { [weak self] continuation -> Void in
+        try await withCheckedThrowingContinuation { [weak self] continuation in
             Database.shared.performBackgroundTask { [weak self] context in
                 guard let self else {
                     continuation.resume()
--- a/ViewModel/LibraryViewModel.swift
+++ b/ViewModel/LibraryViewModel.swift
@@ -87,6 +87,8 @@ final class LibraryViewModel: ObservableObject {
 
             // refresh library
             guard let data = try await fetchData() else {
+                error = nil
+                process.state = .complete
                 return
             }
             let parser = try await parse(data: data)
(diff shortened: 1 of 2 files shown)


ViewModel/LibraryViewModel.swift
```diff
@@ -171,14 +173,23 @@ final class LibraryViewModel: ObservableObject {
...
     private func fetchData() async throws -> Data? {
         do {
-            let request = URLRequest(url: Self.catalogURL, timeoutInterval: 20)
+            var request = URLRequest(url: Self.catalogURL, timeoutInterval: 20)
+            request.allHTTPHeaderFields = ["If-None-Match": Defaults[.libraryETag]]
             let (data, response) = try await self.urlSession.data(for: request)
             guard let response = response as? HTTPURLResponse else { return nil }
-            guard response.statusCode == 200 else {
+            switch response.statusCode {
+            case 200..<300:
```
rgaudin, line 181: Use `200`. There's no reason to expect other and this could lead to issue

ViewModel/LibraryViewModel.swift
```diff
@@ -171,14 +173,23 @@ final class LibraryViewModel: ObservableObject {
...
+                if let eTag = response.allHeaderFields["Etag"] as? String {
+                    Defaults[.libraryETag] = eTag
+                }
+                // OK to process further
+                os_log("Retrieved OPDS Data, size: %llu bytes", log: Log.OPDS, type: .info, data.count)
+                return data
+            case 300..<400:
```
rgaudin, line 188: Is that for `304`? Then limit it to it.

rgaudin (changes requested):
Mostly OK but those wide status code ranges are unnecessary and fragile

rgaudin:
Question about code base: Are we currently following redirects? Should we return a redirect on the catalogURL (and eventually return the content), would that work as expected?

ViewModel/LibraryViewModel.swift
BPerlakiH: Ok, see it updated.

rgaudin (approved):
Good

Issue #1044, opened 2024-12-03
Strangely behaving Deeplinks


kutulus:
Kiwix v. 3.6.0 (184) MacOS
Deeplinks in an installed wiki behave strangely.
Example:
zim://DE071ED9-7067-6C01-0EB9-3EC704AB5940/A/Europe

1. if Kiwix is not open, the programme get started and the page of installed wikis is displayed.

2. if Kiwix is open (with the page of the installed wikis), the deeplink is opened in a new window, which is positioned behind the active window

BPerlakiH:
Thank you @kutulus for reporting it. The fix for this will come in our next release.

kutulus:
Hi,
thanks for the prompt reply - and the clarification.

Are there plans to give the user more control over the link target in the future?
Opening a new tab instead of a new window might be a good approach - if the content concerns the same Zim file?

To be honest, the behaviour is very annoying for my use case. I have +1k POI's in a map application from each of which I would like to link to a detail page (e.g. Wikivoyage) in a ZIM archive.

Issue #1033, opened 2024-11-22
Implement Apple Pay donations for macOS


BPerlakiH:
Re-enable macOS entitlements, that is currently taken out on CI/CD.
See: https://github.com/kiwix/kiwix-apple/pull/1010

Re-enable the UI button for macOS, make sure that the payment session is set up correctly.

BPerlakiH:
I am re-opening this issue, as the CI part still need some fixing.

@rgaudin I have tested locally the following scenarios:
- removed all former certificates
- used the same command line build command as we have on CI:
for macOS:
```
xcodebuild -workspace Kiwix.xcodeproj/project.xcworkspace/ -scheme Kiwix -destination generic/platform=macOS -configuration Release -onlyUsePackageVersionsFromResolvedFile -allowProvisioningUpdates -verbose -archivePath ~/Desktop/Kiwix-CI-mac.xcarchive build
```
for iOS:
```
xcodebuild -sdk iphoneos -workspace Kiwix.xcodeproj/project.xcworkspace/ -scheme Kiwix -destination generic/platform=iOS -configuration Release -onlyUsePackageVersionsFromResolvedFile -allowProvisioningUpdates -verbose -archivePath ~/Desktop/Kiwix-CI-iOS.xcarchive build
```

- I tested that without any certificates both of them are failing (it's expected)
- I created a **macOS Development certificate**, and the macOS build is failing with that:
```
Provisioning profile "Mac Team Provisioning Profile: self.Kiwix" doesn't include signing certificate "Mac Developer ... (9294M4J3B9)". (in target 'Kiwix' from project 'Kiwix')
```
iOS obviously also fails with:
```
No Accounts: Add a new account in Accounts settings. (in target 'Kiwix' from project 'Kiwix')
/Users/balazs/dev/ios/kiwix/Kiwix.xcodeproj: error: No signing certificate "iOS Development" found: No "iOS Development" signing certificate matching team ID "L7HWM3SP3L" with a private key was found. (in target 'Kiwix' from project 'Kiwix')
```
- I created an **iOS Development certificate**, and installed only that, in this case both builds are failing. iOS with:
```
error: No Accounts: Add a new account in Accounts settings. (in target 'Kiwix' from project 'Kiwix')
[...]

PR #1048, opened 2024-12-07
Fix deeplink url opening on macOS

BPerlakiH:
Fixes: #1044


BPerlakiH:
CI will probably won't build macOS until: https://github.com/kiwix/kiwix-apple/issues/1033 is not fully fixed.

Issue #752, opened 2024-05-06
Temporarily remove the "Support Kiwix" link from the apps


BPerlakiH:
Apple's suggested solution for non-profits can be found here:
https://developer.apple.com/apple-pay/nonprofits/

According to App Store rules, it is not allowed to use external links (this way) that leads to any kind of purchase (even donations).
For this reason in release 3.3.0 the links to "Support Kiwix"should be removed.

kelson42:
OK, a ticket ahould be created for the alternative solution

BPerlakiH:
@kelson42 ticket created: https://github.com/kiwix/kiwix-apple/issues/753

rgaudin:
Note written but I believe this is due to our publishing it into the Mac App Store.
Do we want to keep it for the DMG version? How easy is it to toggle this based on a build flag?

BPerlakiH:
I think it is possible to toggle this based on a build flag. The more important question is if we really want to.
Once Apple Pay will be in implemented, that is a better user experience, than the external link IMHO (users are more likely to pay if it's easier, leading to more donations, I suppose).

rgaudin:
I don't know enough about this but:

- Would this Apple Pay donation mechanism work with a DMG distributed outside the App Store?
- Would it allow direct donations without requiring an Apple account?

BPerlakiH:
- I am not sure if it works on DMG, never tried that to be honest. There's also Apple Pay for web, so theoretically it should work. We can also test it before release, there's even a sandbox testing environment.
- [Apple Pay can only be used with an Apple ID a compatible device and a card (participating)](https://support.apple.com/en-us/HT204506#:~:text=To%20use%20Apple%20Pay%2C%20you,in%20to%20your%20Apple%20device.). It has some limits on the supported countries / participating banks as well: https://support.apple.com/en-in/102775.
- Alternatively we can include both the link and Apple Pay in the DMG version...

kelson42:
@rgaudin's proposal makes sense, this is what we do for Kiwix Android.

PR #1049, opened 2024-12-07
Remove support button from custom apps

BPerlakiH:
Removing the support button from custom apps, to make sure they do not appear there.

### DIFF
--- a/App/App_macOS.swift
+++ b/App/App_macOS.swift
@@ -195,8 +195,10 @@ struct RootView: View {
             }
             .frame(minWidth: 160)
             .safeAreaInset(edge: .bottom) {
-                SupportKiwixButton {
-                    openWindow(id: "donation")
+                if Payment.paymentButtonType() != nil {
+                    SupportKiwixButton {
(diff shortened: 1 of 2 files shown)


BPerlakiH:
iOS already has the same condition:
https://github.com/kiwix/kiwix-apple/blob/8708589e84c67eb1a74aac2a30e3da173e244f09/Views/Settings/Settings.swift#L317-L321

kelson42:
@BPerlakiH I don't understand, I though this was already done in https://github.com/kiwix/kiwix-apple/issues/752

BPerlakiH:
@kelson42 To sum up: We were counting on the fact that if the merchant id (for Apple Pay) is not defined for the custom apps build target, the Apple SDK's function: ``PKPaymentAuthorizationController.canMakePayments()`` will return false.
After double checking it, that is not true, even if the payment capability is missing, it still returns true, as it's only checking the device itself if it's capable of Apple Pay. 
Therefore we need to do an additional check in the same place, if it's custom apps or not.

Issue #975, opened 2024-09-12
Improve launch sequence


BPerlakiH:
## The problem:
Currently there's no splash screen on iOS, and the initial loading state is not shown to the user. This results in a completely white screen on app start, for relatively long time:

https://github.com/user-attachments/assets/f00712a6-b31a-411f-a718-a9ad0b2c022e

BPerlakiH:
@kelson42 I double checked it, there's no way to add an animated gif to the splash screen. That would animate properly.

kelson42:
@BPerlakiH Then I guess we will have to keep current approach. Le see how looks next screen cast you will upload.

PR #1050, opened 2024-12-07
Fix macOS opening bookmark URLs for custom apps

BPerlakiH:
Fix opening the bookmark URLs for custom apps on macOS

### DIFF
--- a/App/App_macOS.swift
+++ b/App/App_macOS.swift
@@ -208,11 +208,6 @@ struct RootView: View {
             case .tab(let tabID):
                 let browser = BrowserViewModel.getCached(tabID: tabID)
                 BrowserTab().environmentObject(browser)
-                    .withHostingWindow { [weak browser] _ in
-                        if FeatureFlags.hasLibrary == false {
-                            browser?.loadMainArticle()
-                        }
-                    }
             case .bookmarks:
                 Bookmarks()
             case .opened:


kelson42:
@BPerlakiH Why we need this? Do the bookmark system has been broken somehow?

BPerlakiH:
@kelson42 Yes, after the changes around [the launch sequence](https://github.com/kiwix/kiwix-apple/issues/975) and asynchronous loading the library, we already loading the url for the browser instance even before it is literally displayed within the window.
This custom app specific code meant to make sure we always load the main screen. Since we already do that earlier, it is "overwriting" the bookmark url we want to load (as it happens later in time), and the result was that it displayed only the Home Screen instead.
The issue is macOS only specific, I've tested the iOS version of DWDS with the latest library, and that works as expected. I also tested Kiwix App and the bookmarks work OK there, both on main branch and on 3.6.0 from the App Store.
Since we haven't released any custom apps since those changes, "no harm" was done publicly.

kelson42:
@BPerlakiH I still don't understand why we load (display) bookmark URLs at start of a custom app. At the start of a custom app, we should work like for Kiwix: display the tabs which were open a the time the app has been closed.

BPerlakiH:
It is the other way around. 
This piece of code is a left-over, and should have been removed as part of the earlier refactoring.
With this code in place, if you click on a bookmark url, the home page opens, and not the bookmark url. Without this code it's working properly.

PR #1046, opened 2024-12-06
Remove hack around apple-pay capability for macOS

BPerlakiH:
Fixes: #1033

PR #1054, opened 2024-12-15
Fix CD process for macOS

BPerlakiH:
Trying to fix the CD process for macOS, applying the same solution as in: 
https://github.com/kiwix/kiwix-apple/pull/1046/files


kelson42:
@BPerlakiH Not sure which problem it foxes, but let me know when this is OK your side

BPerlakiH:
> @BPerlakiH Not sure which problem it foxes, but let me know when this is OK your side

@kelson42 The release CD builds are failing for macOS, so we probably need the same solution we did for the CI builds.

BPerlakiH:
I think it is good to be merged now.

Issue #1067, opened 2025-01-03
Create test-bed for macOS FTP builds


BPerlakiH:
In order to test the GitHubActions for macOS FTP build (which is part of the release process), we need a test environment, that can be run easily (without doing a full release).

kelson42:
I don't think we need that, at least I don't understand. What is the exact purpose? Testing the packaging of a dmg?

kelson42:
@BPerlakiH @rgaudin Iguess I can just close this as WONTFIX!?

PR #1068, opened 2025-01-03
Add CD test

BPerlakiH:
Fixes: #1067 
Testing CD steps


kelson42:
@rgaudin @BPerlakiH I'm not sure I want this is `main`. This shoudl be done adhoc in a PR when needed and then removed... or do I miss the point?

rgaudin:
There are a couple of issues with this PR:

- the goal is a bit fuzzy. What do we want to test exactly? Compilation? Notarization? Upload? 
- It's a duplicate of the current CD workflow. If this is meant to be ran adhoc on a test branch, that's fine (and we dont need a PR) but if this is going to be merged that's a duplicate of a long and complicated file.
- it's expecting a manual trigger to run (workflow_dispatch) which is great for tests but needs to be on `main` branch. This is tied to previous point.

I believe what we want is the CD workflow to be able to run outside the `release` event on a different event (specific tag or workflow_dispatch) and in this case have conditions on some step (like the upload one).

BPerlakiH:
@kelson42 @rgaudin At the moment, this is only temporary. Without merging it I don't have the option to run the CD build manually. I only see these options:
Screenshot 2025-01-05 at 12 38 04

and under CD, there's no manual run.
Since the original CD file is conditionally based on the trigger (either schedule, TestFlight or release), simply adding the workflow_dispatch (manual run) is not enough,
therefore I thought, that duplicating the file for test purposes makes more sense. Once the macOS FTP issue is solved, I am happy to delete this file.

BPerlakiH:
Or maybe there's a way to run this without merging it, that would be even better, but I couldn't find it.

kelson42:
> Or maybe there's a way to run this without merging it, that would be even better, but I couldn't find it.

This is what @rgaudin said a few days ago: all worflowsmcan be run in PR, you just need to put the right event there to be triggered.

BPerlakiH:
@kelson42 @rgaudin Maybe I am getting this wrong, and there's a way to do this... I am not sure.

By the first look, it seems to me that it's a security implementation from GitHub, so that unverified code (not merged) cannot modify the current workflows.

If you know a way how to run this without merging, I would be happy to learn that.

kelson42:
I don't understand why fixing a CI/CD which got broken a few weeks ago generates all that fuzz. why you don't just fix it in a PR?

rgaudin:
Just repeating my previous comment, workflow_dispatch are tied to `main` (but can select which branch to run on when triggered). You can use the push event on any branch even outside PR.

kelson42:
@BPerlakiH https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#push

It's important we fix this quickly, no reason to wait ages to fix CD and release.

BPerlakiH:
No longer needed. Closing it.

Issue #1071, opened 2025-01-06
Fix "flaky" unit tests


BPerlakiH:
After double checking our build issues, I realised that some unit-test results, are heavily dependent on User Defaults, which is a key/value storage by Apple, which is persisted across sessions.
By running the same Unit-tests several times (repeatedly without building) some of them might fail.

## Proposed solution:
- Wrap our user-default calls for those failing tests
- Use a test instance instead, that won't persist the results
- Also make sure that the created test db entries are cleaned-up properly

kelson42:
This does impact only the devs IMHO, portponing.

PR #1072, opened 2025-01-06
Fix "flaky" unit-tests

BPerlakiH:
Fixes: #1071

After double checking our build issues, I realised that some unit-test results, are heavily dependent on User Defaults, which is a key/value storage by Apple, which is persisted across sessions.
By running the same Unit-tests several times (repeatedly without building) some of them might fail.

## Proposed solution:
- Wrap our user-default calls for those failing tests
- Use a test instance instead, that won't persist the results
- Also make sure that the created test db entries are cleaned-up properly

### DIFF
--- a/Model/CategoriesToLanguage.swift
+++ b/Model/CategoriesToLanguage.swift
@@ -13,17 +13,23 @@
 // You should have received a copy of the GNU General Public License
 // along with Kiwix; If not, see https://www.gnu.org/licenses/.
 
-//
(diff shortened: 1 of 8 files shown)


kelson42:
CI is red!

rgaudin (dismissed):
Makes sense, indeed

rgaudin:
CI to be fixed before merging, **abviously**

Issue #1055, opened 2024-12-15
MacOS FTP release signing fails


BPerlakiH:
With the Apple Pay macOS builds, we have the following notarisation error:

```
** EXPORT FAILED **
Error Domain=DeveloperAPIServiceErrorDomain Code=5 "Cloud signing permission error" UserInfo={IDEDistributionIssueSeverity=3, NSLocalizedDescription=Cloud signing permission error, NSLocalizedRecoverySuggestion=You haven't been given access to cloud-managed distribution certificates.
```

https://github.com/kiwix/kiwix-apple/actions/runs/12340833380/job/34439982068

I am not yet sure which part we need to change to make this step work.

PR #1074, opened 2025-01-10
Remove support/donation for macOS FTP builds, as it's not supported outside of AppStore

BPerlakiH:
Fixes: #1055

Fix the CD process for macOS.
In short: the in-app-purchase is not supported by Developer ID signed apps (outside of the App Store), therefore we cannot use that capability in the FTP macOS build.
Although Apple Pay itself is supported, the In App Purchase capability that we use (for Donation - a purchase of a non-digital good) is not supported.

The found Apple reference: 
https://developer.apple.com/help/account/reference/supported-capabilities-macos

We need a follow up discussion, if we want to hide the support button in this case or maybe use a direct link instead (since this build is not going to be verified by Apple).

----- 

After the comments, the approach is that we inject the `HIDE_DONATION: true` key/value into the Support/Info.plist file for macOS FTP builds. For all other builds the fallback value is `false` (do not hide). The condition whether to display the Support / Donation button is also dependent on the type of brand we run, therefore custom apps are not affected.

### DIFF
--- a/.github/actions/xcbuild/action.yml
+++ b/.github/actions/xcbuild/action.yml
@@ -25,9 +25,6 @@ inputs:
   KEYCHAIN_PASSWORD:
     required: false
     default: mysecretpassword
-  KEYCHAIN_PROFILE:
-    required: false
-    default: build-profile
   XC_WORKSPACE:
     required: false
     default: Kiwix.xcodeproj/project.xcworkspace/
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -9,6 +9,12 @@ name: CD
 # |-----------------|---------|------------------------|----------------|---------------------|
 # | UPLOAD_FOLDER   | nightly | -                      | -              | release/kiwix-macos |
 # |-----------------|---------|------------------------|----------------|---------------------|
+
+# Signing certificates:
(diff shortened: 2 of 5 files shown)


kelson42:
> We need a follow up discussion, if we want to hide the support button in this case or maybe use a direct link instead (since this build is not going to be verified by Apple).

@BPerlakiH Please open an issue for that asap, otherwise we will forget

kelson42:
> > We need a follow up discussion, if we want to hide the support button in this case or maybe use a direct link instead (since this build is not going to be verified by Apple).
> 
> @BPerlakiH Please open an issue for that asap, otherwise we will forget

Actually, if the feature is not working, it has to be hidden. You can do that here in this PR

rgaudin (approved):
LGTM but please rename the PR to something more meaningful.

BPerlakiH:
@kelson42 @rgaudin I've updated both the name of the PR, and the code. With this the support/donation button is conditionally hidden for macOS FTP builds (for Kiwix only).

kelson42:
CI red

BPerlakiH:
Please merge this other PR first: https://github.com/kiwix/kiwix-apple/pull/1076

Issue #1081, opened 2025-01-17
macOS unit tests are still timing out on CI


BPerlakiH:
GitHub action is still misbehaving with our unit tests running on macOS (times out as never started).
Sometimes it passes for the first time, sometimes it times out, and when re-tried it passes.
It's hard to tell even, when it's fixed, as the former attempt of fixing (to not to use code signing) passed the first time, but then it failed on another PR.

Latest example PR for this:
https://github.com/kiwix/kiwix-apple/actions/runs/12833615870

It failed 3 times, while passed the 4th time without any new commits!

Other GH action users were also complaining about similar issues:
https://github.com/actions/runner-images/issues/6098

Since we are running the very same unit tests on iPhone simulator, which gives us the code coverage, I am proposing to turn them off for macOS, as it's a duplicate run of the exact same tests. The macOS run of them is currently giving us certainly more pain then benefits.

Note: We are still building the app for macOS as part of this process, we just would not run the unit-tests again.

BPerlakiH:
Also it seems to me that GH Action is using some virtual env. magic, which is really slow.
To compare a clean build on M1 of the macOS Kiwix app is under 30 sec, on GH Action it takes over 3 minutes!

kelson42:
@BPerlakiH Upstream issue you refert to is closed, so should work. We can not remove the whole CI for a whole platform. Compiling and running on/for macOS is not exactly the same as for iOS. The root cause has to be better understood.

PR #1082, opened 2025-01-17
Remove in app payment for unit tests on macOS for CI

BPerlakiH:
Fixes: #1081


BPerlakiH:
@kelson42 I've updated this PR, and built it 6 times with success: 
https://github.com/kiwix/kiwix-apple/actions/runs/12958311333

Thank you for the hint @rgaudin: indeed when I moved the removal before the build step (as suggested) it is working as expected.

Issue #1086, opened 2025-01-21
Missing string in 3.7.1 donation button


Popolechien:
A user reported:

> The iOS app is missing a string for the newly introduced donate button. My device is in German.
> version 3.7.1 Build 204. That’s where the issue appears when donating a custom amount.

See image:

![Image](https://github.com/user-attachments/assets/427d62c7-85cc-4ff2-9833-e595282999b8)

kelson42:
@BPerlakiH This is urgent and requires s patch release ASAP. We also need to understsnd how this happened exactly.

BPerlakiH:
@kelson42 I've created a PR for this.
It seems, I forgot to add the `.localised` part to the usage of this string.
I double checked the other-ones for donation, they are OK in this regard.

kelson42:
@BPerlakiH Do you now a way to lint this? We should have detected such an error in the CI.

BPerlakiH:
As for why it happened:
If the string is found by the pure key in the current locale (eg. English), the value for that will be picked up automatically.
If the translation is missing, it will fall back to the key value itself.
What we do, is use this custom extension ``.localized``, that will do the fallback for us to English, if there is no value for the current local (as we don't yet have one for German).

BPerlakiH:
Created a new task for this: https://github.com/kiwix/kiwix-apple/issues/1088
I have a couple of ideas for this, I need to do some experiments to see which is more feasible.

PR #1087, opened 2025-01-21
Fix localisation, remove debugPrint statement

BPerlakiH:
Fixes: #1086


rgaudin (approved):

kelson42:
@BPerlakiH Actually we can not really make a patch release, so we will have to release 3.8.0 with only a few fixes.

BPerlakiH:
@kelson42, no problem, I have reverted the project file to be 3.8.0

Issue #1088, opened 2025-01-21
Make translations safer


BPerlakiH:
We want to make sure no missing translation issue slip through, such as: #1086.
Therefore we need a safe and secure way to ensure that the translations are properly used.

PR #1089, opened 2025-01-22
Safe translations applied

BPerlakiH:
Fixes: #1088 

## How the translations work:
We have a key / value association for each language. English is the default and the reference one.
If a key is missing in another language, we want to fall-back to English. This fallback is a custom solution, Apple doesn't do this for us.

## How it worked so far:
We were using the keys in the code, and applied the ``"payment.confirm.button.title".localized`` extension on it (our custom fallback function, or in some rare cases the ``"".localizedWithFormat(withArgs: ..)`` function, that can handle placeholders.

## How could it go wrong:
 - if we do a typo in the key itself, it's not going to work obviously. It will show up badly even in English, but we need to notice it (go to that specific section of the app).
 - if we forgot to add ``.localized``, it will work in English, but [won't fall back to English for languages, that do not have this translation yet.](https://github.com/kiwix/kiwix-apple/issues/1086)

## What else can be (and now is) improved:
- we have the qqq "special language", that contains the explanations for translators. Only that we do not really know which keys are really missing from there
- we might have translation keys we do not use at all
- we might use keys that were valid some time ago but no longer contained in the translation files

# Solution:

## Generate
- in order to avoid mistyped keys, we can rely on the compiler itself. To do so, we generate a special swift file, that contains all the vars we can use, based on the English translation files. This gives us code-completion and compile time check.
- we make sure that every key contains (at least one) dot '.'
- we convert these into "_" for the swift variables eg: "payment.confirm.button.title" becomes: 
``static let payment_confirm_button_title = "payment.confirm.button.title".localized``
- we make the ``.localized`` and the ``"".localizedWithFormat(withArgs: ..)`` function file private. This way it can not be used directly. This avoids the problem of eg. re-applying it twice on the same value, or using it on some other keys, the translation file doesn't know of.
- we make this special swift file generated (at the same time we generate the project)
- we add the generated file to .gitignore

## Validate
- we check if each key has a dot (".")
- with the above, we can search all the swift files (apart from the generated one), and make sure that those keys are not used anywhere (directly), since the key itself contains the ".", while the generated swift var is not, a text search is enough for this purpose
- we can print out all the missing "qqq" explanation entries (it's not an error, only info)
- we can print out all the keys that are not used anywhere in the project (it's not an error, only info)
- we add this validation step to the CI

## Notes on additional changes
Therefore in light of the above, I have:
- removed the unused translations
- changed a few translation keys to contain "."
- filled in all the qqq explanations, that were missing
- found one more invalid key in the code, that I've replaced with a valid one
- since the CI is linting (only) the changed files, I had to add some minor lint related fixes as well


kelson42:
@BPerlakiH Here we need to generate an error in CI instead of just warnings if:
* unused strings
* missing translation string
* undocumented string

BPerlakiH:
@kelson42 , no problem, please see it updated.

kelson42:
@BPerlakiH Woukd like to merge it and will do it.... once CI passes :)

BPerlakiH:
@kelson42 now it seems to be good. There's still "flaky-ness" on the CI part :(

Issue #1097, opened 2025-02-03
Bottom toolbar disappearing on iPad Slide Over mode


Popolechien:
From a [user](https://kiwix.freshdesk.com/a/tickets/84680):

> In iPad slide over mode, the bottom toolbar to navigate screen and look up headings in page disappears and can only be accessed by pressing the search bar at the top, then canceling again, which makes the bottom bar appear.

iPad mini 6th gen
Model A2567
iOS version 18.1.1

BPerlakiH:
I can confirm, it is 100% reproducible.

PR #1098, opened 2025-02-03
Fix iPad split mode scroll to appear bottom bar

BPerlakiH:
Fixes: #1097 

Make sure we apply the `showBars` function depending on the screen size type, and not on the device, this solves the issue on iPad split screen.


kelson42 (changes requested):
red CI

PR #1103, opened 2025-02-07
Change url type to be copied for macOS, remove id from menu

### LINKED ISSUE #1099: Copying Zim File Cell URL can't paste in terminal
When I tried to copy the URL from an item in categories and then paste that into the terminal it doesn't work. Nothing happens. This will also happen when trying to copy it to a text editor. It does however work when copied into a browser's URL tab and even the Notes app.

I found that this is because the data type that is set for the string is `.URL`. While this might technically be correct it seems to limit the number of places one can paste the value, presumably because some applications can't handle the `.URL` type.

https://github.com/kiwix/kiwix-apple/blob/5f8e4dfdbbe8ec0b586dcde51569b37426352504/Views/Library/Library.swift#L172

I tried changing that line to treat the download URL string as a regular string and everything works fine:

```swift
NSPasteboard.general.setString(downloadURL.absoluteString, forType: .string)
```

I can paste the URL string anywhere I like with this change.

Is this something the maintainers would be open to changing? I would be happy to submit a PR for this if it would be welcomed and helpful.

BPerlakiH:
Fixes: #1099 

macOS (right clicking on a catalog item)
Screenshot 2025-02-07 at 11 45 39

iOS (long pressing on a catalog item)
Screenshot 2025-02-07 at 11 47 03

### DIFF
--- a/Support/br.lproj/Localizable.strings
+++ b/Support/br.lproj/Localizable.strings
@@ -55,7 +55,6 @@
 "library.zim_file_context.main_page.label" = "Degemer";
 "library.zim_file_context.random.label" = "Pajenn dre zegouezh";
 "library.zim_file_context.copy_url" = "Eilañ an URL";
-"library.zim_file_context.copy_id" = "Eilañ an ID";
 "zim_file_opened.overlay.no-opened.message" = "Restr ZIM digor ebet";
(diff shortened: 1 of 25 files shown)


rgaudin (approved):
LGTM

Issue #1104, opened 2025-02-07
Update Defaults dependency


BPerlakiH:
The Package dependency we have "Defaults", should be updated to avoid a macOS bug, where the side navigation menu might get "stuck" and response too slow.

BPerlakiH:
It seems we can only use this if we update the CI as well to a newer macOS image and newer XCode...

BPerlakiH:
@kelson42 
After updating the dependency the macOS side menu does not seem to improve, and can still get slightly stuck, so there are no user facing benefits at the moment. 

Possibly can be postponed to 3.9.0, due to the cascade of things to be updated, which seems a bit risky right before release.

kelson42:
@BPerlakiH Wich macos die menu are we exactly talking about. Can you please open an issue?

PR #1111, opened 2025-02-13
Update defaults dependency to 8.2

BPerlakiH:
Fixes: #1104 

Tested on iPad, iPhone and macOS.

Updating not to the latest version, but the highest we can still run on Xcode 15.4


kelson42 (changes requested):
CI does not pass

BPerlakiH:
This one should build OK, once updated with the revert from: https://github.com/kiwix/kiwix-apple/pull/1123

Issue #1116, opened 2025-02-14
iOS Library "New" section scroll is not performing well


BPerlakiH:
When scrolling through the "New" section the scroll is not smooth enough.

Number 1 suspect is the way we load the thumbnail images for the tiles, which might not be "async".

BPerlakiH:
With the last 2 PRs the performance of the scroll has been greatly improved.
One last thing is left, is to make sure the "first time" usage of this section is even smoother. There's a micro-hang, when we first scroll the categories (on a fresh install) on iOS, especially on iPad (where a larger amount of cells can fit on the screen). The reason is that our attempt to prefetch the icons does not work (see related issue: https://github.com/kiwix/libkiwix/issues/1180)
and we try to do both fetching and saving the data into our local store (CoreData / SQLite) at the time of the display in a not so asynchronous manner.

It is to be decided, what should be done with this remaining issue:
1. fix the original source of the problem in libkiwix, and update libkiwix dependency once done, and re-test if it still occurs.
3. implement and rely on a prefetching solution on the client (swift Kiwix app) side, executed at the time of category fetching
4. postpone fixing it, as it's not that serious of a "ui hang"

kelson42:
@BPerlakiH We are ready now to fix this?

BPerlakiH:
> [@BPerlakiH](https://github.com/BPerlakiH) We are ready now to fix this?

Yes, we are good to go with fixing this.

kelson42:
@BPerlakiH Then please carry on

BPerlakiH:
@kelson42 I managed to do some improvements on this.
Additionally there's more we could do, I am just not sure which repo to report it on, but here it is:

Certain favicons are the very same (by look), although their URLs are generated by / tied to the id of the file. 

Eg: all TED items have the very same favicon:

<img width="537" height="2048" alt="Image" src="https://github.com/user-attachments/assets/9efb1a68-51cf-45d9-87fd-7245df8a1bea" />

But all of them are under a different URL, hence we load all of them, like:

https://library.kiwix.org/catalog/v2/illustration/c9f6953a-f776-b105-c603-7c667630ec98/?size=48
https://library.kiwix.org/catalog/v2/illustration/fb24b763-a807-64ff-ba71-7af2f7602095/?size=48

I suppose it's a question how we represent them in the catalog XML. 
We could gain some further optimization, if they were pointing to the same URL.

PR #1117, opened 2025-02-16
Fix library scroll issue - part 1

BPerlakiH:
Mostly fixing the issue: #1116

There were 2 main issues:
 - the [earlier search fix](https://github.com/kiwix/kiwix-apple/pull/1094), wasn't optimal in terms of scroll performance. 
 - it turned out that SwiftUI gets lost, when we use `ViewModifier` within a `ForEach` loop. It was hard to track down, but the possible explanation for this, is that SwiftUI needs to see the view hierarchy tree in order to make performant diffs between states. Now it seems that the former `ViewModifier` was erasing too heavily the type of the underlying views, and the performance got worse. Solution is to use view composition instead of view modifier.

The good news is that it performs a lot better now. 

There needs to be a part 2 fix for this as well. There's a "micro hang" on a freshly installed app, when scrolling the new section. This is mainly due to the icon loading and saving to the DB at the same time, for which I do have some solutions, but they are not final yet, so will leave them to a new PR.

Managed to add back the animations as well, and it works great:

https://github.com/user-attachments/assets/df8e10b0-5555-42b6-a37d-e6c8d2b52357

### DIFF
--- a/Views/Library/Library.swift
+++ b/Views/Library/Library.swift
@@ -141,40 +141,10 @@ struct LibraryZimFileContext: ViewModifier {
             #endif
         }.contextMenu {
             if zimFile.fileURLBookmark != nil, !zimFile.isMissing {
-                Section { articleActions }
+                Section { ArticleActions(zimFileID: zimFile.fileID) }
             }
-            Section { supplementaryActions }
-        }
-    }
-
-    @ViewBuilder
-    var articleActions: some View {
-        AsyncButton {
-            guard let url = await ZimFileService.shared.getMainPageURL(zimFileID: zimFile.fileID) else { return }
-            NotificationCenter.openURL(url, inNewTab: true)
-        } label: {
-            Label(LocalString.library_zim_file_context_main_page_label, systemImage: "house")
-        }
-        AsyncButton {
-            guard let url = await ZimFileService.shared.getRandomPageURL(zimFileID: zimFile.fileID) else { return }
-            NotificationCenter.openURL(url, inNewTab: true)
-        } label: {
-            Label(LocalString.library_zim_file_context_random_label, systemImage: "die.face.5")
-        }
-    }
-
-    @ViewBuilder
-    var supplementaryActions: some View {
-        if let downloadURL = zimFile.downloadURL {
-            Button {
-                #if os(macOS)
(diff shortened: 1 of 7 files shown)


Views/Library/Library.swift
```diff
@@ -112,70 +112,47 @@ struct LibraryZimFileDetailSidePanel: ViewModifier {
...
             if zimFile.fileURLBookmark != nil, !zimFile.isMissing {
-                Section { articleActions }
+                Section { ArticleActions(zimFileID: zimFile.fileID) }
             }
-            Section { supplementaryActions }
-        }
-    }
-
-    @ViewBuilder
-    var articleActions: some View {
```
BPerlakiH, line 151: this became: ArticleActions

Views/Library/Library.swift
```diff
@@ -112,70 +112,47 @@ struct LibraryZimFileDetailSidePanel: ViewModifier {
...
-        AsyncButton {
-            guard let url = await ZimFileService.shared.getRandomPageURL(zimFileID: zimFile.fileID) else { return }
-            NotificationCenter.openURL(url, inNewTab: true)
-        } label: {
-            Label(LocalString.library_zim_file_context_random_label, systemImage: "die.face.5")
-        }
-    }
-
-    @ViewBuilder
-    var supplementaryActions: some View {
```
BPerlakiH, line 167: this became: CopyPasteMenu

Views/Library/ZimFilesNew.swift
```diff
@@ -14,26 +14,89 @@
...
-    @EnvironmentObject var viewModel: LibraryViewModel
+    @EnvironmentObject var library: LibraryViewModel
     @Default(.libraryLanguageCodes) private var languageCodes
-    @FetchRequest(
-        sortDescriptors: [
-            NSSortDescriptor(keyPath: \ZimFile.created, ascending: false),
-            NSSortDescriptor(keyPath: \ZimFile.name, ascending: true),
-            NSSortDescriptor(keyPath: \ZimFile.size, ascending: false)
-        ],
-        animation: .easeInOut
```
BPerlakiH, line 31: this is the animation we lost, since the fetching is done in a different way, we might need to add it back in a different form, but I will leave that for later.

PR #1122, opened 2025-02-22
Fix iOS build and use iOS 18.1 simulators on Xcode 16.2

BPerlakiH:
The former PR updating to Xcode 16.2, made the [build fail](https://github.com/kiwix/kiwix-apple/actions/runs/13473510711).
Reason being is that the default iOS SDK for simulators in Xcode 16.2 is the "latest", which is currently iOS 18.2. The issue is that Github runners (macOS-14) do not have that pre-installed yet, so we need to fall back to simulators running iOS 18.1, which is installed:
https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md

well it's still not ok, maybe it's because non of the simulators are pre-installed as per Apple: 
https://developer.apple.com/documentation/xcode-release-notes/xcode-16-release-notes

I am not convinced at this point to do this update, if we need to download the simulator environment for every single build we run on GitHub Actions ...

### DIFF
--- a/.github/actions/xcbuild/action.yml
+++ b/.github/actions/xcbuild/action.yml
@@ -116,6 +116,10 @@ runs:
       shell: bash
       run: xcrun xcodebuild -workspace $XC_WORKSPACE -scheme $XC_SCHEME -showBuildSettings
 
+    - name: Show eligible build destinations for the ${{ inputs.XC_SCHEME }}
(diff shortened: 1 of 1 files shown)


BPerlakiH:
This is not completely ready for Xcode 16, closing it for now.

Issue #1107, opened 2025-02-08
Improve LiveActivities


BPerlakiH:
- Change background fill for live activities to mach other widgets
- Make sure long running download processes still updates the progress of the live activity
- Use "guess-timated" values for progress if needed, to make the updates more smooth

kelson42:
Still not fully fixed

kelson42:
@BPerlakiH I'm still mixing a new approach to speed up liveactivities refresh rate

kelson42:
We still have massive problems.

First one is that the progressbar are not in sync at-all, even after waiting 10s. See

<img width="1080" alt="Image" src="https://github.com/user-attachments/assets/02d56ecd-69d5-4b33-8964-99d5a6802a03" />

The second one is that when I clicked on the live activity, it opens Kiwix, but not the book which is currently downloading.

BPerlakiH:
Well, the progress bar is either going to be in sync with the amount, but won't move for 10sec, or moving constantly but not in sync with the amount..
.
I was thinking about it, and maybe it would be better to just replace it with a spinning progress indicator, so that it's visible that something is happening, that way we do not need to be precise exactly where it is (at least not for the next 10 sec), and could leave the amount text updated every 10 sec.

kelson42:
> Well, the progress bar is either going to be in sync with the amount, but won't move for 10sec, or moving constantly but not in sync with the amount...

It is not in sync even after 10s waiting. It's not a sync problem, it is a computation problem.

kelson42:
> > Well, the progress bar is either going to be in sync with the amount, but won't move for 10sec, or moving constantly but not in sync with the amount...
> 
> It is not in sync even after 10s waiting. It's not a sync problem, it is a computation problem.

Honestly I really don't care much how this is done, but it ahould be obvious that such a screenshot is not acceptable.

BPerlakiH:
Ok, I've removed it from 3.8.0:
https://github.com/kiwix/kiwix-apple/pull/1130

kelson42:
Still not working fine, no update every few (not even 10) seconds

<img width="1080" alt="Image" src="https://github.com/user-attachments/assets/80094c0d-a332-4d86-b775-327cc87b4eb6" />

kelson42:
One bug is that the live activity seems not to be updated when kiwix is in the background:
* starting download
* clicking on home (backgrounding the app)
* swiping down to display live activties

I get:
[...]

PR #1126, opened 2025-02-28
Use progress bar with timer updates for live activities

BPerlakiH:
Fixes: #1107

-----

Screenshot 2025-02-28 at 11 32 46

## Solution

I replaced the progress indicator, with a linear bar that is updating constantly, so for our users it is more smooth. We do update the "real" progress every 10 seconds, which changes the amount of downloaded values displayed above the bar, and that updates the timer below the bar, and makes the bar either speed up or slow down.

I've also added a timestamp for each update, this helps the system decide which updates can be dropped if they do queue up, so once the system "unblocks" the live activity it can apply the latest update. 

I think, this is the best we can do considering the constraints that the iOS (especially above 18.0) puts on us.

## The Dynamic Island updated

Screenshot 2025-02-28 at 11 31 56

## The compact view remained unchanged:

Screenshot 2025-02-28 at 09 31 54

## Localisation

Screenshot 2025-02-28 at 11 40 02

## Paused state

Screenshot 2025-02-28 at 15 44 06

## Downloads file updates

The original `DownloadService` file became too long, so I split that up by classes and moved under a new folder.

### DIFF
--- a/Common/DownloadActivityAttributes.swift
+++ b/Common/DownloadActivityAttributes.swift
@@ -86,7 +86,14 @@ public struct DownloadActivityAttributes: ActivityAttributes {
             progressFor(items: [self]).localizedAdditionalDescription
         }
         
-        public init(uuid: UUID, description: String, downloaded: Int64, total: Int64, timeRemaining: TimeInterval, isPaused: Bool) {
+        public init(
+            uuid: UUID,
+            description: String,
+            downloaded: Int64,
+            total: Int64,
+            timeRemaining: TimeInterval,
+            isPaused: Bool
+        ) {
             self.uuid = uuid
             self.description = description
             self.downloaded = downloaded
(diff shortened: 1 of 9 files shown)


kelson42:
The black bird part of the logo is too small, it should be like in on the app icon in the dashboard.

kelson42:
* Pause/resume should be handled somehow
* localisation (RTL) should be checked

BPerlakiH:
@kelson42 the points above are addressed now, it should be good to be merged.

kelson42:
@BPerlakiH CI fails :(

BPerlakiH:
Ok, I will have a look, it seems that the SDKRoot to build is too low on CI...

BPerlakiH:
@kelson42 we need to merge this one first: https://github.com/kiwix/kiwix-apple/pull/1127

kelson42:
@BPerlakiH Still red

BPerlakiH:
@kelson42 after merging https://github.com/kiwix/kiwix-apple/pull/1128, this should be good as well. I did run a CI job on the combination of this branch and the Xcode branch here:
https://github.com/kiwix/kiwix-apple/actions/runs/13607824062
and that is green.

Issue #1115, opened 2025-02-14
macOS / iPad side menus are "sticky" sometimes


BPerlakiH:
When switching between macOS sideMenus, they feel slow / "sticky", there's a noticeable UI delay.
Additionally there's a SwiftUI error in the logs:

```
Publishing changes from within view updates is not allowed, this will cause undefined behavior.
```

<img width="163" alt="Image" src="https://github.com/user-attachments/assets/8e12bb2d-0d09-4519-946f-7d8eb4588cbf" />

kelson42:
@BPerlakiH What is the status here after #1125 has been merged?

BPerlakiH:
@kelson42 The error in itself is solvable in a specific way, but it stops using a single source of truth for what is really selected, and introduces a "middle-man" variable at the view level. I think this is risky to do before our 3.8.0 release.

PR #1135, opened 2025-03-08
1104 side menu optimalization

BPerlakiH:
Fixes: #1115

Solves the error message:
```
Publishing changes from within view updates is not allowed, this will cause undefined behavior.
```

Also improves the side menu performance. After careful measuring in Instruments, it turned out the checking for Apple Pay status is quite expensive to be done synchronously on the main thread. This check was causing the side menu to be slow.


kelson42:
@BPerlakiH Any kind of slow stuff impacting UI should be async per default. Let me know if one of them should be handled differently.

Issue #1101, opened 2025-02-06
Rethink accessibility of the donation button


kelson42:
Few remarks on the topic let me think we could do better:
* Apple reviewer was not able to find it
* It's not really proeminent
* Maybe we should think about a popup

BPerlakiH:
I was thinking about this a bit:

Apart from an additional pop-up, we should probably keep the current place (or some other place) for donations, so if the user changes his/her mind can still donate

In order not to be too "pushy" towards users with such pop-up interruptions, we can use the local storage (User Defaults) to store any/all of the following events:
    - time of last launch of the app
    - time of last successful download
    - time of last donation
    - time of last dismissal of the donation pop-up

We can obviously use other events as well, it's just the first things that came into my mind.
Based on such list, we can decide when to display and when not to display the pop-up.
Any suggestions are welcomed, how you envision these rules.

The only drawback of this approach is that User Defaults are deleted together with the app. Meaning in case of an uninstall and re-install of the app (upgrades won't count), we are "looking at" a completely new user again (instead of a returning one). I suppose it's not that bad, taking user privacy into account, since we won't send their data anywhere.

rgaudin:
Couple of things to consider:

- Requesting donation is annoying to users
- Users who might be willing to donate should be informed that it's possible and actually much welcomed.

Those statements are contradictory so we need a subtle approach that allows information without being annoying.
Here's a proposal:

- no popup. Visual popups are now associated with dismiss-able information and very little attention is paid to them. I propose to replace the content area with a new view instead.
- first-launch message. It's important new user knows about this. I believe we already have a first-launch view where we invite to browse or refresh the catalog. We can add our message there: “Kiwix is Free Software made by a non-profit. Please consider [ donating ] to xxx”
- successful donation should be recorded in the profile
- reminder message. After `x` number of launchs of the App (without existing donation), present a message inviting to donate. From there, an option to “never be reminded again” should be present. It's important we tie reminder with actual use of the app so a non-using or low-using user is not bugged with donation requests every time he launches the app. This requires increasing a number for every launch.
- reminder on updates. Similar to regular reminders, also requiring a number of previous launches but triggered after an app update.
- Keeping the in-settings menu action of course

rgaudin:
Ah @BPerlakiH I didn't see your comment ; glad we suggested similar things. As per user data being removed, I don't see this as an issue. Are we iCloud syncing this by the way?

BPerlakiH:
[...]

PR #1139, opened 2025-03-09
Re-arrange donation button on iPad

BPerlakiH:
As part of #1101 
we want the donation button to be more prominent.
As a first step we can solve the issue on iPad, and conditionally put the button to the end of the side menu:

https://github.com/user-attachments/assets/7f6b3d50-404e-47dc-b433-c66a5dbb4d34

In full screen iPad we show the support button as part of the side bar, and won't show it under settings.
In split view we do show it under the settings (as before):

Screenshot 2025-03-10 at 00 34 42

On iPhone it remained the same, we can tackle that as a next step.

## Other changes:
Until now the navigation state and the menu items we displayed were the same possible values. Since this new Support button, is a different kind, I had to separate these 2 concepts. On the plus side, we could remove some translation entries, that we never really displayed, since eg: the application could be in a loading state, but we do not have such menu item, that you could select, therefore we don't need a menu item title (and icon) for that.

### DIFF
--- a/App/App_iOS.swift
+++ b/App/App_iOS.swift
@@ -27,18 +27,6 @@ struct Kiwix: App {
     @UIApplicationDelegateAdaptor(AppDelegate.self) private var appDelegate
     private let fileMonitor: DirectoryMonitor
     private let activityService: ActivityService?
-
-    // MARK - donation
-    enum DonationPopupState {
-        case selection
-        case selectedAmount(SelectedAmount)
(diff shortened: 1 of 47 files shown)


kelson42:
@BPerlakiH Thx, LGTM, but you please put just below Sttings (aligned to the left)

kelson42 (approved):
LGTM

PR #1146, opened 2025-03-28
1107 LiveActivities instant updates fixes

BPerlakiH:
Fixes: #1107 

Make sure we handle background tasks.
Use more frequent updates, without timers, that the background task can cancel.
Do a direct update of live activities in scheduled background mode.
Fix the remaining time calculations.


kelson42:
@BPerlakiH Please make a new testflight once merged

PR #1147, opened 2025-03-30
1107 Live Activities

BPerlakiH:
After many attempts, here's what I've found, what is possible and what is not possible with live activities:

As per [Apple's doc](https://developer.apple.com/documentation/foundation/url_loading_system/downloading_files_in_the_background#2928518):

> In iOS, your app could be in the foreground, suspended, or even terminated by the system.
> If your app is in the background, the system may suspend your app while the download is performed in another process. In this case, when the download finishes, the system resumes the app and calls the ... completion handler

What it really means is that the background download is running in a completely different daemon, and it won't give us any feedback of the progress itself, if the app is suspended (or terminated), only the fact that it has finished.

Additionally to that, currently Apple is not providing any way to tell, when the app goes into suspended mode...

What we do have though is:
- around 30 seconds (sometimes more) of processing, even if the app is backgrounded, during this time we can continue to update our live activity with the accurate progress
- we can use an approximation to figure out how much time could possibly left of to download, and move the progress bar, and the count down accordingly. This gives the user a feeling of continuity.

From the above the best I could come up with is:
- do not display the amount downloaded (eg: 20MB of 2Gb), since we do not know when updates will stop coming, and we would be stuck on the last value, while the progress bar (and or the countdown would keep going).
- we do display the approximated time remaining. Obviously it can go up or down, if we get a new update in the meantime
- we can not use a large amount of samples for the approximation. If we pause / resume a download or re-try a failed download, the approximation needs to "fix itself". Eg. using 30 seconds of samples, might get filled up with 0 (or close) bytes of download samples, and any approximation later, when we do start to progress again, will be based on those zeros, and will be wrong for another 30 seconds. For this reason a shorter period is better. (We do get about 2-3 samples per seconds on average.)
- I managed to figure out to actually stop the live-activities, if the app is terminated, which was a previous problem
- managed to recreate and fix the issue with the live activity not leading back the user to the downloads section

NOTE:
Former attempts of using the **BackgroundTask** API, were not successful. That is designed more to create one of updates or maintenance work on apps (eg once a day / once a week), and there's no guarantee when they are going to be launched by the system. The approach is more user friendly (taking into consideration power, performance and privacy) than developer friendly, and the system decides when it is going to be the best time to launch such a task.
For this reason it's not reliable to count on it, that it will solve our LiveActivity updates problem.

Alternatives:
Using [Apple Push Notifications](https://developer.apple.com/notifications/). This is very hairy to be honest for our use case.
It's based on [Push API](https://www.w3.org/TR/push-api/), and we would need to essentially set up a download session for each user, with a specific token, where the server streaming the download would be aware of what the download status is, and based on that could "push" the progress to the live activity, and that could update accordingly.


kelson42:
We should rely on the approximation only within a few seconds between updates. Any else than this is going to be doomed.

kelson42:
I will try to find a FOSS having a download with notification

Issue #1016, opened 2024-10-31
Fix BrowserViewModel memory leak


BPerlakiH:
Even if the tab is closed/deleted and the related BrowserViewModel is cleared from the cache, it stays around in the memory (a leak). One of the reasons, I have found so far, is that it's easy to put a class into Swift UI's `@Environment`, but there's no easy way to take it out from there (so far I haven't found a solution for that).

PR #1148, opened 2025-04-03
Fix memory leaks by cleaning up BrowserViewModel references

BPerlakiH:
Fixes: #1016

Many things had to change in order to clean up the BrowserViewModel references, this (rather large object) is holding a reference to the entire WKWebView for each tab, so it's important to clean it up properly once the tab is closed, otherwise we are using more and more memory.

Tested on iPhone, iPad and macOS.


kelson42:
@BPerlakiH Inguess zou have remarked that the CI does not pass? Please ask for review only if the CI pass and keep the PR in draft as long as it does not.

rgaudin (approved):
Thank you! It looks good to me but it's mostly beyond my understanding…
Anyway it's such a big change that we need to keep this in mind for a while as there might be unforseen consequences in other places

kelson42:
I will try to find a FOSS having a download with notification

Issue #1152, opened 2025-04-05
Add feature flag for immersive reading (for PhET)


BPerlakiH:
In our upcoming PhET custom app, immersive reading (hiding the top and bottom bars) is colliding with the content. 
It is very hard for the user to navigate away from a full screen loaded experience, only swiping from the side of the screen helps, whereas the remaining menu buttons are not accessible:

Instead of the current state:

![Image](https://github.com/user-attachments/assets/c4264e7d-c4a0-43c2-a11d-83fb664da25e)

We want (via custom configuration) to have:

![Image](https://github.com/user-attachments/assets/2cb5fe61-772b-404e-8ee8-4f53f9db8e0b)

NOTE: the above examples are from iPad, but the same problem applies for iPhones as well.

kelson42:
@BPerlakiH what domyou mean with "immersive reading"? You mean autohiding of the topbar?

PR #1153, opened 2025-04-05
Add feature flag to turn off immersive reading (custom apps)

BPerlakiH:
Fixes: #1152 

Adding an optional way to turn off immersive reading.


kelson42:
We need a related PR for custom apps. Not sure about the feature name, no clear enough (at least to me).

Issue #1090, opened 2025-01-23
MacOS navigating with keyboard focus


BPerlakiH:
As requested here: https://github.com/kiwix/kiwix-apple/pull/1079
It would be desirable to be able to navigate between the search results and the search bar using the keyboard and the macOS focus.
Ideally it could be achieved with native focus solution only, and no custom workarounds.

kelson42:
@BPerlakiH Can you please make a concrete proposal?

PR #1154, opened 2025-04-05
macOS shortcuts to search

BPerlakiH:
Partial implementation of: #1090 

## Search shortcuts:
- Now it's possible to activate the search menu with command + F. It applies for all the relevant screens.
- Added escape key to dismiss web content search
- Note: it will fully work only from macOS 15

## Additional
- Additional small fix to update the side menu selection, when the navigation is updated programatically (eg redirecting to categories, when user has no ZIM files)
- Moved the `HostingWindowFinder` to a new file, because `App_macOS.swift` just got over the lint limit in terms of number of lines.

## Preview

https://github.com/user-attachments/assets/d143a502-cf33-414d-bb33-22601359f6f9

FYA: @jarrodmoldrich

### DIFF
--- a/App/App_macOS.swift
+++ b/App/App_macOS.swift
@@ -83,15 +83,6 @@ struct Kiwix: App {
                     isSearchFocused = true
                 }
                 .keyboardShortcut("f", modifiers: [.command])
-                // intentional duplicate! When on reading tab with loaded webview content
-                // command + F will search the content itself
-                // therefore we want command + shift + F to trigger top bar ZIM file search
-                // The priority which one is picked is based on view tree hierarchy, see:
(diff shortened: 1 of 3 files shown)


rgaudin:
What about find-in-page if command-F is now mapped to Search?

rgaudin:
OK, I'm not sure how I feel about changing shortcuts depending on where you are in the app but the fact that the most-expected command-F triggers find-in-page and that search is anyway always available from the task bar makes it usable.
It's just that search shortcut is replaced with find-in-page when in a ZIM and that other shift-command-F is more a hidden shortcut for those who know 👍

rgaudin (approved):
LGTM

Issue #1155, opened 2025-04-09
Add "Do not ask again" tick button when unlinking files


Popolechien:
If I have a large batch of files to remove from my directory I am being asked every time to confirm that I want to unlink. There should be an option "Do not ask again" that I can tick (and untick in the preferences) that makes it so that future deletion will not require confirmation.
![Image](https://github.com/user-attachments/assets/8be73c3e-a8da-4d60-9f7b-b2619fbf09b3)

kelson42:
I'm not sure this is the right approach. I would we consider a way to delete many books in obe shot.

Popolechien:
Yes as well but I figured this one would be a lower-hanging fruit

PR #1157, opened 2025-04-12
1155 multi select zimfiles + selection fixes

BPerlakiH:
Fixes: #1155 

## Multi selection:

Introducing CMD+click to select multiple (local / opened) ZIM files, and now it's possible to unlink those in one go:

https://github.com/user-attachments/assets/855385b3-1852-4f6d-acd0-de24cb3e8f21

This type of selection is native to macOS, eg: you can do the same in Finder.

## Additional selection issues fixed:

Additionally fixed multiple selection issues on macOS:
- the selected items are actually now highlighted, so it's easier to see what was selected, eg in categories, or downloads:
Screenshot 2025-04-12 at 22 23 53
- I changed the code around a bit as well, as the former solution was a global selection value, which caused some strange behaviour in multi window mode (selection and the details displayed was synced between windows), which was wrong:

## BEFORE:

Screenshot 2025-04-12 at 22 27 33

as you can see, the selected cell on the left window is not highlighted, so it's hard to tell the details of which cell we are displaying. By selecting anything in the left window was causing the right window details to be updated. In this case the right window is showcasing a list of items in the TED category, but the selected item is actually coming from the left window, which is in a completely different category. In short it was not possible to compare the details of 2 different ZIM files in 2 parallel windows.

## AFTER:

Screenshot 2025-04-12 at 22 26 05

The selected ZIM file details are now per scope of window (not global), and the selected cell item is actually highlighted.

## Tested
on macOS 15.4, and also tested on iOS (iPhone + iPad) to make sure nothing got broken there.

### DIFF
--- a/App/App_macOS.swift
+++ b/App/App_macOS.swift
@@ -182,12 +182,8 @@ struct RootView: View {
     @State private var currentNavItem: MenuItem?
     @StateObject private var windowTracker = WindowTracker()
     @State private var paymentButtonLabel: PayWithApplePayButtonLabel?
-<<<<<<< HEAD
     var isSearchFocused: FocusState<Bool>.Binding
     @StateObject private var selection = SelectedZimFileViewModel()
-=======
-
->>>>>>> 1f496b8e (Make selection tab scoped)
     private let primaryItems: [MenuItem] = [.bookmarks]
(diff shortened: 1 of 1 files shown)


rgaudin (approved):
LGTM

rgaudin (approved):
Makes sense, I didn't think of multi window to be honest

PR #1159, opened 2025-04-13
macOS search navigation v2

BPerlakiH:
Fixes: #1090

Adding the capability to move between the search bar and the search results with up/down arrow on macOS.

The ideal would be to do so in other search results as well, eg. on the news / categories or bookmarks section, however those are displaying the results in a grid, and so far finding the displayed items (left/right/up/down) is non trivial in SwiftUI, so will keep this change narrowed down to the ZIM file search results, which is a 1 column list.

The solution is based partially on the ideas presented here: https://github.com/kiwix/kiwix-apple/pull/1079/files

https://github.com/user-attachments/assets/8ba85bf1-d280-4044-b236-a07ed3de5bce

FYA: @jarrodmoldrich

### DIFF
--- a/Views/SearchResults.swift
+++ b/Views/SearchResults.swift
@@ -134,7 +134,9 @@ struct SearchResults: View {
                         }
                         if nextIndex < viewModel.results.startIndex {
                             $focusedSearchItem.wrappedValue = nil
+                            #if os(macOS)
                             NotificationCenter.default.post(name: .zimSearch, object: nil)
+                            #endif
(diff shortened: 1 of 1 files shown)


kelson42:
@BPerlakiH Thank you for this new attempt. Iwill let @rgaudin review this

kelson42:
@rgaudin ?

jarrodmoldrich:
@BPerlakiH @kelson42  Trying this now.  Works great!  Thanks for making it happen.

Issue #1170, opened 2025-04-19
Article title appears twice in the TOC


kelson42:
<img width="1080" alt="Image" src="https://github.com/user-attachments/assets/f59033a0-8223-4fa9-92f6-44f7259eca2a" />

In this example "Santé publique" is writtentwice. It should be only once.

Remarks: I believe this is because all h* nodes are displayed in the TOC and the first one is the "title" of the whole article. Actually here nothing is really wrong either in the DOM or in Kiwix... but the result is still not convincing.

I propose to hide the first h* if the following two conditions are fullfilked:
* h* entry is the very same string as the article/entry title AND
* there is no other entry in the DOM at this h* level (for example only one h1 or h2)

kelson42:
@BPerlakiH BTW this screenshot shows an other bug: "santé publique" and "histoire" are at the same horizontal level in the TOC althouh they are not in the DOM. Something is wring somewhere.

kelson42:
@BPerlakiH I would prefer that we agree about what to do before having a PR to review.

So, you have removed the whole title of TOC. I didn't asked this. But you are lucky as I'm not even sure we need it.

For the rest please implement what I asked.

Regarding the restthere is and should not be anything collapsable.

kelson42:
The first item has to disappear if given conditions are fullfilked. Why you don't do that?

PR #1171, opened 2025-04-19
1170 fix outlines and collapsibility

BPerlakiH:
Fixes: #1170 

There were 2 issues: 
- the former optimisation broke the Outlines, since the single instance web configuration was setting up the javascript listener, from where we get the outline items via a callback.
- I removed the title from the outline drop down, since it is not clickable, it just makes unnecessary noise (and might do duplication as well)
- If the top level item is a single one, we are removing it, instead of the former solution of flattening the hierarchy:

https://github.com/user-attachments/assets/7c576e22-752b-4e56-b799-945524b5bc09

### DIFF
--- a/Model/Entities/Entities.swift
+++ b/Model/Entities/Entities.swift
@@ -91,23 +91,15 @@ final class OutlineItem: ObservableObject, Identifiable {
     let index: Int
     let text: String
     let level: Int
-    let isCollapsible: Bool
     private(set) var children: [OutlineItem]?
 
     @Published var isExpanded = true
-    
-    func asNonCollapsible() -> OutlineItem {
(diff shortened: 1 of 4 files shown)


kelson42:
I don't think this what I asked for.

kelson42:
@BPerlakiH The video in the description has not been updated... or it still does not make what I have asked for.

kelson42:
AFAIK, it does not verify if this is the same as the article title

kelson42:
@BPerlakiH Better but you don't have deHTMLize the H* content.

kelson42:
@BPerlakiH https://www.w3docs.com/tools/html-encoder/

Issue #1184, opened 2025-04-25
Strange to see "Support Kiwix" in the Phets app


kelson42:
It should be something like "Support Phets app" or something around these lines. We should check how it is on Android Phets app.

PR #1187, opened 2025-04-27
Fix the donation/support button on iPad

BPerlakiH:
Relates to #1184

After investigating this, the current state is, that we should not have support button for custom apps:
https://github.com/kiwix/kiwix-apple/blob/39b1f9d56ef0c2c3bcd6e99b332170f5de509e31/Model/Payment.swift#L126

We do not have the support button in current PhET on macOS and on iPhone, so as of today it is showing up by pure incident on iPad.

It turned out that after the former perf. optimisation + moving the support button to be part of the side menu item, we were no longer calling the function that determines the payment button type, which also checks for things like: is apple pay supported on this device, is it a custom app, and so forth.

So this had to be fixed for the both Kiwix itself, and custom apps as well.

We can treat adding the support button, and "re-branding" it for custom apps as a separate task, if that's what we want.

### DIFF
--- a/App/SidebarViewController.swift
+++ b/App/SidebarViewController.swift
@@ -149,7 +149,7 @@ final class SidebarViewController: UICollectionViewController, NSFetchedResultsC
         // show the donation async
         Task {
             if snapshot.sectionIdentifiers.contains(.donation),
-               let _ = await Payment.paymentButtonTypeAsync() {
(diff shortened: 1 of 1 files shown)


kelson42:
Support item has been implemented for custom apss / Dwds, see https://github.com/kiwix/kiwix-apple-custom/pull/62.

@BPerlakiH I don't understand why you write this feature does exist!

Issue #1179, opened 2025-04-23
Opening Deeplinks MacOs Splitscreen Mode


kutulus:
Hardware: Macbook Pro 2017
Example link: zim://DE071ED9-7067-6C01-0EB9-3EC704AB5940/A/Germany

No external deep links are opened in splitscreen mode.
If Kiwix is in splitscreen mode, a new tab is created for an external link, but the content is not loaded.
Only the ‘Load Spinner’ is displayed.

kutulus:
Kiwix: Version 3.8.0 (213)
Macbook Pro 2017
MacOS 13.7.5

![Image](https://github.com/user-attachments/assets/a3d04898-b0de-408d-b98a-b9bba95617d5)

PR #1188, opened 2025-04-27
1179 macos window split screen loading fix

BPerlakiH:
Fixes: #1179 

It fixes both opening a deep-link in split screen mode, and also opening a ZIM file in split screen mode.


kelson42:
@BPerlakiH Can we have an qutomwted test for these two scenarios?

kelson42:
@BPerlakiH We will need UI automated tests. The sooner, the better. Please open issue and think about a framework/approach which would best make the job.

Issue #1197, opened 2025-05-06
Get the link of the current page


kutulus:
hi,
is there a way to see/copy the URL of the currently open page of an archive?
It is possible to copy the link to a page (right click, copy on the link) but to find out the link of the current page?

kelson42:
@kutulus Do you have a concrete example with ZIM, article and URL you woukd like to get?

kelson42:
@BPerlakiH I understand now but why should we implement such a feature? What is the use case?

kutulus:
Hi,
Right, in my case I have POI's on a map that have links in ZIM files. 
To create new POI's with links it would be helpful to be able to copy the URL directly from a given Zim page.

However, I see a universal benefit in saving deep links outside of Kiwix. E.g. in note apps like Obsidian.md - basically possible as with any web browser

rgaudin:
As a user I'd expect this to be available as _Copy URL_ under the Share menu. Would that work?

kelson42:
@kutulus At the core I really want to support this kind of integration with third party software. But:
* I prefer to fully understand the use case before talking about a solution
* We have a serious problem with all these links, they are related to a specific ZIM. Once the ZIM updated, all the links will be broken.

rgaudin:
@kutulus my understanding is that this is to be settled in #15 and your ticket is _just_ about providing the link to users. Let's not mix those two different issues

kutulus:
is just a suggestion...

kelson42:
@BPerlakiH Can we have a "Cooy PDF" and "Copy URL" on this widget?

<img width="1080" alt="Image" src="https://github.com/user-attachments/assets/8998a44d-d588-43be-a9aa-40fd99b9dfa9" />

kelson42:
@BPerlakiH Does that mean you can do it like dpecified at https://github.com/kiwix/kiwix-apple/issues/1197#issuecomment-2865798137 ?

PR #1199, opened 2025-05-10
1197 link to current page

BPerlakiH:
Fixes: #1197 

Now it's possible to both copy the current URL either via the drop down menu.
Or by pressing CMD + SHIFT + C (when there's a URL loaded that is):

## SCREENSHOTS:

Screenshot 2025-05-10 at 23 10 07

------------------------------------------------------------
## 2 options to choose from
------------------------------------------------------------

Screenshot 2025-05-10 at 23 10 19

## Notes:
if we select Copy URL it will get copied to the paste board and the drop down closes, no further pop-ups.

------------------------------------------------------------
# after selecting "Share as PDF"
------------------------------------------------------------

Screenshot 2025-05-10 at 23 10 26

## Additional changes:
The `App_macOS.swift` file grew too big, so I moved the `RootView` component to its own file.


kelson42:
@BPerlakiHFew remarks:
* First option should "Share PDF"
* In "Share PDf" option, title of the Widget is (in the screenshot):"Main_Page". Either it is the title of title of the page and it should be "Main Page", or it is the filename and it should be "Main_Page.pdf"
* Any reason why "Copy URL" makes the copy straight and does not open a widget like "Share PDF"?

kelson42 (approved):
LGTM

Issue #1205, opened 2025-05-19
Phets content does not load properly at app start


kelson42:
If I launch phets app, it tries to reload open tabs and fails at the loading stage

<img width="1080" alt="Image" src="https://github.com/user-attachments/assets/b2d87c20-68f8-49f1-9dba-a50ed29251ad" />

kelson42:
@BPerlakiH This is the only reader where I see this? Why does itbwork fine on Android for example?

benoit74:
This works fine on other readers, and even works fine on MacOS Sequoia 15.5 with official Phet app:

![Image](https://github.com/user-attachments/assets/a744d9d2-669f-4b92-bfe0-772f9b848083)

![Image](https://github.com/user-attachments/assets/c3a3d789-c507-4323-8883-add13b0b9b2e)

BPerlakiH:
Need the same fix for iPhones as well!

PR #1210, opened 2025-05-24
Fix PhET initial webpage to be on screen for html canvas calculations

BPerlakiH:
Fixes: #1205 

## The problem
My investigation lead me to the following conclusions. This problem occurs only on app start, and not always.
It is a timing issue. What we do is displaying the initial "loading data..." state, and in the meantime we start loading the web content, once the content is loaded (in full), we display the web view. This works fine for most content but not for PhET.
The reason being, is that PhET subpages (we want to restore), are using javascript to draw on ad-hoc canvases, but that fails if the web view is off-screen (hasn't been added yet to the view hierarchy on iOS).

## Solution:
Add the web-view to the view hierarchy up-front, but make it invisible, and once we switch states from "Loading data...", swap around the opacity. This way the canvas drawing can be done properly.

### DIFF
--- a/Views/BrowserTab.swift
+++ b/Views/BrowserTab.swift
@@ -184,27 +184,30 @@ struct BrowserTab: View {
                             #endif
                     } else {
                         switch model.state {
-                        case .loadingData:
-                            LoadingDataView()
-                        case .webPage(let isLoading):
-                            WebView(browser: browser)
-                                .ignoresSafeArea()
-                                .overlay {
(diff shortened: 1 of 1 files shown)


benoit74:
Probably important to mention this issue is not specific to PhET but any content needing access to the canvas at page load. Only content which we know about is PhET, but I'm sure others exists or will exist, typically some zimit ZIMs.

PR #1217, opened 2025-05-26
1214 ci is failing since last week

### LINKED ISSUE #1214: CI is failing since last week
See https://github.com/kiwix/kiwix-apple/actions/workflows/ci.yml?query=branch%3Amain

BPerlakiH:
Fixes #1214 

Overall the Kiwix app starts with fetching the catalog on iOS, which after success opens the catalog page. Since the UI tests are heavily time based, there's no guarantee when the fetch will end (network dependent).
For this reason, we can pass a special launch argument to the application, when running the UI tests, and based on that turn off the category fetching functionality.

Note: There were some files to clean up, that were no longer needed.


kelson42:
> Overall the Kiwix app starts with fetching the catalog on iOS, which after success opens the catalog page. Since the UI tests are heavily time based, there's no guarantee when the fetch will end (network dependent).

This kind of timing issues are a standart issue. the UI testing framework must have a way to deal with it without having to modify the app. The UI should trigger the test to continue when element are loaded. The whole operation has anyway a timeout implemented in the catalog fetching.

PR #1220, opened 2025-06-01
1212 stabilize CI tests, add tab selection UI test (local)

BPerlakiH:
Fixes: #1212 

Attempting to stabilize CI unit / UI tests.

After several attempts on GitHub actions, and also on BrowserStack couldn't make the UI tests deterministic enough...

**Proposing to have the UI tests running locally**, where they are stable.
Also added a couple of fixes, that are making the Unit tests more stable on CI.
**I've run them 7 times in a row, and all of them are green:**
https://github.com/kiwix/kiwix-apple/actions/runs/15666712426

Fixes:
- delete the kiwix app before launching the unit tests on the booted simulator
- make sure that the unit tests and the UI tests are running on different simulators
- use valid bundle ids for our test targets
- test targets are limited to the appropriate devices (iPad, iPhone, mac)
- the unit tests around ODPS parser, making sure the DB is clean up before running the test, and that the parsing happens on the same ZimActor (no concurrency conflicts).

UI testing the selection actually works in a way that we navigate to the Catalog, and pick a small ZIM file, that we download, load the main page, restart the app, and then check if the tab is selected.
Double checked the test results, by reverting [the fix made earlier](https://github.com/kiwix/kiwix-apple/pull/1211), and the test fails in that case as expected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,12 +68,16 @@ jobs:
 
       - name: Run Unit Tests on macOS
         if: matrix.platform == 'macOS'
-        run: xcodebuild test -scheme Kiwix -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO
+        run: xcodebuild clean test -scheme Kiwix -configuration Debug -destination 'platform=macOS'
 
       - name: Upload code coverage
         uses: codecov/codecov-action@v4.2.0
         env:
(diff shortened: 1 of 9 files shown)


kelson42:
CI still does not pass... what should we do?

BPerlakiH:
@kelson42 it does pass after these changes, please see it here:
https://github.com/kiwix/kiwix-apple/actions/runs/15666712426

PR #1221, opened 2025-06-07
Fix Phets initial webView sizing

### LINKED ISSUE #1205: Phets content does not load properly at app start

BPerlakiH:
Fixes #1205 for iPhones.

It turns out it was working on Simulator, but not on real device.
This is the same fix [as for iPads here](https://github.com/kiwix/kiwix-apple/pull/1210/files).

Also fixing the iPhone UI tests here.


kelson42:
CI does not pass!

BPerlakiH:
@kelson42 now it's good to be merged.

PR #1227, opened 2025-06-22
Move support button in iOS settings above the fold

BPerlakiH:
Related to: #1101 

Proposing to move the "Misc" section higher in the Settings on iOS, this way the Support Kiwix button is at least above "the fold":

## BEFORE
IMG_3240 Medium

## AFTER
IMG_3241 Medium
Screenshot 2025-06-22 at 16 29 09 Medium

It might not fully solve the problem, but should convert better.


kelson42 (approved):
LGTM, at least worth a try

Issue #1244, opened 2025-07-15
Spacing and title of Hotspot button


kelson42:
The badge with the number of books served should not stick to "Stop server" <img width="744" height="60" alt="Image" src="https://github.com/user-attachments/assets/f2b1037f-987b-4a48-9062-911dd6ffeda7" />

kelson42:
@rgaudin @BPerlakiH I wonder if it would not be better to write "Stop Hotapot" in place of "Stop server"

PR #1245, opened 2025-07-16
Fix hotspot badge position

BPerlakiH:
Fixes: #1244 

Simulator Screenshot - iPad Pro 13-inch (M4)


kelson42:
@BPerlakiH There is no way to do that in an elegant way, by speficying the margin between both items? What will happen there if you have 1000 ZIM files selected?

BPerlakiH:
Your are right @kelson42, please see it updated:

Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-07-17 at 20 49 47
Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-07-17 at 20 49 14

Issue #1251, opened 2025-07-20
UI design of Hotspot panel is not satisying


kelson42:
<img width="738" height="498" alt="Image" src="https://github.com/user-attachments/assets/dd2ac8c9-25b6-4239-bd2b-4c57bd7a213b" />

kelson42:
@BPerlakiH Can younplease propose a mockup with a better UI?

BPerlakiH:
Ok. Currently the "Share" button is copying the URL value itself only and not the QR code image. 
I am not sure about the use case of sharing a QR code to be honest, as the QR code in itself is a "sharing mechanism", but I can add it.

kelson42:
> Ok. Currently the "Share" button is copying the URL value itself only and not the QR code image. 
> I am not sure about the use case of sharing a QR code to be honest, as the QR code in itself is a "sharing mechanism", but I can add it.

Yes, it is just a simplification for the user (one click in place of two). I believe the good approach is to provide both icons aside (without any text). Please don't forget the visual hint when clicking on the copy icon.

BPerlakiH:
iPad
----
<img width="2752" height="2064" alt="Image" src="https://github.com/user-attachments/assets/d848423a-1293-40d8-9260-35527e7c437e" />


iPhone
----
<img width="1320" height="2868" alt="Image" src="https://github.com/user-attachments/assets/5d478815-16e1-43d6-ab06-6d0c1cf5f945" />

iPhone landscape
----
<img width="2868" height="1320" alt="Image" src="https://github.com/user-attachments/assets/4f24c24a-aa34-4bac-ac35-32685a368ce8" />

mac
---
<img width="930" height="797" alt="Image" src="https://github.com/user-attachments/assets/9cbad958-17e8-49d1-8aac-8251d630dd22" />

kelson42:
@BPerlakiH Copy/share are missing for qrcode

BPerlakiH:
@kelson42 please see it updated above ^^

kelson42:
@BPerlakiH Thx, actually we should sort out the question of the WiFi access point before IMHO. See my comment https://github.com/kiwix/kiwix-apple/issues/170#issuecomment-3094089418. For the moment we have only the server (not the acces point part). See Kiwix Android, if you don't understand what I expect.

BPerlakiH:
**iPad**

[...]

PR #1253, opened 2025-07-22
1251 hotspot UI

BPerlakiH:
Fixes: #1251

- updated with the latest port fixes
- using a centered single column layout for Hotspot details
- dedicated explanation for macOS and iOS 

**iPhone**

Simulator Screenshot - iPhone 16 Pro Max - 2025-08-03 at 19 44 47
Simulator Screenshot - iPhone 16 Pro Max - 2025-08-03 at 19 44 36

-----

**iPad**

Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-08-03 at 19 46 14
Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-08-03 at 19 45 59

-----

**macOS**

Screenshot 2025-08-03 at 19 48 01
Screenshot 2025-08-03 at 19 47 51

-----


kelson42:
Let us have an agreement on the mockup before your start implementation

kelson42:
@BPerlakiH What is the status here regarding our latest comments on the mockup?

rgaudin:
- URL block has been moved up
- We want to remove the link on the URL (on iOS/iPadOS) string because it leads to an impossible situation.
- We want to move the second sentence of the last block next to the URL so it's more visible and because it makes more sense there.
- We need to check how the server behaves regarding the share button action. If this leads to it being unavailable, we could have a popup on click to warn about it

BPerlakiH:
@kelson42 posting here the latest UI as well:

**iPad:**

Image

------------
**iPhone:**

Image

------------
**macOS:**

Image

Issue #1256, opened 2025-07-25
Problems with hotspot port setting


kelson42:
How can you justify the port boundaries you have given? To me allowed ports are from 1 to 65535? 

When I type the port, it vanish behind the virtual keyboard. The settings panel should smootly vertically adjust to avoid current focuse element to be behind virtual keyboard.

<img width="2160" height="1620" alt="Image" src="https://github.com/user-attachments/assets/d565e638-1f67-461d-a6cb-831e7c22f105" />

No error message shoukd be displayed liked this. I would recomment to have shorter and fixed message similar in the form to "Change will only apply".

Make the input so that only allowed values can be put there

What happens if the chosen port is occupied at the time you start the server?

BPerlakiH:
Please give more clarification on what the error message should be.

> Make the input so that only allowed values can be put there

At the moment invalid values won't be saved, so this part is implemented ... in a way.
The alternative input mode is a "stepper", I tried that, but it's very painful to get through this wide range using one tap/click a time (eg: getting from 1000 to 2025 is 1025 clicks).

> What happens if the chosen port is occupied at the time you start the server?

So far I have seen 2 outcomes:
- the URL will take you somewhere else (eg, I had a local server running on the same port on macOS)
- the browser won't open the kiwix server either with or without an error

kelson42:
> Please give more clarification on what the error message should be.
> 
> > Make the input so that only allowed values can be put there

No error should be displayed in he port settings. You just put a clear sentence like: "Network port used by the Hotspot, should be between 1 and 65535".

> At the moment invalid values won't be saved, so this part is implemented ... in a way.

It should not be allowed to put a wrong port number in the i put... anytime. No check at save time.

In addition, it makes little sense to test if port is busy at settings time. It should be check only once: just before starting the server... Even better woukd be if the server would fail an return a proper error code (please verify if this is possible).

> The alternative input mode is a "stepper", I tried that, but it's very painful to get through this wide range using one tap/click a time (eg: getting from 1000 to 2025 is 1025 clicks).

I let you free to use the most appropriate widget.

> > What happens if the chosen port is occupied at the time you start the server?

The server should not start and an error shoukd be displayed "The port xyz is currently busy and therefore the Hotspot can not be launched. Free the port ot change the Hotspot port in the settings."

BPerlakiH:
I am looking into if it can be done without an error message on the Settings side..

BPerlakiH:
**macOS:**

[...]

PR #1257, opened 2025-07-26
Change valid port range, add port validation, fix error message

BPerlakiH:
Fixes: #1256

- Added description to Settings input field about the valid range of port numbers
- The input field is "self correcting" as the user inputs the value for the port number
- Added new error message when the port is already in use, and how to fix it

**macOS:**

Image
Image

----------
**iPhone**

Image
Image

-----------
**iPad**

Image
Image

### DIFF
--- a/Model/Hotspot/Hotspot.swift
+++ b/Model/Hotspot/Hotspot.swift
@@ -80,13 +80,6 @@ final class Hotspot {
         return URL(string: address)
     }
     
-    func runningPort() async -> Int? {
-        guard let port = await self.hotspot?.__port() else {
-            return nil
-        }
-        return port.intValue
-    }
-    
     static func fixedUp(port: Int) -> Int {
         guard minPort < port else {
             return minPort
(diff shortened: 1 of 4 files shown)


Model/Hotspot/Hotspot.swift
```diff
@@ -21,9 +21,9 @@ final class Hotspot {
     @MainActor
     static let shared = Hotspot()
     
-    private static let minPort = 1024
+    private static let minPort = 1
```
rgaudin, line 24: I'm skeptical about this. Usually (and even historically on macOS) users cant open ports under 1024 (the reason you started from there). I see that on my computer, I can open such ports as a regular user but I don't know if that's something I did. I doubt it's possible on iOS but I didn't check. Pretty sure older versions of macOS/iOS wouldn't support it without some privilege excalation.

kelson42: @rgaudin To articulate even a bit more about my request, I just want to give freedom - as far as reasonable - to the user and not arbitrary put a preset range of allowed ports. If possible, to me, `80` is a very valid port. This is why I have asked about the "rationals".

rgaudin: Indeed, I think a privileged-less default is fine and allowing any port (via escalation if required) would be great. If it happens that all the platforms we support allow binding to 80 without intervention, then defaulting to 80 would be even better

kelson42:
@BPerlakiH What is the status here? I don't see any PR description or screenscast or screenshot!? I also see you have again merge main in the branch, please fix this error.

BPerlakiH:
@kelson42 I've found a way to validate the port number directly to see if it's in use or not.

BPerlakiH:
@kelson42 PR updated. Please see the new description and screenshots above.

Issue #953, opened 2024-08-25
Multiple tabs with video content collide with Picture in Picture


BPerlakiH:
## Used content:
- [TED 3d Printing, 30 Jul 2024](https://download.kiwix.org/zim/ted/ted_mul_3d-printing_2024-07.zim.meta4)
- [TED humanities 2 Aug 2024](https://download.kiwix.org/zim/ted/ted_mul_humanities_2024-08.zim.meta4)

## Steps to reproduce:
- We need 2 tabs (A and B) both with video content, each of them needs to be at a non-zero progress position (eg: at the middle of the stream)
- From tab A start the video, launch in full screen, switch to picture in picture
- Select tab B
- Notice that tab B is "fixed up" (by play and pause), but that pauses the picture in picture (PIP) playback

## Recording:
https://github.com/user-attachments/assets/7a22631f-1073-4e1a-b7ee-e30f629fcbcf

Initial investigation:
It seems we have opposing forces:
- if a video content re-appears we want to "fix it" with play + pause, so it's not causing [the black screen issue](https://github.com/kiwix/kiwix-apple/issues/772)
- currently we only fixing it if the current tab's browser is not using Picture in Picture (and has some progress, meaning it's not the poster)
- the issue is that the picture in picture video can come from another tab, further more it can even come from a tab, that has been closed

## Possible solution:
Unfortunately from swift there seems to be no easy way to get to the picture in picture video player that was created by the WKWebView
With some not so elegant JS code, by listening on each video elements [entering picture and picture](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/enterpictureinpicture_event), and [leaving picture in picture events](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/leavepictureinpicture_event), but we need to pass these back from the individual webView instances to swift and gather them, so we can track if we have any picture in picture or not.
This way we could potentially fix the current issue, BUT:
- it would open up the possibility to have [the black screen issue](https://github.com/kiwix/kiwix-apple/issues/772) again. Under the condition of foregrounding the Kiwix app and landing on tab B, while tab A's picture in picture video was playing all the time
- it would create even more JS hacks, similar and based on [the ones we wished not to have](https://github.com/kiwix/kiwix-apple/issues/939).

## Question: 
I think it's a rare corner case, so worth to discuss if fixing it in a very complicate way is the way to go.

AlecsSantos:
Download 
[...]

PR #1262, opened 2025-08-02
Use video fixup for black screen only under iOS 18

BPerlakiH:
Fixes: #953 
remove the black screen fix from iOS 18 and above.

From my findings described here:
https://github.com/kiwix/kiwix-apple/issues/953#issuecomment-3146744619

**Summary:**
- we had a black screen issue on iOS 17, that we originally fixed
- this fix came with a side-effect, this was reported in the form of the ticket: #953
- after testing on iOS 18 without the original fix, there is no black screen issue and no side effect
- this PR is removing the original fix from iOS 18 (and above)
- for iOS 17 the original fix is still needed, as we decided that the black screen one is a bigger issue, than the side-effect (when using Picture in Picture).

### DIFF
--- a/ViewModel/BrowserViewModel.swift
+++ b/ViewModel/BrowserViewModel.swift
@@ -367,9 +367,11 @@ final class BrowserViewModel: NSObject, ObservableObject,
 
     @MainActor
     func refreshVideoState() {
-        Task { [weak webView] in
-            await MainActor.run { [weak webView] in
-                webView?.evaluateJavaScript("refreshVideoState();")
+        if #unavailable(iOS 18.0) {
+            Task { [weak webView] in
+                await MainActor.run { [weak webView] in
+                    webView?.evaluateJavaScript("refreshVideoState();")
+                }
             }
         }
     }


kelson42:
Please a description about your fix.

How does it behave with iOS26? Your test seems to ignore cases with iOS above 18? You test is only about iOS 18.0!!!

Why do we keep the fix at all? How the video.js latest version behaves with versions pre 18?

BPerlakiH:
@kelson42 I do not (yet) have iOS 26. The problem is that there's no way to downgrade an iOS device, once updated there's no turning back. iOS 26 is still in beta.
What we can do though is to create a TestFlight version with the solution from this PR, and you can compare it on your iOS 26.

kelson42:
> @kelson42 I do not (yet) have iOS 26. The problem is that there's no way to downgrade an iOS device, once updated there's no turning back. iOS 26 is still in beta. What we can do though is to create a TestFlight version with the solution from this PR, and you can compare it on your iOS 26.

Thank you for putting a comment. Please do that systematically. I don't compile the PR so I need a clear hih level explanation with screencast or screenshot to understand what you do.

Regarding iOS26, we have to be prepared before the official release of a new iOS/macOS. I really wonder why you don't have a simulator for this!? If not, could you consider having an other device to allow us to do that?

But, even without a test device on your side (I have one!), we need to consider what will happen... and even if you say this PR applied 18 and above, this code if `#unavailable(iOS 18.0)` does not do that IMHO (glad you explain to me how I'm wrong).

Then we have the hack introduced by #1262 for iOS17. The point is not if we want to fix #953 (obviously we want). I ask, because we use now a new version of Video.js! Can you please confirm the hack is still necessary, even with this new version of Video.js?

BPerlakiH:
After more testing, it seems this won't be good on iOS 18...
Closing this, and leave my summary in the original issue.

kelson42:
@BPerlakiH I had questions around testing of latest versions of iOS which are unanswered.

BPerlakiH:
I will look into what is the best option for me in terms of having a real device running on iOS 26.
Unfortunately the simulator is not supporting Picture in Picture, therefore this specific issue cannot be tested on a simulator alone.

Issue #1261, opened 2025-07-31
Unclear tab title / content if ZIM file is unreacheable


kelson42:
I have deleted the Angular docs ZIM file and this is what I get

<img width="2160" height="1620" alt="Image" src="https://github.com/user-attachments/assets/7791767c-bcfa-47d1-94c4-04fc80643216" />

BPerlakiH:
It might be an iOS 26 issue only.
By the code it should close all related tabs, when the ZIM file is deleted.
So far I could not reproduce it on iOS 18:

https://github.com/user-attachments/assets/842cde68-2b2d-44bb-bcf2-663f7105c558
https://github.com/user-attachments/assets/3896634b-f485-4a2c-9507-ebbc72b71051

kelson42:
You should delete the file in the file browser, not in kiwix itself.

BPerlakiH:
I managed to implement the improvements for iOS, but came across something really strange though.

If I delete the file via the "Files" app, it goes into "Recently deleted", if I immediately check it's there. If I remove it from there it is gone forever, as expected.

The problem occurs, if I just delete the file, but won't remove it from "Recently deleted", but go back to Kiwix.

Since we are using system bookmarks, we are able to track if the file was moved, or renamed etc, and still being able to load it.
This results in a really spooky situation. The Kiwix app is seeing the ZIM file being able to load it as usual, but the ZIM file cannot be found via the Files app!
Debugging the URL location, it is pointing to the Application Trash though, eg:

```
Found ZIM (502ABB22-8D5D-0F3A-BE90-C41DED088930) 
URL: file:///private/var/mobile/Containers/Data/Application/0B83AD93-A969-410A-AA81-ADB6A06663E5/Documents/.Trash/wikipedia_en_ray-charles_mini_2025-08.zim
```

Whereas it's cannot be found externally...
It might be some iOS bug, it happened before:
https://discussions.apple.com/thread/252372601?sortBy=rank

The file itself can be properly deleted from within Kiwix though, which solves the problem, nevertheless it's very strange...

PR #1301, opened 2025-09-14
1261 unreachable zim

BPerlakiH:
Fixes: #1261 

We want to make sure that we are re validating the system bookmarks (via these we track ZIM files, even if they are moved / renamed), and flag missing ones. (function renamed from reopen -> revalidate, as that is what it is really doing)

The next step is to make sure we delete the tabs that are associated with missing ZIM files.

Also, we need to make sure, we are listing tabs that have non-missing ZIM files associated.

Plus, added some extra safety, to make sure we are not crashing, when trying to read metadata from a non existing file.

### DIFF
--- a/SwiftUI/Model/LibraryOperations.swift
+++ b/SwiftUI/Model/LibraryOperations.swift
@@ -96,28 +96,6 @@ struct LibraryOperations {
             "Reopened \(successCount, privacy: .public) out of \(zimFiles.count, privacy: .public) zim files"
         )
     }
-    
-    /// Marks all missing zimfiles in the DB
-    static func markMissingZIMFiles() async {
-        let zimFileURLs = await ZimFileService.shared.getZIMFileURLs()
-        var missingIDs: [UUID] = []
-        for (zimFileID, url) in zimFileURLs where !FileManager.default.fileExists(atPath: url.path) {
-            missingIDs.append(zimFileID)
-        }
-        let context = Database.shared.viewContext
-        let zimRequest = ZimFile.fetchRequest(fileIDs: missingIDs)
-        guard let zimFiles = try? context.fetch(zimRequest) else {
-            return
-        }
-        for zimFile in zimFiles where !zimFile.isMissing {
-            zimFile.isMissing = true
-        }
-        await MainActor.run {
-            if context.hasChanges {
-                try? context.save()
-            }
-        }
-    }
 
     /// Scan a directory and open available zim files inside it
     /// - Parameter url: directory to scan


kelson42:
@BPerlakiH I don't understand why this PR is related to the "Documents folder" considering ZIM can be at any place (almost) on the filesystem.

BPerlakiH:
> @BPerlakiH I don't understand why this PR is related to the "Documents folder" considering ZIM can be at any place (almost) on the filesystem.

Sorry, wrong description, fixed it now. The solution is not specific to the "Documents" folder in any way.

kelson42:
How does this file monitoring work? The system triggers our code when a file disappears? Or we do actively check and the the question is how-often? At which occasion?

BPerlakiH:
@kelson42 My understanding is the following:

We have a system event listener (DirectoryMonitor) that is hooked into "write" events on the Documents folder, which I suppose solves the issue of auto discovery of newly added ZIM files. It is more of a system level trigger and has no timer attached.

Now, with this PR, I've also added 2 new events to look out for: both on "app start" and on "app becoming active/foregrounded": we check if the ZIM files in opened tabs are still reachable on the file system, if not we close those tabs.

kelson42:
> which I suppose solves the issue of auto discovery of newly added ZIM files. It is more of a system level trigger and has no timer attached.

This phrasing is a bit elliptic to me. Which problem/issue are you referencing? Do you mean we could auto load any new ZIM file written on the device?

BPerlakiH:
What we have is the following sequence of storing paths:

real file system path of a zim file --> [security-scoped bookmark data](https://developer.apple.com/documentation/professional-video-applications/using-bookmark-data) --> local sqlite DB entry

In the DB we associate the ZIM file metadata (it's name, id etc to the bookmark data).
If we want to open a ZIM file, we need to "resolve" the bookmark data to a URL we can open.

The above synchronization done in the DirectoryMonitor is to secure that the real file paths within the Documents folder are still valid, and if any "write" change occurs, we update the bookmark data, and the local DB.

BPerlakiH:
@kelson42 
Referring to our conversation. 
This is the already existing "state", when a ZIM file is missing. (It's on main branch and in release):

https://github.com/user-attachments/assets/00c87ac5-599a-4837-a86e-da5d1f431a95

It can be either "located" again, or unlinked.

BPerlakiH:
@kelson42 I've also excluded the Bookmarks from listing, that has a missing ZIM file.

kelson42:
@BPerlakiH Please fix onflicts by rebasing

Issue #1302, opened 2025-09-17
iOS 26 bottom bar to be fixed


BPerlakiH:
The current bottom bar we have on iOS (26) is not fully fitting into most screens:

<img width="1320" height="2868" alt="Image" src="https://github.com/user-attachments/assets/e0d701ab-6085-4dcc-8523-20cf99bf79a5" />

BPerlakiH:
@kelson42 after some research, I've found that we can do a "more" button on the toolbar, that could contain the less prominent buttons.

## on iOS 18:

https://github.com/user-attachments/assets/cee9bae6-6058-4bf3-9487-fbe6b7364155

## on iOS 26:

https://github.com/user-attachments/assets/81dfde7f-20da-488e-a9c8-7094135d81fd

BPerlakiH:
Also there's another interesting solution Apple does in the new Safari. They are conditionally hiding the forward navigation button. At the moment in Kiwix we always display it, just disable it when there's nothing to navigate forward.

I think we could apply the same technique easily, de-cluttering our toolbar even more:

https://github.com/user-attachments/assets/ca662f47-a63a-4794-b784-597c2f980eeb

BPerlakiH:
So something like this:

https://github.com/user-attachments/assets/91617e7e-aa77-4b1d-957f-5e02b45b25de

kelson42:
OK for partial hidding of forward button
OK for the more button with inside "random" and "share" (in that order

BPerlakiH:
@kelson42 please have a look:

https://github.com/user-attachments/assets/887e21c0-8713-4684-9a3b-75c8a06f22ee

PR #1303, opened 2025-09-17
Fix toolbar spacing for iOS 26

BPerlakiH:
Fixes: #1302 

Simulator Screenshot - iPhone 17 Pro Max - 2025-09-17 at 21 20 46 Medium

### DIFF
--- a/Views/Buttons/MoreTabButton.swift
+++ b/Views/Buttons/MoreTabButton.swift
@@ -18,7 +18,7 @@ import SwiftUI
 
 struct MoreTabButton: View {
     
-    let browser: BrowserViewModel
+    @ObservedObject var browser: BrowserViewModel
     @FocusedValue(\.hasZIMFiles) var hasZimFiles
     
     @State private var menuPopOver = false


kelson42:
@BPerlakiH I see that the bottom bar is now fooating. Not a big fan, but I guess this is the new norm. What I have a problem with, is there is no padding left and right, the bar just touch the border of the screen. This is really not looking nice.

BPerlakiH:
@kelson42 Yes, floating bars is the new thing for iOS. 
I think our best / quickest solution would be to move out one of the buttons (eg. the random button), there's no more space left for us really.

There are also a new "compact" mode for safari tab bars, but that's also user controlled:
https://www.idownloadblog.com/2025/06/12/ios-26-safari-tab-bar-designs/

kelson42:
@BPerlakiH What is the status here? You have made additional commits, but I hvae no clue what has changed as I have no additional comment, screecast update.

BPerlakiH:
I've added the recordings to the issue itself:
https://github.com/kiwix/kiwix-apple/issues/1302#issuecomment-3315132741

BPerlakiH:
Adding it here as well:

https://github.com/user-attachments/assets/887e21c0-8713-4684-9a3b-75c8a06f22ee

kelson42:
@BPerlakiH bookmark should not be behind the more button IMO

BPerlakiH:
@kelson42 I've tried to leave the bookmarks button out of the popover menu, but it's just becoming very strange to be honest. We have a popover that only "saves" us 1 space really, as the (...) button is taking up 1 space, and we have (max) 2 buttons in it (the random and the share):

https://github.com/user-attachments/assets/79484c9b-ac4d-4a50-9fa7-5b7fc6c37d32

Since Safari is also "hiding" the bookmark menu under a more menu popover, I think it's legit for us to do it as well:


https://github.com/user-attachments/assets/4b77d57c-78e1-409b-a642-4db71d96b1f6

NOTE: that on Safari even the tabs are hidden under the popover.


Overall I think it's better the way it is now.

Issue #1308, opened 2025-09-19
Proper error message if download fails like in #1295


kelson42:
As a follower of #1295

We have to be sure the whole ZIM file has been moved/created at the right place so we can open it BEFORE we say that everything is good. And if not, we should display an error like `Kiwix download failed to succeed, ZIM file not available at expected path "/xyz/foobar"`.

Issue #1313, opened 2025-09-26
Diagnostic Report feature


BPerlakiH:
[As discussed](https://github.com/kiwix/kiwix-apple/issues/1295#issuecomment-3276380135), we should add an easy way for user to report technical issues, and send us a Diagnostic Report, similar to Kiwix Android, see here: https://github.com/kiwix/kiwix-android/issues/4095

BPerlakiH:
@kelson42 Can we setup a dedicated email address for Apple reports ?

kelson42:
> @kelson42 Can we setup a dedicated email address for Apple reports ?

I believe we have one somewhere, but I would prefer that people share it via copy/paste in a GitHub issue or on Slack. Any reason not to do so?

rgaudin:
Obviously it requires an account on Github or Slack so several steps decreasing the chances the person goes through.

BPerlakiH:
I agree with @rgaudin, it's much harder to achieve this without a friction to our end users.
For example my phone allows to "share" to Slack, but that requires the Slack app to be installed, logged into and pick the Kiwix workspace, and then select the right destination. There's no out of the box equivalent of that for a Github issue though.

With the email we can automate for the user (as it is on Android) to have the address, the subject, and the logs attached, and start the user's preferred email client.

We could try to automate it from our end, of receiving the email and creating an issue / sending it to slack. We can think of some precautions, in order it not to be abused (like we probably don't want to post junk mail content to github).

I think the optimal approach is to start with the e-mail and figure out how many of those we receive. If it becomes burdensome to hand-pick them and convert them to Github issues, we can think about automating that part.

kelson42:
OK, your points make sense. Just a note on my perspective on things: I have been running a reflexsion about the Android diagnostic feature which needs to be improved, see https://github.com/kiwix/kiwix-android/issues/4014. Just try to take the opportunity to merge the effort and achieve to achieve a similar feature on both OSes.

Then from my understanding, the feature should allow:
* To send the diagnostic via Email
* To "share" the diagnostic via iOS (macOS?) share feature (and should allow to copy as well).

I will create a dedicated Email box, or maybe just a redirect to the developer mailbox? What about `apple-diagnostic@kiwix...` ?

BPerlakiH:
Thank you. After reading the current state on Android, maybe we should do it in a slightly different way.
[...]

Issue #1295, opened 2025-09-07
Apparent failure to download wikipedia.


resuna:
This was my first time using Kiwix. I downloaded the app from `https://download.kiwix.org/release/kiwix-desktop-macos/kiwix-desktop-macos.dmg` and installed it. 

I requested a download of the full wikipedia with all the mod cons from the Kiwix user environment, seems to be `https://download.kiwix.org/zim/wikipedia/wikipedia_en_all_maxi_2025-08.zim.meta4`. It spent an expected amount of time downloading it, but it doesn't seem to have actually downloaded and installed anything.

When I open the app I don't see it and I don't see where it was downloaded. I would have expected that to be displayed front and center, but it doesn't seem to think it has any local content at all. There's no download locations in the settings. Do I have to do "find ~ -name *.zim" to open it?

plutil -p /Users/peter/Library/Containers/self.Kiwix/Data/Library/Preferences/self.Kiwix.plist

The only big chunk of data is the WAL file for an sqlite database in /Users/peter/Library/Containers/self.Kiwix/Data/Library/Caches/self.Kiwix/

The read me for this version seems only to be related to building it. Are there any better directions, where am I expected to find the content that I downloaded?

BPerlakiH:
Can you tell us which version of macOS you are on, and which version of Kiwix you have please?

resuna:
Kiwix: Version 3.10.1 (230)

Mac OS: Sequoia 15.6.1 intel

I selected the 119.27GB image from categories.

<img width="437" height="234" alt="Image" src="https://github.com/user-attachments/assets/b143c384-251e-493e-8cbc-3560a5a9ae89" />

BPerlakiH:
Can you check if there's anything under "Opened" or under "Downloads" tab ?

<img width="158" height="182" alt="Image" src="https://github.com/user-attachments/assets/02977913-bf8a-4500-98a6-181d17c8af79" />

resuna:
"No opened Zim file"

"No download tasks"

BPerlakiH:
Thank you.
Can you try to download something really small (in size), to see if that is working for you ?

Eg: you can try to look under Categories / Others.

BPerlakiH:
<img width="556" height="303" alt="Image" src="https://github.com/user-attachments/assets/e7708176-a961-44e8-bb5c-fb3c67d1b16e" />

resuna:
[...]

PR #1314, opened 2025-09-26
Add detailed download errors

BPerlakiH:
Fixes: #1308 

We can now display more detailed errors in various points of the download process:

Screenshot 2025-09-26 at 23 10 47
Screenshot 2025-09-26 at 23 09 29
Screenshot 2025-09-26 at 23 08 28

### DIFF
--- a/Model/Downloads/DownloadService.swift
+++ b/Model/Downloads/DownloadService.swift
@@ -258,6 +258,16 @@ final class DownloadService: NSObject, URLSessionDelegate, URLSessionTaskDelegat
                                total: totalBytesExpectedToWrite)
         }
     }
+    
+    private func showAlert(_ alert: ActiveAlert) {
+        Task { @MainActor in
+            NotificationCenter.default.post(
+                name: .alert,
+                object: nil,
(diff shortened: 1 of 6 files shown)


BPerlakiH:
I also tried to add the temporary file path to the message like so:

Screenshot 2025-09-26 at 23 24 50


but in the end I removed it, since there's nothing the user can do with that really.
It's a no follow path, so it's no use, not even on macOS (possibly even worse on mobile devices).

kelson42:
Woukd this display @ proper error for #1295?
Please add return carriage between code and description
Considering this error should be used to open an issue, we should add a copy button

BPerlakiH:
@kelson42 yes this should cover all the possible errors mentioned in #1295.
I also did found that the alert function we were using was deprecated by Apple, and it wasn't working properly on iOS that is also fixed.

I don't think we should add the copy button.
The way I see it, is that all the diagnostic logs are collected, and can be sent to as once we have https://github.com/kiwix/kiwix-apple/issues/1313 implemented.
In the meantime the user can do a screenshot and send that.

PR #1318, opened 2025-10-03
Place searchfield to the left

### LINKED ISSUE #1317: Move search field to the left
For the moment, the search field is on the very right

Image

I would prefer to have left, on the right of the history navigation button because:
* this is how it is on a browser
* this is like it is on other ports of kiwix
* search resukts would be displayed directly below the search box

BPerlakiH:
Fixes #1317 

# iPad
Simulator Screenshot - iPad Air 13-inch (iOS 26) - 2025-10-04 at 00 47 02 Medium

# macOS
It can be placed to the left, but it won't work for us, as search suggestions are not showing up in this case. Probably a SwiftUI bug.

# iPhone
not applicable: the search-bar is already in the middle of the screen in both landscape and portrait

### DIFF
--- a/Views/BrowserTab.swift
+++ b/Views/BrowserTab.swift
@@ -129,7 +129,11 @@ struct BrowserTab: View {
         .focusedSceneValue(\.canGoBack, browser.canGoBack)
         .focusedSceneValue(\.canGoForward, browser.canGoForward)
         .modifier(ExternalLinkHandler(externalURL: $browser.externalURL))
-        .searchable(text: $search.searchText, placement: .toolbarPrincipal, prompt: LocalString.common_search)
+        .searchable(
+            text: $search.searchText,
+            placement: searchFieldPlacement(),
+            prompt: LocalString.common_search
+        )
         .onChange(of: scenePhase) { [weak browser] newValue in
             if case .active = newValue {
                 browser?.refreshVideoState()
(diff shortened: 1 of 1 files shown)


kelson42:
@BPerlakiH CI fails ;(

BPerlakiH:
hmm... looks strange: https://developer.apple.com/documentation/swiftui/searchfieldplacement/toolbarprincipal
it is available in our range of targets and builds Ok locally...
Looking into it... maybe the Xcode version on CI is struggling with it.

kelson42:
> hmm... looks strange: https://developer.apple.com/documentation/swiftui/searchfieldplacement/toolbarprincipal it is available in our range of targets and builds Ok locally... Looking into it... maybe the Xcode version on CI is struggling with it.

@rgaudin Anything related to what we are doing on jiwix-build?

rgaudin:
> @rgaudin Anything related to what we are doing on jiwix-build?

No

BPerlakiH:
@kelson42 together with [the Xcode / macOS runner update it](https://github.com/kiwix/kiwix-apple/pull/1319) can be merged.
Those 2 changes together build OK, see here:
https://github.com/kiwix/kiwix-apple/actions/runs/18248231247/job/51959116965

For this reason, I will re-target this PR, to indicate the dependency.

BPerlakiH:
There's another issue, that I've discovered with this. It's a no go for macOS, as the search suggestions are not showing up if I change it to be on the left. Don't know why, probably a SwiftUI bug.

So I've updated the code to move the search bar to the left only for iOS.

Issue #1035, opened 2024-11-25
Suggest similarly spelled suggestions results


kelson42:
In case someone has mispelled a word, we should be able to propose a correction

Blocked by https://github.com/openzim/libzim/issues/731

kelson42:
> Related to this, we can do the following suggesting solution from Apple: https://developer.apple.com/documentation/swiftui/suggesting-search-terms

Seems a technical answer, not sure I want to introduce this kind of widget right now... even if this is how other Kiwix ports handle the suggestions.

kelson42:
@BPerlakiH I have refine the requirement. Could you please in TOP prio propose a mockup so we can look with DWDS if this could make the job?

BPerlakiH:
@kelson42 on Apple we are searching multiple ZIM files "at once". DWDS is being a special case here, as a custom app it has only 1 ZIM file.
I suppose we should directly aim to support multi ZIM files with the suggestions as well.

BPerlakiH:
My thinking here is that if there are no results at all (from non of the ZIM files), we can display 1 suggestion from each ZIM file. (We can cut down the duplicates though.)
For DWDS it would solve their issue, but it would work equally well for a multi ZIM file scenario as well.

kelson42:
@BPerlakiH For the moment please implement the feature as an option for custom app only... so only for one ZIM file.

That said, your suggestion seems good, butblets keep focus on DWDS.

BPerlakiH:
@kelson42 @gremid here's a preview how this feature for DWDS could work on macOS and on iPad. The iPhone one is still under construction:

# iPad

https://github.com/user-attachments/assets/1d5f629f-9a51-4631-8e03-9472db27583d

# macOS

https://github.com/user-attachments/assets/f96658cb-0cbe-4800-ad0b-7e9386c19903

kelson42:
@BPerlakiH Thank you for the mockups, to me, somehow, we need to clearly indicate that this is a spell checker suggestion/correction. Therefore the sentence "Do you mean...". My feeling is that current approach is not "clear enough".

I woukd tend to just reuse the current approach to display suggestions, but insteqd to just display one item, I would propose to put just above (so outside the suggestion item) "Do you mean...".

gremid:
[...]

PR #1326, opened 2025-10-09
1035 Spellchecked search results

BPerlakiH:
Fixes: #1035

Adding alternative list items for search suggestions.
Adding feature flag ``SHOW_SEARCH_SUGGESTIONS``, this is going to be activated for DWDS only at the moment. This way it can be merged immediately with no impact on other builds.


kelson42:
@BPerlakiH Thank you for preparing this PR. We are only one step away to merge the feature in libkiwix.

Please have a look to how we demo/use yhe feature in kiwix-tools at https://github.com/kiwix/kiwix-tools/pull/768

What is important to know is that creation of SpellingsDB will take maybe 20s the first time is called because of creation of the index. Therefore it is essential that:
* It is called at start of the app (again: only dwds)
* it runs in background without freezing UI

BPerlakiH:
@kelson42 I've also updated this PR. Now it should contain everything to make the search suggestions work. It does support multiple ZIM files as well, we just need to enable the SHOW_SEARCH_SUGGESTIONS feature flag.

That feature flag should be enabled for DWDS builds, once we merge the kiwix-apple-custom PR here: https://github.com/kiwix/kiwix-apple-custom/pull/126

And **by putting these 2 things together DWDS should be ready to test.**

I do have a slight variation of this PR as well ready, which is using swift task instead of NSOperation, but this has less changes, so we should start with testing this out.

kelson42:
@BPerlakiH Can you please upload screencast of dwds first launch and test of the feature? I would like to see how goes (hopefuly I see nothing) spellchecker indexation and have @gremid be able to give feedback.

BPerlakiH:
we need this to be merged first for the CI test to pass properly:
https://github.com/kiwix/kiwix-apple/pull/1359

it is hanging on the catalog fetching even during tests...

kelson42:
Please rebase an fix conflicts

BPerlakiH:
@kelson42 @gremid
I've tested this feature with a fresh install (starting indexing from start), and it is quite user friendly, it can be used almost* immediately:

# macOS:

https://github.com/user-attachments/assets/eec1fe62-4ffa-452e-8e8a-80fca90d0cec

# iPad:

https://github.com/user-attachments/assets/ca742e80-4868-4beb-bbaa-a901ddb482fb

# iPhone:

https://github.com/user-attachments/assets/7dc644d0-eb46-4a71-9bae-20d78d4f7486

*When I tried to be "unhumanly fast" and immediately start searching on first launch, it will show no results first, and then populates the results as normal, which is still acceptable as it "self corrects itself" as the user types when it reaches the "indexed state". I wouldn't worry about this case though, it is not likely to happen to a real user, and we handle it gracefully.

PR #1343, opened 2025-10-27

translatewiki:
Translation updates


Support/es-formal.lproj/Localizable.strings
BPerlakiH: @kelson42 it seems we do have an issue on the Translatewiki side of it.
This file should be deleted there as well.

BPerlakiH (approved):

PR #1345, opened 2025-10-28
Revert search field on left for macOS

### LINKED ISSUE #1344: Revert Move search field to the left on macOS
The improvement done as part of: [Move search field to the left](https://github.com/kiwix/kiwix-apple/pull/1318)
is not working fully on macOS.
After investigating some dismiss issues, it turned out that since we moved the search field, the dismissal of it is not working as expected on macOS (iOS seems to be fine).
It seems like a strange SwiftUI issue from Apple.

After searching and selecting the item, you can see that the page navigates to the right place (the title of the page is changing properly), but the search is not dismissed as expected and stays in an empty mode.
Clicking in an empty space dismisses it as expected:
https://github.com/user-attachments/assets/dde78ce4-b16b-45c9-a24b-2bbce2eab621

When I reverted the above change for macOS only it works OK.

TLDR: we cannot have the search field on the left side for macOS in a reliable way at the moment.

BPerlakiH:
Fixes: #1344

### DIFF
--- a/Views/BrowserTab.swift
+++ b/Views/BrowserTab.swift
@@ -23,9 +23,16 @@ struct BrowserTab: View {
     @Environment(\.scenePhase) private var scenePhase
     @EnvironmentObject private var library: LibraryViewModel
     @StateObject private var search = SearchViewModel.shared
+    private let searchFieldPlacement: SearchFieldPlacement
     
     init(tabID: NSManagedObjectID) {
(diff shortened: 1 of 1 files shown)


BPerlakiH:
No longer needed, the real underlying issue is better fixed with: https://github.com/kiwix/kiwix-apple/pull/1358

Issue #1340, opened 2025-10-23
macOS crash


BPerlakiH:
As reported by @benoit74, there's an often occurring crash on release 3.11.0 on macOS.

<img width="688" height="317" alt="Image" src="https://github.com/user-attachments/assets/5ae21841-b4be-4eda-9ca2-e15e179977e1" />

BPerlakiH:
@benoit74 can you list out the steps you took, to reproduce this issue please?

kelson42:
Do we have 3.10 impacted? If "yes", why we don't have seen it in the iTunes reports?

benoit74:
It happened to me only once, and I failed to reproduce so I didn't reported.

What I suspect (but this is really a speculation, not sure you should follow this path too long) is that it might have something to do with the fact that the app was open since quite a long time and many ZIMs in the "Opened" tab have been gone missing (like they have been deleted since the app was opened). I opened a new ZIM, browsed it a bit, left the Kiwix application, and then it crashed (if I remember well). When I reopened the application, all gone ZIMs were properly removed from the "Opened" tab. Again, this might have nothing to do with the crash but simply that app cleaned things at next startup.

kelson42:
@benoit74 I see that you justa have reported a new testflight crash report on macOS. From your persoextive, same bug as always?
@BPerlakiH I reopen the issue, please have a look to the crash log and let us know if this is the same or if we should open another issue

benoit74:
Still no clue on what I did which could have caused this crash. I just opened a ZIM by double-clicking it in the finder. I had a spinning well for some time and then crash. And second time it worked well.

BPerlakiH:
Thank you @benoit74 

So far I couldn't directly reproduce it, and the crash logs are somewhat vague.
I did found 1 issue on macOS, and fixed it together with similar ones in the code base:
https://github.com/kiwix/kiwix-apple/pull/1373

Will keep an eye on this.

kelson42:
@benoit74 Once release on testflight on next Monday, can you please test again and let us know if the bug is fixed?

kelson42:
@benoit74 Can you please confirm this is fixed in 3.11.0?

benoit74:
I did not experienced it again. But I never achieved to reproduce it reliably

kelson42:
Lets close it for the moment then.

PR #1354, opened 2025-11-01
Avoid crashes on macOS

BPerlakiH:
Fixes: #1340 

It seems that the problem started after this PR: https://github.com/kiwix/kiwix-apple/pull/1301

This crash is NOT on the AppStore version of Kiwix.

**Reproduction steps:**
- open a couple of local ZIM files (not downloading)
- open the main page in at least one of them
- start unlinking them one by one
- with a high change at some point it will crash

**Solution:**
Due to the delicacy of CoreData (local DB handling), it's a lot safer to keep a side loaded ZIMFile entry in place, than trying to remove it from the DB. We do exactly the same for downloaded ZIM files, their metadata entry is not removed from the DB.

Added a couple of more safeguards as well for the formatters as well.

I've also checked the case when there's a newer version of the ZIM file to be downloaded, and there's no problem with that either.

### DIFF
--- a/Tests/CategoryFetchingTests.swift
+++ b/Tests/CategoryFetchingTests.swift
@@ -25,6 +25,7 @@ final class CategoryFetchingTests: XCTestCase {
         try resetDB()
     }
 
+    @MainActor
     func testFilteredOutByLanguage() throws {
         // insert a zimFile
         let context = Database.shared.viewContext
@@ -43,6 +44,7 @@ final class CategoryFetchingTests: XCTestCase {
         XCTAssertTrue(results.isEmpty)
     }
 
+    @MainActor
     func testCanBeFoundByLanguage() throws {
         // insert a zimFile
         let context = Database.shared.viewContext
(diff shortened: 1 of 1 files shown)


kelson42:
@BPerlakiH Why does it crash?

BPerlakiH:
@kelson42 

If we delete a ZIMFile from the DB, it might get into a strange state where the view is holding on to a ZIMFile instance, but the DB cleared it out from the context, leading to a crash, when trying to access the properties of that instance.
Most of these DB (Core Data) objects are very much not thread safe, and they are based on Objective-C's NSObject, which makes things even harder.
Fixing these kind of issues seems to be a larger piece of work, and we should address them as part of [migrating to Swift 6](https://github.com/kiwix/kiwix-apple/issues/1357).

kelson42:
So entries are forever in the DB and just somehow disabled/ignored?

BPerlakiH:
> So entries are forever in the DB and just somehow disabled/ignored?

I double checked it how it will behave. Once we fully fetch the catalog again (past caching and all that), we are removing old ZIM file metadata that is not included in the catalog listing, and this way such entries will also be removed. So we have "eventual consistency" in the end.

kelson42:
@BPerlakiH OK... not super happy to see that fetching the online catalog has a side-effect on the side-loaded ZIM file in the local library... Anyway...

But CI is still red.

Issue #1347, opened 2025-10-31
Suggestions should be smarter to avoid flickering


kelson42:
We see here a lot of flickering between displaying suggestions https://github.com/user-attachments/assets/dde78ce4-b16b-45c9-a24b-2bbce2eab621

kelson42:
First remarks:
* This happens only if the results are printet/get quickly - so probably not if the result snippets are displayed.
* Please read https://github.com/kiwix/kiwix-android/issues/4438, https://github.com/kiwix/kiwix-android/issues/3608  and associated issues and PR

Here three ideas, each of them should IMHO be implemented, to improve the situation:
* We need debounce delay to avoid to initiate search request to the libzim each keystroke if they happen quickly.
* We have a flash because we display the spinner on empty result. We should consider displaying a spinner without clearing all the results
* Display only the spinner if it takes time to get the results, only after a certain delay and probably (maybe around 15ms after lizim request)

PR #1356, opened 2025-11-01
Search results loader on top

BPerlakiH:
Fixes: #1347 

# BEFORE:

https://github.com/user-attachments/assets/7f52340f-00ab-47a1-af75-6c3b8aed1aaa

# AFTER:

https://github.com/user-attachments/assets/30b5c522-cfc5-403c-aadd-f90c07125728

### DIFF
--- a/Views/SearchResults.swift
+++ b/Views/SearchResults.swift
@@ -70,17 +70,7 @@ struct SearchResults: View {
 
     @ViewBuilder
     var content: some View {
-        if viewModel.inProgress {
-            VStack {
-                Spacer()
-                HStack {
-                    Spacer()
-                    ProgressView()
-                    Spacer()
-                }
-                Spacer()
-            }
-        } else if viewModel.results.isEmpty {
+        if viewModel.results.isEmpty {
             if viewModel.searchText.isEmpty {
                 Spacer()
             } else {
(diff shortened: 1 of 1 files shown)


kelson42:
@BPerlakiH How does it looks like if you have many results? I'm concerned about how ugly a simple supperposition with full transparency might look like.

BPerlakiH:
@kelson42 multi results:


https://github.com/user-attachments/assets/bb158dc9-4ffa-4e39-ac49-60ddd091725d

kelson42:
@BPerlakiH Can we have something nicer? We kind of need a backgorund here to avoid superposition IMHO?

BPerlakiH:
> @BPerlakiH Can we have something nicer? We kind of need a backgorund here to avoid superposition IMHO?

I am not sure what you mean by "superposition". Is it that the search results are switching places ?

kelson42:
> > @BPerlakiH Can we have something nicer? We kind of need a backgorund here to avoid superposition IMHO?
> 
> I am not sure what you mean by "superposition". Is it that the search results are switching places ?

Because the spider has a transparent background, it is hardly visible and it looks ugly.

kelson42:
@BPerlakiH CI should be green and branch rebased to resolve conflicts

BPerlakiH:
> Because the spider has a transparent background, it is hardly visible and it looks ugly.

This is the default SwiftUI progress indicator we use everywhere...

kelson42:
merging and will open a dedicated issue after thinking  bit more about that

Issue #1352, opened 2025-11-01
Improve catalog loading speed on fresh install


BPerlakiH:
When checking the latest libkiwix release, I measured the time it takes to fetch and parse the catalog, both on the latest libkiwix and former builds (on iPad), it is way over 15 seconds from the moment we display "Fetching catalog" to the moment the actual catalog appears.

To me, it just doesn't feel right, all we really do is:
- downloading 3.8MB of data (max 1-2 sec)
- parsing it
- inserting into DB

It really should not take 15+ seconds.

kelson42:
This should be done anyway at the libkiwix level, so we should move to use libkiwix and then see if we still have a problem.

BPerlakiH:
ok, so I got it now.

I think the problem is that is does have the data, but those favicons are downloaded one by one during parsing the catalog, that's why it is taking so long to process the whole catalog.

So in short, there's nothing wrong on the libkiwix side, we just should not call this favicon download function in a loop on 3000+ elements.

Sorry for the confusion, I closed the libkiwix side of the issue.

We need to do another solution on our side, that's working for custom apps (1 favicon), but does better on kiwix which has 3000+ elements.

kelson42:
Considering an illustration download should be done asynchronously (because it is slow), I don't understand why this slow down anything.

BPerlakiH:
Ok, so there are 2 major things to it:

1) **on the app / client side we have **ZimFileMetaData****
This is a class that wraps the title, filename, download path, faviconURL, faviconData etc.
This is also used to insert / and get all (or any of this) info into the local DB.
The implementation was done with a dual purpose:
- we can create the ZimFileMetaData based on a ZIM file, we have locally available (eg side loading).
- we can also create the ZimFileMetaData based on the OPDS xml feed from the fetching the catalog (in this case we do not have the ZIM file locally).
- this process of creating this ``ZimFileMetaData`` for insertion was also trying to get the faviconData. To get the faviconData, it was reaching for a libkiwix function

2) **on the libkiwix side we have the function to get the faviconData.**
As mentioned [on the libkiwix issue](https://github.com/kiwix/libkiwix/issues/1242#issuecomment-3496863577) as well, 
if it is also dual purpose:
- for a ZIM file already available, it will read the faviconData from the file itself
- for OPDS xml data parser it will try to download the faviconData using the faviconURL

We are not blocking the main UI, but we are blocking the catalog fetching itself:

```mermaid
sequenceDiagram
[...]

PR #1359, opened 2025-11-04
Fix extreme time trying to load favicon data

BPerlakiH:
Related to: #1352 

There's a clear regression issue in libkiwix 14.1.

## Context
We have this call:
```objective-c
book->getIllustrations().at(0)->getData();
```

It suppose to return the favicons image data. We are calling it as part of fetching the catalog. In reality, it never returns any data, but that's nothing new, it wasn't returning any data as of libkiwix 14.0.1 either.
(see related past issues: https://github.com/kiwix/libkiwix/issues/1180, I am not sure if that was meant to be fixed)

## The problem
The real problem is **how long it tries** to do return that data (even-though the result is always nil).
I've checked the commit merging libkiwix 14.1, and the one right before, and compared the builds from both by running fresh installs on iPad.
On **libkiwix 14.0.1** the average time to fetch the catalog is about **20 sec.**
On **libkiwix 14.1,** I got fed up measuring it when it went **over 1 minute 30 sec** (and still not finished doing it).

## Solution
If we bypass this call, and return directly ``nil``, the outcome will be the same, but we don't need to wait for it, and we get back to our expected ~20 seconds loading time.

### DIFF
--- a/Model/Entities/ZimFileMetaData/ZimFileMetaData.mm
+++ b/Model/Entities/ZimFileMetaData/ZimFileMetaData.mm
@@ -157,16 +157,20 @@
 }
 
 - (NSData * _Nullable)getFaviconDataFromBook:(kiwix::Book *)book {
-    try {
-        std::string dataString = book->getIllustrations().at(0)->getData();
-        if(dataString.length() == 0) {
-            return nil;
-        }
-        NSData *favIconData = [NSData dataWithBytes: dataString.data() length: dataString.length()];
-        return favIconData;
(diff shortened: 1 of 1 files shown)


kelson42:
@BPerlakiH Open an issue at libkiwix if you think libkiwix has a bug

BPerlakiH:
> @BPerlakiH Open an issue at libkiwix if you think libkiwix has a bug

https://github.com/kiwix/libkiwix/issues/1242

PR #1360, opened 2025-11-06
Disable favicon fetching during catalog parsing

BPerlakiH:
Related: #1352 

As described here:
https://github.com/kiwix/kiwix-apple/issues/1352#issuecomment-3495818719

we should not fetch the each favicon while parsing the whole catalog, it takes forever.
On the otherhand we should try to get the favicon:
- for custom app (built in ZIM file)
- locally added ZIM files (even if offline)

So the solution is to parameterize our solution, that is currently re-used for all those cases


BPerlakiH:
Note: it's a more sophisticated solution than the already merged PR: https://github.com/kiwix/kiwix-apple/pull/1359

PR #1365, opened 2025-11-09
Update min version to iOS17.6 macOS14.6

### LINKED ISSUE #1315: Update minimum versions: iOS 17, macOS 14
In order to move away from deprecated Apple APIs, we should keep supporting only the last 3 major versions.
Since iOS 26 and macOS 26 is out, we can bump up the minimum versions to:

- macOS 14
- iOS 17

BPerlakiH:
Fixes #1315

### DIFF
--- a/Views/Settings/Settings.swift
+++ b/Views/Settings/Settings.swift
@@ -285,7 +285,9 @@ struct Settings: View {
             Text(LocalString.backup_settings_header_text)
         } footer: {
             Text(LocalString.backup_settings_footer_text)
-        }.onChange(of: backupDocumentDirectory) { _, newValue in LibraryOperations.applyFileBackupSetting(isEnabled: newValue) }
+        }.onChange(of: backupDocumentDirectory) { _, newValue in
+            LibraryOperations.applyFileBackupSetting(isEnabled: newValue)
+        }
     }
 
     var miscellaneous: some View {


Lord-Kamina:
Thank you very much for effectively preventing me from using the app unless I build it myself, for literally no real reason.

BPerlakiH:
@Lord-Kamina I do understand your point of view, and that it might look like there's no reason for a min. OS update.
Please try to understand us as well:
- we clearly stated in our [README](https://github.com/kiwix/kiwix-apple/tree/main?tab=readme-ov-file#download): "Most recent versions of Kiwix support the three latest major versions of the OSes (either iOS or macOS)."
- we need to make sure that our apps are working on those OSes
- we need to do special if/else to be able to use new features of newer OSes, and do fall-backs for older ones
- we only have limited amount of devices to test on

kelson42:
> Thank you very much for effectively preventing me from using the app unless I build it myself, for literally no real reason.

AFAIK you can still download older versions of Kiwix. Therefore  you can still install and use Kiwix.

Lord-Kamina:
> > Thank you very much for effectively preventing me from using the app unless I build it myself, for literally no real reason.
> 
> AFAIK you can still download older versions of Kiwix. Therefore  you can still install and use Kiwix.
> 

I managed to find them by dicking around with the download link, abecause the server has directory listings on, but I don't think they're actually linked to anywhere from anywhere on the site (or at least, that I could find)

kelson42:
> > > Thank you very much for effectively preventing me from using the app unless I build it myself, for literally no real reason.
> > 
> > AFAIK you can still download older versions of Kiwix. Therefore  you can still install and use Kiwix.
> > 
> 
> I managed to find them by dicking around with the download link, abecause the server has directory listings on, but I don't think they're actually linked to anywhere from anywhere on the site (or at least, that I could find)

They are not linked from the Web site, but you can get the most appropriate with the Itunes AFAIK.

Lord-Kamina:
Nope. The App Store just flat-out didn't let me download it, which is how the frustration started.

kelson42:
> Nope. The App Store just flat-out didn't let me download it, which is how the frustration started.

OK, I have open an issue to improve the visibility of these older dmg, see https://github.com/kiwix/web/issues/313

Lord-Kamina:
> > Nope. The App Store just flat-out didn't let me download it, which is how the frustration started.
> 
> OK, I have open an issue to improve the visibility of these older dmg, see https://github.com/kiwix/web/issues/313

Thank you

PR #1373, opened 2025-11-15
Fix macOS crash - bad access with weak context

BPerlakiH:
Related to: #1340


kelson42:
@BPerlakiH Please update changelog

Issue #1337, opened 2025-10-17
Build in a ZIM validity check feature


kelson42:
We should add feature to use it.

kelson42:
@BPerlakiH In this milestone, this is IMHO the most priority one.

BPerlakiH:
@kelson42 I put this together to a working state on macOS.

IMHO We need to tweak the notification part a bit I think.
With the original plan (above) we always warn the user that it takes a long time, but it's not always applicable.

Checksum validation, depending on the ZIM file size: 
-  20 MB => almost instant
- 150 MB => 3 seconds
- 300 MB => 6 seconds
-   9 GB => 3 minutes

So we could draw a limit in size, where it makes sense to "alert" the user that it will take a long time, but below that we could just go ahead and do it.
(I mean in 3 seconds the user might not be able to read the alert and press the button, while we could have been done with it by then.)
 What do you think ?

BPerlakiH:
@kelson42 We can do something like this on macOS:

Small files below 100MB are validated without a warning, while above will be interrupted with an alert:

https://github.com/user-attachments/assets/be61abfe-81a0-4404-97a2-a04d2386d7e3

PR #1379, opened 2025-11-23
ZIM file validation feature

BPerlakiH:
Fixes: #1337

There are 3 possible states for validation:
- not yet validated (default)
- valid
- invalid

There's a 100MB limit, below which we won't alert the user, just do the validation right away, above that limit we do warn the user before starting the validation.

If a ZIM file isMissing (meaning we do not know where it is on the disk, it might have been deleted), we do not display the validation state or the give the option to validate it (the button is not displayed).

## Not yet validated ZIM file:
Simulator Screenshot - iPad Air 13-inch (iOS 26) - 2025-11-23 at 11 18 53

## Alert before validation:
Simulator Screenshot - iPad Air 13-inch (iOS 26) - 2025-11-23 at 12 12 23

## Message during validation:
Simulator Screenshot - iPad Air 13-inch (iOS 26) - 2025-11-23 at 12 12 26

## Invalid ZIM file:
Screenshot 2025-11-23 at 11 16 32

Simulator Screenshot - iPad Air 13-inch (iOS 26) - 2025-11-23 at 11 19 01

## Valid ZIM file:
Simulator Screenshot - iPad Air 13-inch (iOS 26) - 2025-11-23 at 11 19 40

Additionally moved out the following structs from the ZimFileDetails.swift file, as it was becomming too long:
- DownloadTaskDetail
- ServiceWorkerWarning
- FileLocator


kelson42:
@BPerlakiH Does the validation happens as well in the diagnostic?

BPerlakiH:
> @BPerlakiH Does the validation happens as well in the diagnostic?

Yes, by logging the start / end of validation and the result of it, it will be included in future diagnostics.

kelson42:
> > @BPerlakiH Does the validation happens as well in the diagnostic?
> 
> 
> 
> Yes, by logging the start / end of validation and the result of it, it will be included in future diagnostics.

The diagnostic should trigger the ZIM validity check

BPerlakiH:
> The diagnostic should trigger the ZIM validity check

Ok, I will need think about it how to approach it. 
In Kiwix you can have multiple ZIM files added, and we would need to run the validation on all of them... or on those that has not yet been validated. 
It can become a blocker in certain cases, if the aim of the diagnostics is to gather what happened in the app...
On the other-hand I do see the point of automatically ruling out content issues from software bugs.

BPerlakiH:
> The diagnostic should trigger the ZIM validity check

@kelson42 with these latest changes, now we do trigger the validation check on all ZIM files, that were not yet validated.

Issue #1380, opened 2025-11-24
Kiwix 3.11 on macOS 26.1 Tahoe crashes when opening download tab


Popolechien:
Basically the title:

<img width="320" height="400" alt="Image" src="https://github.com/user-attachments/assets/866da8ed-d118-41b8-acd0-166539d92577" />

Reported by user, feel free to close if unable to reproduce.

kelson42:
@BPerlakiH To be treated in priority like other crash reporta or download bugs.

PR #1383, opened 2025-11-29
Fix download task out of sync with zimFile

BPerlakiH:
Fixes: #1380

I managed to reproduce this.

## Issue:
The download tasks could go out of sync with the list of ZIM files in the DB, resulting in a crash.

## Solution:
Make sure we are only listing valid download tasks (not faulting ones).

There were some additional getters for zimFileId that was only masking the problem and made it harder to pin down, so I removed those, and made it more explicit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # 3.11.1
   - NEW:
     - ZIM file validation feature(@BPerlakiH #1379)
+  - FIX:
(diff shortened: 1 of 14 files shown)


BPerlakiH:
Needs some more work...

BPerlakiH:
This specific issue is fixed. In the meantime I did find another one, created a new issue for that here:
https://github.com/kiwix/kiwix-apple/issues/1384

BPerlakiH:
@kelson42 now it's good to be merged

Issue #1391, opened 2025-12-01
Improve check interity feature (and diagnostic)


kelson42:
Here is a list of things to improve:

- [x] * The overall feature is an "ZIM file integrity check", not a "validation". The code vocabulary should be changed
- [x] * The button label should be changed from "Validate" to "Check ZIM archive integrity"
- [x] * The result entry should not be mixed with the multimedia content details but put just above the button, together in a section. The label should be changed from "valid" to "ZIM archive integrity".
- [x] * When the diagnostic is saved to a file it should be called "kiwix_diagnostic_mm-dd-yyyy.txt".
- [x] * In the diagnostic each ZIM file (integrity) checked should be an item in the bullet list
- [x] * Rename the button "Share diagnositc report" to "Diagnostic report"
- [x] * At the left of the "email" and "share" button, we should have a "Run" button, and the two other buttons should be deactivated as long as the diagnostic is not (or currently) runs
- [x] * The bullets in the bullet list should be animated, so we can clearly identify  which part has been done, are doing, and is still to do.
- [x] * Replace sentence "Please share,,, " with "Diagnostic items:"
- [x] * Diagnostic items should be slightly indented,

BPerlakiH:
Splitting this into smaller bits, so it won't be a gigantic PR.
First one is here:
https://github.com/kiwix/kiwix-apple/pull/1394

kelson42:
It is important that for the diagnostic, all the ZIM files are checked from scratch, even if a check has been run before. This is essential to trust the result of the diagnostic.

kelson42:
Not sure if this has been part of an other issue or got forgotten, but considering the diagnostic can be pretty long, it is essential we can cancel it... therefore we need a "Cancel" button to let the user stop the process.

kelson42:
Closing at it seems now fully implemented

PR #1394, opened 2025-12-03
Improve check integrity naming and look

BPerlakiH:
Solving the first points from #1391:

> - The overall feature is an "ZIM file integrity check", not a "validation". The code vocabulary should be changed
>  - The button label should be changed from "Validate" to "Check ZIM archive integrity"
>  - The result entry should not be mixed with the multimedia content details but put just above the button, together in a section. The label should be changed from "valid" to "ZIM archive integrity".

## iPad
Simulator Screenshot - iPad Air 13-inch (iOS 26) - 2025-12-03 at 00 53 10
Simulator Screenshot - iPad Air 13-inch (iOS 26) - 2025-12-03 at 00 53 01
Simulator Screenshot - iPad Air 13-inch (iOS 26) - 2025-12-03 at 00 52 52

## macOS
Screenshot 2025-12-03 at 00 46 36
Screenshot 2025-12-03 at 00 46 26
Screenshot 2025-12-03 at 00 46 15

### DIFF
--- a/ViewModel/ZimIntegrityModel.swift
+++ b/ViewModel/ZimIntegrityModel.swift
@@ -41,8 +41,10 @@ final class ZimIntegrityModel: ObservableObject {
     }
     
     @MainActor
-    private func addZimFile(_ zimFile: ZimFile) {
-        checks.append(Info(zimFile: zimFile, state: .enqued))
+    private func addZimFiles(_ zimFiles: [ZimFile]) {
+        checks = zimFiles.map({ zimFile in
+            Info(zimFile: zimFile, state: .enqued)
+        })
     }
 
     @MainActor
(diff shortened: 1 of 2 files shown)


kelson42 (commented):
* On iPad, the look & feel of screenshot #3 is good and should be kept during the check itself. Which means, also keeping the "Cancel" putting to interrupt the test.
* "Checking ZIM file..." dialog title seems to me to be a bit redundant with the button "Check ZIM archive integrity". I wonder if we would not better just get rid of it.

BPerlakiH:
Simulator Screenshot - iPad Air 13-inch (iOS 26) - 2025-12-13 at 11 06 12 Medium
Simulator Screenshot - iPad Air 13-inch (iOS 26) - 2025-12-13 at 11 06 15 Medium
Simulator Screenshot - iPad Air 13-inch (iOS 26) - 2025-12-13 at 11 06 08 Medium

PR #1404, opened 2025-12-14
1391 diagnostics integrity check improvements

BPerlakiH:
From #1391 

> * In the diagnostic each ZIM file (integrity) checked should be an item in the bullet list
> * At the left of the "email" and "share" button, we should have a "Run" button, and the two other buttons should be deactivated as long as the diagnostic is not (or currently) runs
> * The bullets in the bullet list should be animated, so we can clearly identify which part has been done, are doing, and is still to do.

https://github.com/user-attachments/assets/a8e42986-999d-4deb-a9a7-27a86ded8b67

https://github.com/user-attachments/assets/a7b76a54-371f-4e35-be9d-8c7bad3692ea

https://github.com/user-attachments/assets/6bd45939-091d-4d20-8234-44d8277e0b0d

TODO: there are still some lifecycle issues to solve

### DIFF
--- a/Model/Diagnostics.swift
+++ b/Model/Diagnostics.swift
@@ -114,9 +114,9 @@ final class DiagnosticsModel: ObservableObject {
     
     func cancel() {
         Task { @MainActor in
-            items = Const.defaultItems
-            integrityModel.reset()
             cancellable?.cancel()
+            integrityModel.reset()
+            items = Const.defaultItems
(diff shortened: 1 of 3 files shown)


BPerlakiH:
I've added the cancel button, and updated the UI.
Fixed an additional issue with the "isAdmin" check on macOS.

Issue #1417, opened 2026-01-03
Improve "Download failure" dialog


BPerlakiH:
This is a followup of #1397

The current dialog should be improved:
* To propose a unique dialog (look&feel) for download errors
* Title of the book should be included
* URL should be included
* Error message should be clear about what is going on and localised
* Technical part of the issue should be given in a concise/readable/understandable manner (HTTP response code, timeout, SSL error, etc.)

<s>Change the error message displayed from: "Unable to download this file"
to: 'Unable to download "Title of the actual ZIM file"'</s>

BPerlakiH:
<img width="288" height="320" alt="Image" src="https://github.com/user-attachments/assets/f7122af9-973b-49c2-9bd0-6f85eb369a35" />

PR #1418, opened 2026-01-03
Improve download error handling

BPerlakiH:
Fixes: #1417 
Screenshot 2026-01-04 at 19 03 50
Screenshot 2026-01-04 at 18 51 20
Screenshot 2026-01-04 at 16 13 28

### DIFF
--- a/Support/da.lproj/Localizable.strings
+++ b/Support/da.lproj/Localizable.strings
@@ -220,7 +220,6 @@
 "payment.error.description" = "Der har været et problem med din betaling. Prøv venligst igen.";
 "download_service.error.client.backgroundSessionInUseByAnotherProcess" = "Kunne ikke kommunikere med overførselstjenesten i baggrunden";
 "download_service.error.client.backgroundSessionWasDisconnected" = "Mistede forbindelsen til overførselstjenesten i baggrunden";
-"download_service.error.client.None" = "ukendt fejl";
 "download_service.error.client.cancelled" = "annulleret";
 "download_service.error.client.badURL" = "ugyldig URL";
 "download_service.error.client.timedOut" = "Anmodningen udløb.";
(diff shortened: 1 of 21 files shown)


kelson42:
@BPerlakiH Please make an PR with complete revamp of this dialog including:
* Title
* Message/Details (they should be meaningful, "code 335" is of no meaning for the user).
* ..

I just have rephrase the issue description/title

BPerlakiH:
@kelson42 please have a look at the new screenshots above

kelson42:
- [x] Title should be: Unable to download "foobar" (please remove the ":")
- [x] The rest of the text should be the descripton (not bold, maybe smaller font)
- [x] The dialog messages should be localised and no string should come directly from Apple OS or SDK
- [x] Dialog error messages are Kiwix ones and triggered based on Apple error codes (so not string matching)
- [x] The original error codes coming from the system/OS should be in the log (better in English)
- [x] Just print the URL at the first line of the description (so no "From the URL:"), and but two return carriages
- [x] Print the user understandable homebrew error after that (you should have around 5 to create). Add twp carriage returns.
- [x] Last description line would be "The download could not be completed."

BPerlakiH:
Screenshot 2026-01-11 at 22 37 17

------

Screenshot 2026-01-11 at 15 54 56


-----

@kelson42 please see it updated. I've ended up with a lot more error codes than the foreseen 5. I managed to parse them out from the CFNetwork framework and add them to our localizations, so it's a good start as we have quite some languages covered.
We are matching [these specific error codes from Apple](https://developer.apple.com/documentation/foundation/urlerror#Error-codes).

kelson42:
@BPerlakiH Thx, but it seems to me you don't have respected the requirement "The dialog messages should be localised and no string should come directly from Apple OS or SDK"!?

BPerlakiH:
> @BPerlakiH Thx, but it seems to me you don't have respected the requirement "The dialog messages should be localised and no string should come directly from Apple OS or SDK"!?

@kelson42 Those are not coming from Apple OS when the app runs. 
I just copied them over to **our** localization files, we can edit them however we want.

Issue #1419, opened 2026-01-03
Properly handle "going offline/going online" and its impact of download operations


BPerlakiH:
This is a follow-up of [#1397](https://github.com/kiwix/kiwix-apple/issues/1397)

Here is how it should impact the downloads and what user should se:
* If a download is started but device offline, the download should be automatically in pause, user should be informed with a "toast" message "Device is offline. Download is paused and will resume automatically once online again."
* If someone click on pause and device is offline. This situation should not happen as download should automatically be put on pause once the device goes offline. But if it happens anyway, the user should be able to put on pause.
* The message given earlier should also be displayed  if the user wants to resume but device is offline.

kelson42:
Comment from @BPerlakiH (earlier issue description):

> Our current mechanism allows us to automatically resume a download once the user gets back on-line, however at the moment it is unclear why the download is not progressing, when the user is offline. It includes both the very beginning of the download (being stuck at 0% which was reported by one of the users), or any other moment for that matter.

> Proposing to change the UI to reflect the fact of being offline both on the download cell (Downloads section) and the Download details page.

BPerlakiH:
<img width="2732" height="2048" alt="Image" src="https://github.com/user-attachments/assets/152b6eca-0db6-4fa4-ac2a-69d5a1a4d4e0" />
<img width="2732" height="2048" alt="Image" src="https://github.com/user-attachments/assets/4dab4a91-8611-488a-a919-830e863b9af3" />

BPerlakiH:
<img width="1315" height="768" alt="Image" src="https://github.com/user-attachments/assets/7c3a776a-b861-4374-bb73-aae6bc492d8f" />

PR #1421, opened 2026-01-03
Reflect offline state in dowloads

BPerlakiH:
Fixes: #1419 

Image
Image
Image
Simulator Screenshot - iPhone 17 Pro Max - 2026-01-03 at 18 55 40

### DIFF
--- a/Model/Downloads/DownloadState.swift
+++ b/Model/Downloads/DownloadState.swift
@@ -16,15 +16,75 @@
 import Foundation
 import Combine
 
+@MainActor
+struct DownloadUIState {
+    enum State {
+        case resumed
+        case paused(isOnline: Bool)
+    }
+    
+    let percent: String?
+    let size: String
+    let state: State
+    let hasResumeData: Bool
+    let progress: Progress
+    
+    static func empty() -> DownloadUIState {
+        DownloadUIState(
+            percent: nil,
+            size: "",
+            state: .resumed,
+            hasResumeData: false,
+            progress: progressFor(downloaded: 0, total: 1)
+        )
+    }
+    
+    private init(
+        percent: String?,
+        size: String,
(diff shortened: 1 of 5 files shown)


BPerlakiH:
I think the confusion will come from the fact that the download task itself can be paused and resumed, regardless if the user is online or offline.

So let's say, I want to download 2 very large files, and press download on both of them.
After a while I realize it will take a long time, but I want one of them more, so I put the other download on pause (to use my full band-width on the first).
In this situation, after pressing pause, the other file will be marked as "paused", and the first one will continue downloading (hopefully quicker).
Now with the above requirements, if I loose my internet connection, both files will be marked as "paused", and I won't know why.

The other confusing part is that, if I don't have internet connection, I should not have the option to "resume" a download, but that's not going to be clear either.

So in a state machine description, we have:

```
A) task paused
B) task resumed:
   B1) online
   B2) offline
```

Maybe this "offline" naming is not clear enough and needs to be changed.
I think re-using the "pause" is not good either, if you look at the screenshot we still have an active "pause" button.

If I understand your idea correctly, that would mean that we try to squeeze our current 3 states:
```
- paused
- resumed online
- resumed offline
```
into 2 (visible) states:
```
- paused*
- resumed
```
with the *, that if I am offline I won't be able to "resume", and once I get back online it will magically change back to "resumed".

So far the user had the option to "pause" and "resume", now we would magically take away this option from them (if they are offline), without a clear explanation. I think that is confusing.

kelson42:
Decision (when device is offline):
* Status label should ""Paused (device is offline)"
* (Pause/Resume) Button label shuld be "Resume" but button should be greyed

BPerlakiH:
Closing in favor of: https://github.com/kiwix/kiwix-apple/pull/1424

PR #1423, opened 2026-01-04
SSH private key secret has changed its name

kelson42:
* Change the secret name to use new secret
* Improve the naming of a few variables

### DIFF
--- a/.github/upload_file.py
+++ b/.github/upload_file.py
@@ -27,7 +27,7 @@ def main() -> int:
         "--ssh-key",
         required=False,
         help="filepath to the private key to use for upload",
-        default=os.getenv("SSH_KEY", ""),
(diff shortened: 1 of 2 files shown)


BPerlakiH:
@kelson42 We have the same variables on the custom-apps side as well, although those seems not to be in use at the moment (I suspect mainly due to the fact that we do not upload DMGs for custom apps).
Maybe we should update them as well, for consistency:
https://github.com/kiwix/kiwix-apple-custom/blob/4168c85ddf3444b0bcd4d8d7c6774575bf0ae9b1/.github/workflows/cd.yml#L13
https://github.com/kiwix/kiwix-apple-custom/blob/4168c85ddf3444b0bcd4d8d7c6774575bf0ae9b1/.github/workflows/ci.yml#L16

Issue #1425, opened 2026-01-11
Search loading indicator is not immediate


BPerlakiH:
It has been reported by one of our users, that the search loading indicator is not immediately shown.
I could reproduce it on macOS as well, especially visible for the very first search:

https://github.com/user-attachments/assets/f8058a8e-9f93-418a-9d34-781462f21372

kelson42:
@BPerlakiH I don't understand why "no result" is here visible as it should [have been fixed recently](https://github.com/kiwix/kiwix-apple/pull/1355)!? Please secure that no result is written only if there is effectively no result.

BPerlakiH:
Yes, that other fix was slightly different case when there's no user input, but it's not yet loading.
This issue is specific to the loading state itself, so to be more specific we have:
[no user input] -> [user input] -> [loading*] -> [result] -> [more user input] -> [loading on top of former results] -> [new results] 

the one marked with * was only wrong.

PR #1427, opened 2026-01-12
Display progress while searching

BPerlakiH:
Fixes: #1425 

We had the progress view only displaying while there were former search results displayed, and the user was searching (eg: kept typing).
This also covers the case, when we start the searching from scratch.


kelson42:
@BPerlakiH Please post a screencast corresponding to the one in the issue.

BPerlakiH:
https://github.com/user-attachments/assets/9c6d4356-8f8b-4c7c-9cd2-932fedd23f19

Issue #1426, opened 2026-01-12
Search runs out of memory


BPerlakiH:
While working on: #1425 
I discovered there are serious issues with the memory usage of the search. On iOS it can relatively quickly lead to an out of memory crash, especially if the HTML snippets are on:

<img width="967" height="445" alt="Image" src="https://github.com/user-attachments/assets/775001d6-158f-4245-9c2b-332fd711117d" />

Without the snippets, the memory usage is still very high.

BPerlakiH:
- libkiwix 14.1.0 BAD
- libkiwix 14.1.1 BAD. libzim 9.4.0
- libkiwix 14.0.0.1 GOOD: libzim 9.2.3

BPerlakiH:
@kelson42 I opened the related issue here: https://github.com/kiwix/libkiwix/issues/1265

kelson42:
This issue is serious, but it will take time to analyse and fix. Moving to next milestone.

BPerlakiH:
> This issue is serious, but it will take time to analyse and fix. Moving to next milestone.

I am proposing to turn off snippets for this release as a quick / temp fix for our end users.
Once we find the real cause / solution we can turn it back on from the next release.

kelson42:
@BPerlakiH AFAIK nobody has complained about the impact of this bug, so I propose to just do nothing for the moment which would/might impact even more the user.

BPerlakiH:
I insist on turning it off for this release. It's a very minor inconvenience for the users. There's a whole set of crash reports backing up, that this is a real problem, even if no-one directly complained about it.
Alternative that I can see, is [to revert back libkiwix to 14.0.0.1](https://github.com/kiwix/kiwix-apple/pull/1434).

BPerlakiH:
@kelson42 please see here:
https://github.com/kiwix/kiwix-apple/pull/1433

kelson42:
@BPerlakiH Then we will wait fix of @veloman-yunkan.

PR #1433, opened 2026-01-21
1426 Turn off search snippets

BPerlakiH:
Related to #1426 

Temporarily turn off html search snippets, until we solve:
https://github.com/kiwix/libkiwix/issues/1265


kelson42:
We won't rollback this, too risky, etc...

PR #1434, opened 2026-01-21
Revert "Update to libkiwix 14.1.0"

BPerlakiH:
Related to: #1426

Alternative temp solution with reverting to libkiwix 14.0.0.1, and removing the spellchecking parts (that are not yet available).

This reverts commit 3730b6477aeef533ecd45294716163946143f86c.
Revert libkiwix update


kelson42:
Not in favour, simply too risky and we move forward, not backwars.

Issue #1447, opened 2026-01-29
iOS CI is broken


kelson42:
https://github.com/kiwix/kiwix-apple/actions/runs/21466105873/job/61829090591

rgaudin:
```
error: iOS 26.0 Platform Not Installed.
```

kelson42:
Sounds like an unpinned dependency! Or a wild move in the CI image?

kelson42:
Sounds like an unpinned dependency! Or a wild move in the CI image?

BPerlakiH:
Seems to be a change on Github's side: https://github.com/actions/runner-images/issues/13570

PR #1448, opened 2026-01-29
Update build settings to Xcode 26.1.1

BPerlakiH:
Fixes: #1447 

Due to changes in: https://github.com/actions/runner-images/issues/13570
The simplest solution is to upgrade to Xcode 26.1.1, otherwise we would need to download the runner image for each build, and that would take forever.

### DIFF
--- a/.github/actions/xcbuild/action.yml
+++ b/.github/actions/xcbuild/action.yml
@@ -53,10 +53,10 @@ runs:
           -T /usr/bin/security \
           -T /usr/bin/productbuild || true
 
-    - name: Set Xcode version (26.0.1)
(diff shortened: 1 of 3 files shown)


rgaudin:
Doesn't seem to be enough:

```
 error: SDK "iphoneos26.0" cannot be located.
```

BPerlakiH:
@rgaudin Github actions by default runs the actions ymls from main branch.
When run on the branch itself it builds OK:
https://github.com/kiwix/kiwix-apple/actions/runs/21473596132

Therefore it's good to be merged.

rgaudin (approved):

PR #1454, opened 2026-01-31
Limit memory usage

BPerlakiH:
Fixes: #1426 

As of the comment:
https://github.com/kiwix/libkiwix/issues/1265#issuecomment-3817993025

we can limit the amount of memory used for search.


kelson42:
<s>@BPerlakiH 16MB seems really little. How did you come to this limit? What is the impact for the user of such a stromg limitation? Have you made speed measurement for most basic scenarios?</s> I see this is the direct recommendation on the libkiwix issue to come back to previous behaviour.

Issue #1446, opened 2026-01-28
Bookmarks and opened files disappear


plesubc:
On Kiwix 3.11.0 (232) for MacOS 15.7.3, I am experiencing an unusual issue where the Kiwix.app doesn't seem to store:

* Bookmarks
* Opened files

However, even more weirdly, it's not necessarly *completely* reset; sometimes one or two opened files or bookmarks remain.

Removing ~/Library/Application Support/Kiwix and/or reinstalling works briefly, but then the problem reappears.

BPerlakiH:
@plesubc Thank you for reporting this. I did see this once in a while during development, but thought that it's due to clean builds.
Do you have any specific steps to recreate this ?

plesubc:
Unfortunately, I don't. It does seem to get worse after a reboot. And it feels like the ones that are more prone to disappear are the ones that I opened during the last session, but not always.

The local .zim files are stored on an external drive, but the drive is always mounted before kiwix starts, if that makes a difference somehow, which I suspect it does not.

Keeping in mind that I know almost nothing about swift and have spent all of 5 minutes looking at your code,

ZimMigration.swift line 55 migrateFrom()

```
  /// Migrates the bookmars from an old to new zim file,
    /// also updates the bookmark urls accordingly (based on the new zim id as the host of those URLs)
    /// deletes the old zim file in the DB
```

If you are serving .zims from localhost, but you are using some method of identifying that and it changes (because you are connected to a different network or something), could that trigger bookmark deletion?

Again, keep in mind that this is complete speculation based on a 30 second inspection based on a search for "delete" in a code base which I don't know in a language with which I am not familiar.

BPerlakiH:
@plesubc Thank you, we will try to re-create this issue on our end, and see where it goes wrong.

kelson42:
From my experience, the problem is around the identification of the ZIM book. Do we use the ZIM id to attach the bookmarks to a Book?

kelson42:
@BPerlakiH Having the diagnostic should allow easily to identify if the data are not saved... or if they are not rendered, right?

BPerlakiH:
I started to look into this deeper. The mentioned ZimMigration is not the issue, that is only executed for custom apps, and not the Kiwix build.
The most likely part responsible that I can see is the following:
[...]

PR #1475, opened 2026-02-09
Do not mark ZimFile as not downloaded in case of an unexpected error

BPerlakiH:
Fixes: #1446 

Hopefully it will fix the original issue. 
I cannot achieve the same effect at the moment.
If I force the code to that case, then indeed the ZimFile(s) will disappear.
For this reason, I think we should apply this fix anyway.

### DIFF
--- a/SwiftUI/Model/LibraryOperations.swift
+++ b/SwiftUI/Model/LibraryOperations.swift
@@ -93,8 +93,7 @@ ZIM file missing: \(zimFile.name, privacy: .public) |\
  \(downloadPath, privacy: .public)
 """)
             } catch {
-                zimFile.fileURLBookmark = nil
-                zimFile.isMissing = false
+                zimFile.isMissing = true
                 Log.LibraryOperations.notice("""
 ZIM file cannot be opened: \(zimFile.name, privacy: .public) |\ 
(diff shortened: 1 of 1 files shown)


kelson42:
@BPerlakiH What does younpatch exactly does? I don't understand why your comment speak about downloads... considering this tooic is not related. I also don't understand what `zimFile.fileURLBookmark` does!

BPerlakiH:
> @BPerlakiH What does your patch exactly does? I don't understand why your comment speak about downloads... considering this topic is not related. I also don't understand what `zimFile.fileURLBookmark` does!

Sorry for the confusion.
Here's the explanation:
- we store the list of ZIM files from the catalog with all their metadata. If you download something from there, it will be marked as "downloaded", by pointing the ``fileURLBookmark`` to a local file. The same thing happens if you "side-load" / add your own file to Kiwix. We read out the metadata from the ZIM file add it to the list and set the ``fileURLBookmark`` to your local file.
If you want to "unlink" a file, that will be done by settings the ``fileURLBookmark`` to nil. 
In the "Opened" section of the app we only list ZIM files that are "downloaded" (or otherwise added) based on the ``fileURLBookmark`` property (being not nil).

So if it's set for some strange reasons to nil, that makes it disappear.

Issue #1481, opened 2026-02-14
Improve order of sections in settings


kelson42:
Today it is like this:

<img width="2160" height="1620" alt="Image" src="https://github.com/user-attachments/assets/807f09c0-87ef-41ec-834e-17f78e96423f" />

Which means:
* Reading
* Downloads
* Catalog
* Misc
* Hotspot
* Backup

My opinion is:
* "Misc" should be at the end of list
* "Downloads" could be merged with "Catalog"
* "Backup" could be merged in "Misc"

What do you think?

BPerlakiH:
@kelson42 with the current solution we are a bit limited in grouping these sections.
Each section can contain of a header, body, and footer.
The header is the title, like: "Misc", "Hotspot" or "Backup".
The body is the list of items in the table.
The footer is the small printed text below the section. This is where the problem starts.
Each of those footers are related to the last item of the table.
Eg: in the "Hotspot" section, the footer text is:
"The network port used by the Hotspot, should be between ..."

In the suggested merges, we cannot have 2 footers at once.

I will try to put 2 sections one below another and see how that would look, without using a new header.

The easy change I could do immediately is moving the misc to the end and merging it with back-up:

## BEFORE:

<img width="320" height="240" alt="Image" src="https://github.com/user-attachments/assets/1acabf4a-7043-42f3-bf23-ca031a1e8598" />

## AFTER:

<img width="320" height="240" alt="Image" src="https://github.com/user-attachments/assets/5640cd71-d88f-48ac-a6d3-784d307fd47d" />

BPerlakiH:
The closest we can get with:
> "Downloads" could be merged with "Catalog"

is the following:

<img width="320" height="240" alt="Image" src="https://github.com/user-attachments/assets/79aad56e-35d7-4250-a0c9-b488b63e0121" />

PR #1487, opened 2026-02-15
1481 settings order

BPerlakiH:
Fixes: #1481 

IMG_0001 2
IMG_0002 2


kelson42:
Thinking twice about it, please move the following part to a dedicated "Backup" section positioned between "Hotspot" and "Misc"
image

BPerlakiH:
@kelson42 I've updated the PR and the screenshots above.

PR #1494, opened 2026-02-22
Allow to download ZIM files to an arbitrary location in macOS

### LINKED ISSUE #627: Permit Custom Zim Download Folders
Please add the option to customize the location of the download folder. This will permit downloading with the app to external drives.

Verryx-02:
cc @kelson42  

I couldn't find the PR template, and since my change is considered quite significant, I wrote a PR that tries to explain everything thoroughly.  
Let me know if you need further explanation or if you want me to change anything.

# Important note
I haven't run the last test yet and it would be best to avoid merging before completing them.  
I'll take care of that in the next few days.
I've opened the PR so you can start looking at it.

# Contributor checklist
- [x] I have read the [contributing guidelines](https://github.com/kiwix/.github/blob/main/CONTRIBUTING.md)
- [x] All new code is wrapped in `#if os(macOS)` 
- [x] Tested on: MacBook with Crucial external SSD model BX500 (SATA3 to USB-A to USB-C to Macbook), macOS Tahoe 26.23

# Problem description

On macOS, ZIM files always download to `~/Downloads`.  
For large ZIM files (some exceed 100 GB) this quickly fills up the mac ssd, with no option to redirect downloads to an external drive.

Simply pointing `URLSessionDownloadTask` at a different directory wouldn't fully solve this either because it writes to a temp location first and then moves the file to the destination.  
Streaming directly to the destination also removes any file size limit imposed by the source filesystem (e.g. FAT32's 4 GB cap on temp files).

Closes #627.

---

# Solution

This PR adds a direct-write download path for macOS:  
Instead of using the temp file, data is streamed straight to the destination with `URLSessionDataTask` and `URLSessionDataDelegate`.   
The iOS path (`URLSessionDownloadTask` + background session) is completely unchanged.

## Note:
The direct-write path is only used when the user sets a custom directory in the new Downloads Settings tab. With the default `~/Downloads`, everything behaves exactly as before.

---

# How it works

Data received from the network gets buffered in a 16 MB in-memory buffer, then flushed to disk in one shot when the buffer fills up or at most every 30 seconds. This keeps writes efficient on slow USB drives and avoids hammering the filesystem with thousands of tiny writes.

The `URLSessionDataDelegate` runs on a dedicated serial `OperationQueue`, completely off the main thread. At 400 Mbps you'd get something like ~780 delegate calls per second, so routing all of that through `MainActor` would be a problem. The UI gets updated every ~1 second and on each buffer flush, which is more than enough.

Downloads are resumable via HTTP Range headers — on resume it just sends `bytes=<offset>-` and picks up where it left off. The download state is saved to `UserDefaults` as a `Codable` struct after every flush.  
There's also a 30-second periodic timer that flushes the buffer and saves state, but only if at least 30 seconds have elapsed since the last save — to avoid redundant writes right after a flush-triggered save.

One thing that took some debugging: security-scoped bookmarks on macOS require you to call `startAccessingSecurityScopedResource()` on the URL you get back from resolving the bookmark data, not on a plain path URL. A plain URL just silently returns false and writes fail.  
The bookmark is resolved via `DownloadDestination.customDownloadDirectory()` every time access is needed, including when restoring interrupted downloads at launch.

Speaking of which — `DirectWriteDownloadService` is eagerly initialized at app startup (when a custom directory is set) so that `restoreInterruptedDownloads()` runs automatically.  
Interrupted downloads are restored as paused, so the user can resume them manually.

To keep the Mac awake during downloads I use `ProcessInfo.beginActivity(.userInitiated, .idleSystemSleepDisabled)`. This prevents idle sleep but not forced sleep (lid close or Apple menu → Sleep) — that's a macOS limitation that can't be worked around, even with IOKit. If the Mac does sleep mid-download, the `URLSessionDataTask` will eventually fail with a network error. When the system wakes, a `NSWorkspace.didWakeNotification` observer fires and automatically resumes any downloads that ended up in an error state.

For transient network errors (`networkConnectionLost`, `timedOut`, `cannotConnectToHost`, `cannotFindHost`, `notConnectedToInternet`) there's an automatic retry with a 3-second delay, up to 5 attempts.

Before starting a download, `validateDestination()` checks that the directory is accessible, writable, and has enough free space. For the space check it uses `volumeAvailableCapacityForImportantUsageKey` (APFS-aware, includes purgeable space) with a fallback to `volumeAvailableCapacityKey` for ExFAT, plus a 100 MB safety margin.

When a download finishes, `DirectWriteDownloadService` fires a `directWriteDownloadCompleted` notification. `DownloadService` picks it up and runs the same completion flow as background downloads: `LibraryOperations.open(url:)`, push notification, CoreData cleanup.

---

# New files

## `Model/Downloads/DirectWriteDownloadService.swift`

The existing `DownloadService` uses `URLSessionDownloadTask`, which writes to a temp file first and then moves it to the destination, so even when the target is an external drive, the full file still has to fit on the internal SSD before it gets moved. `DirectWriteDownloadService` replaces that flow on macOS: it uses `URLSessionDataTask` with `URLSessionDataDelegate` to stream data directly to the final destination.

`final class DirectWriteDownloadService: NSObject, ObservableObject, URLSessionDataDelegate`  (singleton).  
The `URLSession` runs on a dedicated serial `OperationQueue` (not in the main thread) because (for example) at 400 Mbps you get ~780 delegate calls per second and routing all of that through `MainActor` would be a bottleneck.  
To avoid the cost of synchronising with the UI on every chunk, there are two separate task registries: 
- `tasks: [UUID: DownloadTask]` protected by `NSLock` for delegate callbacks
- `@MainActor @Published var activeDownloads` updated every 1 second for the UI.   

All disk writes go through a dedicated `fileIOQueue`.  
Each task buffers incoming data in a 16 MB `dataBuffer` and flushes to disk when the buffer fills up or every 30 seconds (this keeps writes efficient on slow USB drives).  
Sleep prevention via `ProcessInfo.beginActivity(.userInitiated, .idleSystemSleepDisabled)`, wake recovery via `NSWorkspace.didWakeNotification`.  
Retry system: 5 attempts with a 3-second delay for transient network errors.  

## `Model/Downloads/DirectWriteDownloadState.swift`

When the app quits (or crashes) mid-download, we need to know how many bytes were already written so we can resume with an HTTP Range request instead of starting over.  
This struct is the checkpoint. It gets serialized to `UserDefaults` after every disk flush, so on next launch the service can pick up exactly where it left off.

`struct DirectWriteDownloadState: Codable` captures everything needed to resume a download: 
- destination URL, 
- bytes already written, 
- expected size, 
- paused flag, 
- and the server's ETag/Last-Modified headers for validation. 

Saved to `UserDefaults` under `directWriteDownloadState_<UUID>` after every flush, restored at launch via `loadAll()`. 

## `Views/Settings/DownloadLocationPicker.swift`

The user needs a way to choose where ZIM files will go. This is the SwiftUI view in the new "Downloads" settings tab.
It lets you pick a folder and shows which volume is selected and how much space is left.

`struct DownloadLocationPicker: View` with local `@State` for `currentDirectory`, `availableSpace`, `volumeName`, `isCustomDirectory`.
 - "Choose…" opens `NSOpenPanel` (directories only) -> validates write permission -> calls `DownloadDestination.setCustomDownloadDirectory()` to store a security-scoped bookmark  
- "Reset" calls `DownloadDestination.clearCustomDownloadDirectory()` and goes back to `~/Downloads`

The info text below the path adapts based on `isCustomDirectory`. `refreshCurrentDirectory()` queries `DownloadDestination` for the effective folder, available space, and volume name to keep the UI in sync.

---

# Changed files

## `Model/Downloads/DownloadDestination.swift`
New macOS-only API for directory management: 
- `effectiveDownloadFolder()` returns the custom directory if valid or falls back to `~/Downloads`.
- `customDownloadDirectory()` resolves the security-scoped bookmark and refreshes stale bookmarks automatically.
- `setCustomDownloadDirectory(_:)` and `clearCustomDownloadDirectory()` save or clear the stored directory.
- `shouldUseDirectWrite: Bool` is a routing flag used in `DownloadService`.

Additional utilities for validation and queries: 
- `validateDestination(directory:requiredBytes:)` performs pre-flight checks for access, write permission, and disk space. 
- `availableDiskSpace() -> Int64?` and `volumeName() -> String?` query the download volume.
- `filePathFor(downloadURL:)` now uses `effectiveDownloadFolder()` on macOS.

## `Model/Downloads/DownloadErrors.swift`
Added `DirectWriteDownloadError: LocalizedError`: 
- `.destinationNotAccessible`, 
- `.insufficientDiskSpace`, 
- `.volumeUnmounted`, 
- `.cannotCreateFile`, 
- `.writeError`, 
- `.rangeRequestsNotSupported`, 
- `.partialFileCorrupted`, 
- `.invalidServerResponse`, 
- `.cancelled`

## `Model/Downloads/DownloadService.swift`
- `init()` now eagerly initializes `DirectWriteDownloadService.shared` when `shouldUseDirectWrite` is true, so `restoreInterruptedDownloads()` runs at launch. 
- `setupDirectWriteObserver()` listens for `directWriteDownloadCompleted` notifications.  
- `handleDirectWriteCompletion(zimFileID:fileURL:)` opens the downloaded file and cleans up CoreData.  
- `start()` routes downloads to direct-write or background based on `shouldUseDirectWrite`.  
- `cancel()`, `pause()`, and `resume()` check `activeDownloads` first, falling back to the background session if not found.

## `Model/Diagnostics.swift`
`freeSpace()` on macOS now uses `DownloadDestination.availableDiskSpace()` so reports free space on the actual download volume.

## `Views/Library/ZimFileDetail/ZimFileDetail.swift`
`freeSpace` on macOS uses `DownloadDestination.availableDiskSpace()` so the low-space warning now checks the right volume.

## `App/App_macOS.swift`
Added `case downloads` to `SettingsTab` and `DownloadSettings()` to the Settings `TabView`.

## `Views/Settings/Settings.swift`
New `DownloadSettings` struct: Settings tab with `DownloadLocationPicker` and a short description of the feature.

## `SwiftUI/Model/DefaultKeys.swift`
Added `downloadDirectoryBookmark = Key<Data?>("downloadDirectoryBookmark")` (macOS-only).

## `Support/en.lproj/Localizable.strings`
Added localization keys for the new UI: `download_location.*` (picker labels, buttons, error messages), `download_settings.*` (settings tab title and description), `direct_write_error.*` (error messages from `DirectWriteDownloadError`).

---

# Platform isolation

Everything new is inside `#if os(macOS)`.   
iOS is not affected.

---

# Test performed (macOS Tahoe 26.23)
(Tell me if you want more testing before merging obv)

**Test 1:  Pause and resume**
- [x] Progress resumes correctly from the right offset
- [x] ZIM file opens and works after completion
- [x] No crash 

**Test 2: Cancel mid-download**
- [x] Download stops immediately
- [x] Partial file deleted from disk
- [x] No crash 

**Test 3: Multiple simultaneous downloads**
- [x] All downloads proceed without conflicts
- [x] Pausing one doesn't affect the others
- [x] ZIM files opens and works after completion
- [x] No crash

**Test 4: App restart with interrupted download** (Kiwix terminated using `command + q` during a download) 
- [x] App restarts without crash
- [x] Partial file are still on disk
- [x] Interrupted download is restored and can be resumed
- [x] ZIM file opens and works after completion
- [x] No crash

**Test 5: Wi-Fi disconnected while downloading**
- [x] Download stops immediately and resumes as soon as the Wi-Fi is turned back on.
- [x] ZIM file opens and works after completion
- [x] No crash

**Test 6: >4GB download** (4,06 GB)
- [x] Download completes without fatal errors
- [x] ZIM file opens and works after completion
- [x] No crash

# Test NOT performed

**Test 7: 190 GB Wikipedia download**
- [ ] Download completes without fatal errors
- [ ] Retries work correctly on network drops
- [ ] ZIM file opens and works after completion

### DIFF
--- a/Model/Downloads/DirectWriteDownloadService.swift
+++ b/Model/Downloads/DirectWriteDownloadService.swift
@@ -27,7 +27,7 @@ struct DownloadTaskInfo: Sendable {
     let zimFileID: UUID
     let downloadURL: URL
     let destinationURL: URL
-    let expectedTotalBytes: Int64
+    let expectedTotalBytes: UInt64
     var isPaused: Bool = false
     var error: DirectWriteDownloadError?
     var securityScopedAccess: Bool = false
@@ -75,7 +75,7 @@ final class DirectWriteDownloadService: NSObject, ObservableObject {
         )
         super.init()
 
-        sessionDelegate.onFlush = { [weak self] zimFileID, bytesWritten in
+        sessionDelegate.onFlush = { [weak self] (zimFileID: UUID, bytesWritten: UInt64) in
             self?.didFlush(zimFileID: zimFileID, bytesWritten: bytesWritten)
         }
         sessionDelegate.onCompletion = { [weak self] zimFileID, error in
@@ -129,7 +129,7 @@ final class DirectWriteDownloadService: NSObject, ObservableObject {
     // MARK: - Public Methods
 
     func start(
-        zimFileID: UUID, downloadURL: URL, expectedSize: Int64, allowsCellularAccess: Bool = true
+        zimFileID: UUID, downloadURL: URL, expectedSize: UInt64, allowsCellularAccess: Bool = true
     ) async {
         guard activeDownloads[zimFileID] == nil else {
             Log.DownloadService.warning(
@@ -150,7 +150,7 @@ final class DirectWriteDownloadService: NSObject, ObservableObject {
 
         let destinationURL = directory.appendingPathComponent(downloadURL.lastPathComponent)
         let validation = DownloadDestination.validateDestination(
-            directory: directory, requiredBytes: expectedSize
+            directory: directory, requiredBytes: Int64(expectedSize)
         )
 
         switch validation {
@@ -172,7 +172,7 @@ final class DirectWriteDownloadService: NSObject, ObservableObject {
             return
         }
 
-        var resumeOffset: Int64 = 0
+        var resumeOffset: UInt64 = 0
         if let state = DirectWriteDownloadState.load(for: zimFileID),
            state.validatePartialFile() {
             resumeOffset = state.bytesWritten
@@ -210,7 +210,7 @@ final class DirectWriteDownloadService: NSObject, ObservableObject {
         ).withBytesWritten(resumeOffset).save()
 
         DownloadService.shared.progress.updateFor(
-            uuid: zimFileID, downloaded: resumeOffset, total: expectedSize
+            uuid: zimFileID, downloaded: Int64(resumeOffset), total: Int64(expectedSize)
         )
 
         let fileID = zimFileID.uuidString
@@ -221,7 +221,7 @@ final class DirectWriteDownloadService: NSObject, ObservableObject {
 
         if expectedSize > 0 {
             let sizeStr = ByteCountFormatter.string(
-                fromByteCount: expectedSize, countStyle: .file
+                fromByteCount: Int64(expectedSize), countStyle: .file
             )
             Log.DownloadService.info(
                 "Download progress: 0% (Zero KB / \(sizeStr))"
@@ -275,7 +275,7 @@ final class DirectWriteDownloadService: NSObject, ObservableObject {
 
         DownloadService.shared.progress.updateFor(uuid: zimFileID, withResumeData: nil)
         DownloadService.shared.progress.updateFor(
-            uuid: zimFileID, downloaded: currentOffset, total: info.expectedTotalBytes
+            uuid: zimFileID, downloaded: Int64(currentOffset), total: Int64(info.expectedTotalBytes)
         )
 
         if !info.securityScopedAccess {
@@ -336,11 +336,11 @@ final class DirectWriteDownloadService: NSObject, ObservableObject {
 
     // MARK: - Delegate Callbacks
 
-    private func didFlush(zimFileID: UUID, bytesWritten: Int64) {
+    private func didFlush(zimFileID: UUID, bytesWritten: UInt64) {
         guard var info = activeDownloads[zimFileID] else { return }
 
         DownloadService.shared.progress.updateFor(
-            uuid: zimFileID, downloaded: bytesWritten, total: info.expectedTotalBytes
+            uuid: zimFileID, downloaded: Int64(bytesWritten), total: Int64(info.expectedTotalBytes)
         )
 
         if info.expectedTotalBytes > 0 {
@@ -350,10 +350,10 @@ final class DirectWriteDownloadService: NSObject, ObservableObject {
                 info.lastLoggedPercentage = milestone
                 activeDownloads[zimFileID] = info
                 let writtenStr = ByteCountFormatter.string(
-                    fromByteCount: bytesWritten, countStyle: .file
+                    fromByteCount: Int64(bytesWritten), countStyle: .file
                 )
                 let totalStr = ByteCountFormatter.string(
-                    fromByteCount: info.expectedTotalBytes, countStyle: .file
+                    fromByteCount: Int64(info.expectedTotalBytes), countStyle: .file
                 )
                 Log.DownloadService.info(
                     "Download progress: \(milestone)% (\(writtenStr) / \(totalStr))"
@@ -417,7 +417,7 @@ final class DirectWriteDownloadService: NSObject, ObservableObject {
 
     // MARK: - Download Lifecycle
 
-    private func startURLSessionTask(zimFileID: UUID, url: URL, fromOffset offset: Int64) {
+    private func startURLSessionTask(zimFileID: UUID, url: URL, fromOffset offset: UInt64) {
         var request = URLRequest(url: url)
         if offset > 0 {
             request.setValue("bytes=\(offset)-", forHTTPHeaderField: "Range")
@@ -578,8 +578,8 @@ final class DirectWriteDownloadService: NSObject, ObservableObject {
             let written = await sessionDelegate.fileWriter.getBytesWritten(for: zimFileID)
             let buffered = await sessionDelegate.buffer.currentSize(for: taskID)
             DownloadService.shared.progress.updateFor(
-                uuid: zimFileID, downloaded: written + Int64(buffered),
-                total: info.expectedTotalBytes
+                uuid: zimFileID, downloaded: Int64(written) + Int64(buffered),
+                total: Int64(info.expectedTotalBytes)
             )
         }
     }
@@ -643,8 +643,8 @@ final class DirectWriteDownloadService: NSObject, ObservableObject {
             activeDownloads[state.zimFileID] = info
 
             DownloadService.shared.progress.updateFor(
-                uuid: state.zimFileID, downloaded: state.bytesWritten,
-                total: state.expectedTotalBytes
+                uuid: state.zimFileID, downloaded: Int64(state.bytesWritten),
+                total: Int64(state.expectedTotalBytes)
             )
             let placeholderResumeData = Data([0x01])
             DownloadService.shared.progress.updateFor(
--- a/Model/Downloads/DirectWriteDownloadState.swift
+++ b/Model/Downloads/DirectWriteDownloadState.swift
@@ -29,10 +29,10 @@ struct DirectWriteDownloadState: Codable {
     let destinationURL: URL
     
     /// Number of bytes successfully written to disk
-    var bytesWritten: Int64
-    
+    var bytesWritten: UInt64
+
     /// Expected total size of the file (from Content-Length header)
-    let expectedTotalBytes: Int64
+    let expectedTotalBytes: UInt64
     
     /// Whether the download is currently paused
     var isPaused: Bool
@@ -60,8 +60,8 @@ struct DirectWriteDownloadState: Codable {
     }
     
     /// Remaining bytes to download
-    var remainingBytes: Int64 {
-        max(0, expectedTotalBytes - bytesWritten)
+    var remainingBytes: UInt64 {
+        bytesWritten <= expectedTotalBytes ? expectedTotalBytes - bytesWritten : 0
     }
     
     // MARK: - Initialization
@@ -70,7 +70,7 @@ struct DirectWriteDownloadState: Codable {
         zimFileID: UUID,
         downloadURL: URL,
         destinationURL: URL,
-        expectedTotalBytes: Int64
+        expectedTotalBytes: UInt64
     ) {
         self.zimFileID = zimFileID
         self.downloadURL = downloadURL
@@ -86,7 +86,7 @@ struct DirectWriteDownloadState: Codable {
     // MARK: - State Updates
     
     /// Creates a new state with updated bytes written
-    func withBytesWritten(_ bytes: Int64) -> DirectWriteDownloadState {
+    func withBytesWritten(_ bytes: UInt64) -> DirectWriteDownloadState {
         var newState = self
         newState.bytesWritten = bytes
         newState.lastWriteTime = Date()
@@ -195,7 +195,7 @@ extension DirectWriteDownloadState {
         
         do {
             let attributes = try fileManager.attributesOfItem(atPath: destinationURL.path)
-            guard let fileSize = attributes[.size] as? Int64 else {
+            guard let fileSize = attributes[.size] as? UInt64 else {
                 return false
             }
             
--- a/Model/Downloads/DirectWriteSessionDelegate.swift
+++ b/Model/Downloads/DirectWriteSessionDelegate.swift
@@ -61,10 +61,10 @@ actor DownloadDataBuffer {
 /// All disk writes are serialized through this actor.
 actor DownloadFileWriter {
     private var handles: [UUID: FileHandle] = [:]
-    private(set) var bytesWritten: [UUID: Int64] = [:]
+    private(set) var bytesWritten: [UUID: UInt64] = [:]
 
     /// Opens or creates a file for writing at the given offset.
-    func prepare(zimFileID: UUID, url: URL, offset: Int64) throws {
+    func prepare(zimFileID: UUID, url: URL, offset: UInt64) throws {
         try? handles[zimFileID]?.close()
         handles[zimFileID] = nil
 
@@ -73,7 +73,7 @@ actor DownloadFileWriter {
 
         if offset > 0 && fileManager.fileExists(atPath: path) {
             let handle = try FileHandle(forWritingTo: url)
-            try handle.seek(toOffset: UInt64(offset))
+            try handle.seek(toOffset: offset)
             handles[zimFileID] = handle
             bytesWritten[zimFileID] = offset
             Log.DownloadService.info("Opened existing file for resume at offset \(offset)")
@@ -94,12 +94,12 @@ actor DownloadFileWriter {
 
     /// Writes data to disk for a given download. Returns total bytes written.
     @discardableResult
-    func write(_ data: Data, for zimFileID: UUID) throws -> Int64 {
+    func write(_ data: Data, for zimFileID: UUID) throws -> UInt64 {
         guard let handle = handles[zimFileID], !data.isEmpty else {
             return bytesWritten[zimFileID] ?? 0
         }
         try handle.write(contentsOf: data)
-        bytesWritten[zimFileID, default: 0] += Int64(data.count)
+        bytesWritten[zimFileID, default: 0] += UInt64(data.count)
         return bytesWritten[zimFileID] ?? 0
     }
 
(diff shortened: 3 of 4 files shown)


Verryx-02:
I'm working on the CodeFactor issues, I'll push a fix shortly.

BPerlakiH:
Thank you @Verryx-02 for the PR. I think it is heading into the right direction.
There are a couple things I have noticed:
- after restarting the app, it won't automatically restart the download, it will be in "paused" state.
- there are severe hangs on the UI while downloading. I used an external SSD USB drive, which is not super fast, but still. After having a quick look in instruments, I've found this:

Screenshot 2026-02-22 at 19 22 45

Please note: we are moving towards **Swift 6**: https://github.com/kiwix/kiwix-apple/pull/1495/changes#top

Therefore, I am not sure about mixing too many concurrency solutions: Actors, GCD and locks.

I will continue reviewing this PR.

Model/Downloads/DirectWriteDownloadService.swift
```diff
@@ -0,0 +1,950 @@
...
+        guard let task = getTask(forSessionTaskID: dataTask.taskIdentifier) else { return }
+        
+        // Check if paused (thread-safe)
+        if isTaskPaused(task) { return }
+        
+        // Process data on I/O queue (no MainActor involved)
+        fileIOQueue.async { [self] in
+            // Append to buffer
+            task.dataBuffer.append(data)
+
```
BPerlakiH: I would suggest to start with something simpler. There are too many locks and hops in-between, eg:
- ``getTask(forSessionTaskID:)`` is using ``tasksLock``
- ``isTaskPaused`` is using ``tasksLock`` again
- then we hop on ``fileIOQueue`` and mutating ``task``.

I think the only thing we should do here is appending the ``data`` to something we identify by the ``dataTask.taskIdentifier``. Then maybe trigger 1 thing to process it further (somewhere else).

If the system is calling this function so many times, it might not be able to call it again, if we "spend" to much time here.

Something like:
```swift
actor DownloadBuffers {
    let buffer: [Int: Data] = [:]
}
```
and then:
```swift
func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data) {
        
        guard var buffer = downloadBuffers.buffer[dataTask.taskIdentifier] else { return }
        buffer.append(data)
```

Model/Downloads/DirectWriteDownloadService.swift
```diff
@@ -0,0 +1,950 @@
...
+    /// Protected by tasksLock
+    private var taskIDToZimFileID: [Int: UUID] = [:]
+    
+    /// Dedicated queue for file I/O operations
+    private let fileIOQueue = DispatchQueue(label: "org.kiwix.directwrite.fileio", qos: .userInitiated)
+    
+    // MARK: - Nested Types
+    
+    /// Marked as @unchecked Sendable because we manage thread safety manually via tasksLock/fileIOQueue
+    class DownloadTask: @unchecked Sendable {
```
BPerlakiH:
```swift
unchecked Sendable
```
This is definitely the wrong approach!
The problem is that there are too many values here (related to different scopes), and they are mixed into 1 place:
- related to the the ZimFile
- related to updating the progress 
- restarting the download

The key is to define which parts can operate on their own (eg: storing the incoming data, triggering progress, etc) and define the minimum amount of struct for those parts. We can communicate between those parts by using "safe" identifiers, such as the ``taskId`` or the ``zimFileID``, as those are ``@Sendable`` between actors.

Please start with defining those scopes and use an ``actor`` for each, and turn on swift(5) concurrency: "targeted" in the project settings (to check your own files):

Image

and once we have that fully working, we can think about optimizing it if needed with some non-actor solutions (if really needed).

Verryx-02:
Thank you so much for the feedback  
(It's actually one of the reasons I opened the PR).
I fully intend to see this through and I'm happy to follow your suggestions and I'll probably come back to you with some questions along the way.
I'll take a few days to go through the review properly and understand how to move forward.  
Don't worry if you don't hear from me right away

kelson42:
Although I'm glad this moves forward... I'm worried about the size of this PR! Almost 2000 lines of code for just allowing to write the ZIM file to a random directory... That sounds a lot.

Verryx-02:
@kelson42 
The size mainly comes from the fact that `URLSessionDownloadTask` always writes to a temporary file on the internal SSD before moving, and I haven't found any way to redirect it.  

The only solution I've found is a completely alternative download engine.  
The code handles the entire download lifecycle, so it's normal for it to be quite large.

That said, I'm willing to split the PR if you feel the parts should be separate PRs. I wouldn't know how to do it though, so I'd appreciate a hand.

@BPerlakiH what do you think about it?

BPerlakiH:
@Verryx-02 Please note we have updated to Swift 6, also the original download manager is now split into a smaller DownloadManager, a DownloadUI actions (alerts and like) and a DownloadDelegate. You will need to adjust your approach to that.
The most ideal solution would be to create a new DownloadDelegate for your approach, and the remaining part could be re-used.

Verryx-02:
@BPerlakiH 
I noticed. No problem. 
I'm currently learning about Swift 6.
I didn't realize how important using the Actor model was. 
I'll try to follow all the advice you gave me. 
(Thank you so much.)

Verryx-02:
# Update

@BPerlakiH 
I think I'm almost done fixing all the issues you raised. In the afternoon or maybe tomorrow I will repeat the tests to see if everything works as it should and then I'll push.  

---

> * there are severe hangs on the UI while downloading. I used an external SSD USB drive, which is not super fast, but still. After having a quick look in instruments, I've found this:

Regarding the severe hang you encountered, I think the situation is much improved now, but I still have some microhangs and occasional hangs to fix.

---

> * after restarting the app, it won't automatically restart the download, it will be in "paused" state.

I think there was a misunderstanding (?) idk.
What I meant is that after **completely closing the app**, on the next launch the download would be in a paused state.  
Anyway, it makes more sense for downloads to automatically resume on app restart, so I’ve updated the behavior.
Now downloads resume on Kiwix startup, picking up exactly where they left off.

---

Thanks for suggesting me to use "instruments". I had no idea it existed, it was very useful to me.
Any other suggestions on xcode are welcome. I don't know it that well

Model/Downloads/DirectWriteDownloadService.swift
Verryx-02: I've replaced class `DownloadTask: @unchecked Sendable` with a lightweight struct  
`DownloadTaskInfo: Sendable`  

I also split the responsibilities into separate actors:

  - `DownloadDataBuffer` actor  -> buffers incoming data, keyed by dataTask.taskIdentifier
  - `DownloadFileWriter` actor    -> serializes all disk I/O, manages file handles and bytes written
  - `@MainActor` (class-level)    -> task state (activeDownloads), UI updates, download lifecycle

Actors communicate via Sendable identifiers (taskID, zimFileID) as you suggested.
The project compiles clean with Swift 6 and no concurrency warnings. 




(I tried enabling “Complete” checking via the project.yml file, but when I later checked the project settings, it was still off. I have now activated it directly in the project settings, as you suggested.)

Verryx-02:
# As for the CodeFactor, 
I have solved some of the problems it reported. Specifically, I changed the name of variables that were too short and shortened the length of some lines. 

But I'm not sure how to solve these:
- function_body_length
- type_body_length
- file_length

After the push can you give me a hint on this?

Verryx-02:
# Testing status (macOS Tahoe 26.3.1)
## I will update this comment as I carry out the tests. 
### Tell me if you want more testing

**Test 1:  Pause and resume**
- [x] Progress resumes correctly from the right offset
- [x] ZIM file opens and works after completion
- [x] No crash 

**Test 2: Cancel mid-download**
- [x] Download stops immediately
- [x] Partial file deleted from disk
- [x] No crash 

**Test 3: Multiple simultaneous downloads (4 downloads)**
- [x] All downloads proceed without conflicts
- [x] Pausing one doesn't affect the others
- [x] ZIM files opens and works after completion
- [x] No crash

**Test 4: App restart with interrupted download** (Kiwix terminated using `command + q` during a download) 
- [x] App restarts without crash
- [x] Partial file are still on disk
- [x] Interrupted download is restored and can be resumed
- [x] ZIM file opens and works after completion
- [x] No crash

**Test 5: Wi-Fi disconnected while downloading**
- [x] Download stops immediately and resumes as soon as the Wi-Fi is turned back on.
- [x] ZIM file opens and works after completion
- [x] No crash

**Test 6: >4GB download** (4,79 GB)
- [x] Download completes without fatal errors
- [x] ZIM file opens and works after completion
- [x] No crash

**Test 7: Hangs during the download**
- [x] No severe hangs
- [x] No hangs
- [ ] No micro-hangs

**Test 8: Download on a slow usb pen drive**
- [x] No severe hangs
- [x] No hangs
- [ ] No micro-hangs
Note that the "Pause" button can take a considerable amount of time to change to "Resume."
This is an issue encountered only in this test, and I'm not sure how to fix it.

Verryx-02:
Okay, now everything should be fine in theory. This time, Git really made me sweat.

Model/Downloads/DirectWriteDownloadState.swift
```diff
@@ -0,0 +1,221 @@
...
+        guard fileManager.fileExists(atPath: destinationURL.path) else {
+            Log.DownloadService.warning(
+                "Partial file does not exist: \(destinationURL.path, privacy: .public)"
+            )
+            return false
+        }
+        
+        do {
+            let attributes = try fileManager.attributesOfItem(atPath: destinationURL.path)
+            guard let fileSize = attributes[.size] as? Int64 else {
```
BPerlakiH: This might truncate, it should be ``UInt64``, please have a look here:
https://developer.apple.com/documentation/foundation/nsdictionary/filesize()

BPerlakiH:
Thank you @Verryx-02 , I will review it further in the upcoming days.

BPerlakiH:
I had a look at this how it is working.
It still needs some major improvements to be able to qualify as a final solution.
In some cases the progress won't start, will stuck at 0%.
I also noticed that it will start the downloading into the right place by creating the ZIM file, initially zero bytes of length.
Now there's a couple of issues with this:
- I think it would be safer to create a .tmp file first, as it will be hard to make a difference if a file is really complete or not.
- If I cancel a download this ZIM file is not clean up properly

What we can do is take this PR apart a bit, and start implementing it in a slightly different way.

I think we also need a better plan for this, including scenarios, and exact behavior for  situations like, what if:
- the user starts a Download to an external drive, but than resets the settings, and starts another Download to the local drive.
- the external drive is no longer available (unmounted), during or while resuming a download
- the file (or the temporary file) already exists on the external drive

I think for the most of the above it should follow the same philosophy we used earlier:
- if the full ZIM file already exists the user should be warned
- if the temp file exists, maybe we should continue downloading it (even if we have no task for that), or maybe we should remove it and start from 0.
- the availability of the external drive, could be treated similarly to the internet access: if we have no access we put the Download automatically on "pause", inform the user with a special state, and auto resume once we gain access again.

**Overall I am getting overwhelmed by this PR.**

I have the hunch there's lots of AI generated code and comments are going on. There's too much noise of comments and explanations of obvious things, with too little direction, why something is done this way. I don't see specific parts or clear division of responsibilities.

@Verryx-02 Honestly, how much of this is AI generated?

kelson42:
> I think for the most of the above it should follow the same philosophy we used earlier:
> 
> if the full ZIM file already exists the user should be warned
> if the temp file exists, maybe we should continue downloading it (even if we have no task for that), or maybe we should remove it and start from 0.
> the availability of the external drive, could be treated similarly to the internet access: if we have no access we put the Download automatically on "pause", inform the user with a special state, and auto resume once we gain access again.

Nothing of what we have developped about user download workflow should change or been newly implemented. We should change the very minimal things.

My recommendation is to create sub-issues to our parent issues and improve tze whole issue piece by piece. @BPerlakiH Maybe you can do thatbpart to gude @Verryx-02 ?

Verryx-02:
@BPerlakiH I'll be honest with you. 
Yes, most of the code was generated with AI. 
I don't know Swift, and I said so openly in the issue before starting. I understand if this raises concerns about the quality and maintainability of the code. 

That said, I genuinely needed to download 190 GB of Wikipedia to an external SSD, so I looked for a solution because I needed it myself. 
I then decided to share it because I noticed you had been looking for a solution to the same problem for a while.  

Regarding the bugs you found: I haven't been able to reproduce either of them. For the progress stuck at 0%, it's possible that progress stays at 0% for a while if the file is very large and the connection is slow, or if the server is sending data slowly, but across all my tests I never saw it actually stuck.  
For the incomplete cleanup on cancel, I'm puzzled, because in my tests the partial file was always deleted correctly.  

That said, I can see there's still a lot of work to do, but I don't think I'm the right person to carry it forward, so I've decided to close this PR. 
I hope the approach turns out to be useful in some form. Thank you for the time you spent reviewing this.

BPerlakiH:
Thank you @Verryx-02 for your honesty. 
I think this approach is very inspiring, and there are several pieces of it we can use in our implementation.

PR #1495, opened 2026-02-22
1357 Udpate to Swift 6

### LINKED ISSUE #1357: Update to Swift 6 and Sendable
This could improve our concurrency handling and Swift 6 comes with much more restrictions of what kind of data can be sent between threads and what not (Sendable).

This is a rather large change, so needs more preparation, and can be done step by step.

Possible candidates to change:
- Core Data objects should not be sent across threads (they are based on NSObject), such as ZimFile, ZimFileMetaData, etc
- NSPredicates (core data queries) should be better secured, they can be modified on the fly, although we are not using these options, Swift 6 will throw an error on those.

BPerlakiH:
Fixes #1357

It is now ready to fully use Swift 6.
Instead of fully updating the Stripe libraries, I marked only the necessary parts to make that part work, those are not crucial parts (and we are not loosing anything by doing so).
This way we can do those as step 2, when time allows, there might be other updates to be integrated from the original Stripe library as well.
This PR is already big enough.

### DIFF
--- a/Views/Hotspot/HotspotZimFilesSelection.swift
+++ b/Views/Hotspot/HotspotZimFilesSelection.swift
@@ -120,10 +120,12 @@ struct HotspotZimFilesSelection: View {
                 } label: {
                     Text(hotspot.buttonTitle)
                         .bold()
+                    #if os(macOS)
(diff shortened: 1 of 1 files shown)


kelson42:
@BPerlakiH Thank you for the PR! Can you please open an issue per thing left to do?

PR #1501, opened 2026-02-27
Fix button text tinting on macOS 26 in Light mode

### LINKED ISSUE #1496: macOS 26 button tinting
On macOS 26 the primary action buttons within the ``ZimFileDetails`` are not tinted as before (below macOS 26) in Light-Mode:

Image
Image

They should have a white text color on the blue background (as the side menu items).

Please note: in Dark mode the colors are OK, we need to fix ONLY macOS.

william-laverty:
## Summary

Fixes #1496 — primary action buttons in `ZimFileDetail` (Open Main Page, Download) show incorrect text color on macOS 26 in Light mode when using `.borderedProminent` style.

## Problem

The `Action` view applied `.foregroundColor(nil)` to all non-destructive button text. On macOS 26, this explicit nil value interferes with `.borderedProminent`'s automatic white text styling in Light mode, causing the text to render with the default label color instead.

## Changes

- **`Action.swift`**: Only apply foreground styling for destructive actions (`.foregroundStyle(.red)`). Non-destructive buttons now inherit their text color from the button style, allowing `.borderedProminent` to correctly set white text on the blue background.
- Migrates from deprecated `foregroundColor` to `foregroundStyle`.

## Notes

Dark mode is unaffected (as noted in the issue). This is a macOS-only fix — the `.borderedProminent` style is only applied on macOS views.

### DIFF
--- a/Views/Library/Action.swift
+++ b/Views/Library/Action.swift
@@ -33,9 +33,14 @@ struct Action: View {
         AsyncButton(action: action, label: {
             HStack {
                 Spacer()
-                Text(title)
-                    .fontWeight(.medium)
-                    .foregroundColor(isDestructive ? .red : nil)
+                if isDestructive {
+                    Text(title)
+                        .fontWeight(.medium)
+                        .foregroundStyle(.red)
+                } else {
+                    Text(title)
+                        .fontWeight(.medium)
+                }
                 Spacer()
             }
         })


BPerlakiH:
Thank you @William-Laverty for fixing this. The end results is looking good to me:

Screenshot 2026-02-28 at 10 30 24
Screenshot 2026-02-28 at 10 30 17

Issue #1510, opened 2026-03-10
Search field is hidden by other, non-dynamic fields


Popolechien:
See screenshot (Kiwix 3.13; iOS 26, iPhone 12 pro): the search field is partly hidden by other buttons so that one can't really see what is being typed.

![Image](https://github.com/user-attachments/assets/baa7bda8-2b30-48ee-9d1d-b087849eefd0)

kelson42:
Looks like almost a bug here. @BPerlakiH Can younplease treat this in priority:
* Regression?
* What are the reason?
* Why we don't have seen this reported earlier? Does not seem an edge case.
* Mockup of solution

Rana-Shumail:
Hi! I'm a CS student interested in contributing to Kiwix for GSoC. I've been looking into this and I think I found the root cause in Views/SearchResults.swift.
On lines 53–57, the sidebar is coded to width: 350 for iOS. Since the iPhone 12 Pro (from the bug report) only has a width of 390pt, that fixed width is likely leaving almost no room for the search bar logic, causing the overlap with the "Cancel" and "Settings" buttons.
I see this is currently assigned to @BPerlakiH, but if they are busy or if you'd like me to test a more dynamic layout fix for compact devices, I would love to take this on!

kelson42:
@Rana-Shumail Thank you for the analysis and proposing yourself. I'm not skilled to assess what you say. But, I have co-assign you the issue. Please propose a PR and @BPerlakiH will review it.

BPerlakiH:
I do have the exact model at hand: iPhone 12 Pro

This is what I've found:
- if you search normally within the app everything is good, and behaves as it should.
- the only place it is happening, if I have the search field "focused", then I background the app, and foreground it again, it won't expand the search field fully.

Initially I do not agree with the hypothesis from above by @Rana-Shumail , if it was due to a width constraint, it would most likely apply in all cases.

This is how it looks usually:

<img width="148" height="320" alt="Image" src="https://github.com/user-attachments/assets/606e61dd-3de2-4d61-92ae-f387a8900b1a" />

after background + foreground:

<img width="148" height="320" alt="Image" src="https://github.com/user-attachments/assets/c5d9f82f-e339-4f06-9a2a-76331dc92a89" />

BPerlakiH:
@Rana-Shumail 

[...]

PR #1514, opened 2026-03-18
Fix search bar in compact mode

BPerlakiH:
Fixes: #1510 

Well, we still have some left overs of UIKit on the iPhone compact / landscape mode.
This caused us some headaches with the search bar state.

So we had the toolbar items controlled from 2 places:
- top down from UIKit's search view controller, trying to replace the items with the cancel button, when the search is active, and restoring those items, when the search is over 
- from the content view in SwiftUI bottom up, setting up other items into the toolbar

The above worked for most cases, but when we background / foreground the app, what really happens is that the UIKit is not refreshing, while the SwiftUI content does refresh, and it will set up the toolbar items as it suppose to, but we are in search mode, and the SwiftUI view has no clue about that.

# Short term solution (this PR):
- jump through some hoops, and do a bridging between UIKit and SwiftUI, and have a single source of state
- set up the cancel button of the search bar from SwiftUI, this way we have the same lifecycle for all toolbar items (they are refreshed at the same time).
- the only difference is that we cannot have the cancel button in the "accent" / blue color, but I think it was too "aggressive" anyway, so if you ask me, it's better this way:

## BEFORE

https://github.com/user-attachments/assets/d47590f2-4833-4931-9f04-da463380cdf2

## AFTER

https://github.com/user-attachments/assets/812ff8ff-eea1-4c89-9611-2cd9f3b1be7a

---------------------
# Possible long term solution:
- is to remove the older UIKit solution altogether
- this will result in less convoluted code
- the down-side is that the search field cannot be displayed in the same place. SwiftUI will place it below the toolbar, at least that is what I've found so far:
https://developer.apple.com/documentation/swiftui/searchfieldplacement
- overall Apple's recommendation is to put the search bar at the bottom of that screen... this brings in even more discussion, so let's leave that for later.

### DIFF
--- a/App/CompactViewController.swift
+++ b/App/CompactViewController.swift
@@ -24,6 +24,13 @@ import UIKit
 import CoreData
 import Defaults
 
+/// Bridge between UIKit and SwiftUI search state and dismiss
+private final class SearchUpdate: ObservableObject {
+    @Published var isSearching: Bool = false
+    var dismiss: (() -> Void)?
+}
+
+// iPhone portrait only
 final class CompactViewController: UIHostingController<AnyView>, UISearchControllerDelegate, UISearchResultsUpdating {
     private let searchViewModel: SearchViewModel
     private let searchController: UISearchController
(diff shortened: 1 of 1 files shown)


kelson42:
@BPerlakiH Please open issue and remove UIkit straight away

BPerlakiH:
> @BPerlakiH Please open issue and remove UIkit straight away

https://github.com/kiwix/kiwix-apple/issues/1515

PR #1523, opened 2026-03-28
[maps] Fix map not displayed full screen on initial load

### LINKED ISSUE #1521: [maps] - map is not displayed "full screen"
This issue is about ZIMs from new maps scraper.

Sample test ZIM: https://mirror.download.kiwix.org/zim/.hidden/dev/maps_en_monaco_2026-03.zim

While I open the issue in kiwix-apple repo, I'm open to consider we need to modify scraper code to do things in a slightly different manner which will work in kiwix-apple

This issue is about the fact than we I open the ZIM in kiwix-apple, the map is not displayed "full-screen".

When I resize window, it works as expected.

https://github.com/user-attachments/assets/aa50d735-02a0-48e7-81db-fdf74436f382

Amith71965:
## Problem                                                                      
  Fixes #1521     
                                                                                            
  On macOS, MapLibre GL JS initialises its WebGL canvas while the WKWebView frame is still settling, resulting in a canvas smaller than the map container (~196pt instead of 500pt). MapLibre's ResizeObserver skips its first callback by design, so when the WKWebView grows to its final size the canvas is never corrected.                                          
                                                  
  ## Root cause
  - WKWebView frame transitions through an intermediate size during SwiftUI layout
  - MapLibre reads _`clientHeight`_ at that intermediate size and initialises the canvas accordingly                                                                               
  - The 196pt→500pt growth triggers the ResizeObserver skip-first callback, consuming it    
  - No further size change occurs, leaving the canvas at the wrong size                     
                                                                                            
  ## Fix                                                                                    
  After _`didFinish`_, if the map canvas is shorter than its container, bump the container height by 1pt and restore it after 200ms. This fires a second ResizeObserver callback (the one that actually calls _`map.resize()`_) at the correct final height.                   
                                                                         
  ## Testing                                                                                
  Tested with _`maps_en_monaco_2026-03.zim`_ on macOS — map now displays full screen on initial load without requiring window resize.

https://github.com/user-attachments/assets/6769714e-f0ec-40ea-9919-b73312651788

### DIFF
--- a/ViewModel/BrowserViewModel.swift
+++ b/ViewModel/BrowserViewModel.swift
@@ -513,7 +513,7 @@ import CoreKiwix
         webView.adjustTextSize()
 #else
         Task { await persistState() }
-        let js = """
+        let mapResizeScript = """
             (function() {
                 var el = document.getElementById('map');
                 var canvas = el && el.querySelector('canvas');
@@ -522,7 +522,7 @@ import CoreKiwix
                 setTimeout(function() { el.style.removeProperty('height'); }, 200);
             })();
             """
-        webView.evaluateJavaScript(js, completionHandler: nil)
+        webView.evaluateJavaScript(mapResizeScript, completionHandler: nil)
 #endif
     }
 


rgaudin:
@Amith71965 thank you but we don't want per-ZIM hacks on the readers

Amith71965:
@rgaudin 

Thanks for the feedback! Just to clarify — is the concern about targeting DOM elements like _`map`_ specifically. Asking because the issue only manifests with map ZIMs due to how MapLibre initialises its canvas during WKWebView layout.

Let me know if the team is thinking to resolve in other way, or maybe your expectations.

rgaudin:
Seems like https://github.com/kiwix/kiwix-apple/pull/1516 might have fixed it already

benoit74:
I confirm we do not need that, looks like issue is now solved in `main` branch

PR #1532, opened 2026-04-05
Move diagnostics start task to async

### LINKED ISSUE #1531: Speed up app start, fix diagnosctics thread
In our former release 3.13 the app start time was highlighted as being slower than before, due to a diagnostic start call on the UI thread:

Image

After verifying it, it is indeed checking the available free space on the device for diagnostics on the main thread:

Image

It should be moved off the UI thread.

BPerlakiH:
Fixes: #1531

Now the initial diagnostics calls are running off the UI thread:

Screenshot 2026-04-05 at 21 11 19

### DIFF
--- a/App/App_iOS.swift
+++ b/App/App_iOS.swift
@@ -30,7 +30,7 @@ struct Kiwix: App {
     @StateObject private var colorSchemeStore = UserColorSchemeStore()
     
     init() {
-        Task { @MainActor in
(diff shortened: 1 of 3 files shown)


kelson42:
I merge, but here the naming of things is incorrect. No reason to bound this freespace method to the diagnostic, or no reason to start it at that time.

PR #1535, opened 2026-04-06
libkiwix-14.2.0-1 (with libzim 9.6.0)

kelson42:
With essential fix https://github.com/openzim/libzim/pull/1053 for macOS


BPerlakiH (approved):
Looks good to me on macOS and on iOS.

PR #1540, opened 2026-04-07
Fix macOS Apple Pay compatibility

boxabirds:
Crashed on my Intel Mac: fixes a macOS launch crash caused by the SwiftUI Apple Pay API on some macOS runtimes.

  The failing path used `PayWithApplePayButton` / `PayWithApplePayButtonLabel` directly on macOS. On at least macOS 14.6.1, that can abort at launch with a missing `_PassKit_SwiftUI` symbol before the app becomes usable.

  This PR keeps donations enabled on macOS, but switches the macOS implementation to native PassKit APIs instead of the SwiftUI Apple Pay view.

### DIFF
--- a/Model/Payment.swift
+++ b/Model/Payment.swift
@@ -122,23 +122,22 @@ struct Payment {
     /// - Returns: Setup button if no cards added yet,
     /// nil if Apple Pay is not supported
     /// or donation button, if all is OK
-    static private func paymentButtonType() -> ApplePaymentLabelType? {
+    static private func paymentButtonType() -> ButtonLabelType? {
         // only kiwix app is supporting donations atm.
         guard case .kiwix = AppType.current else { return nil }
-        
         if PKPaymentAuthorizationController.canMakePayments() {
-            return ApplePaymentLabelType.donate
+            return ButtonLabelType.donate
         }
         if PKPaymentAuthorizationController.canMakePayments(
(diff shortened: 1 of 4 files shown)


Model/Payment.swift
```diff
@@ -122,23 +122,22 @@ struct Payment {
...
-            return ApplePaymentLabelType.setUp
+            return ButtonLabelType.setUp
         }
         return nil
     }
     
-    /// Sendable version of PayWithApplePayButtonLabel
-    private enum ApplePaymentLabelType: Sendable {
+    /// Sendable app-side representation of the Apple Pay button label.
+    enum ButtonLabelType: Sendable {
```
BPerlakiH, line 140: This was already a custom type, what is the benefit of renaming it ?

BPerlakiH:
I will need to dig deeper into this. For me it is unclear yet, what real the source of the problem is, and why is it crashing...
Btw:
[PayWithApplePayButtonLabel](https://developer.apple.com/documentation/passkit/paywithapplepaybuttonlabel) is supported from macOS 13, so it certainly not a "compatibility" issue as described.

Are there any proofs that it is indeed not working, maybe crash logs could help to identify the problem.

kelson42:
I will close this PR to allow to start at the beginning, this means a proper issue with proper reproduction case.

PR #1541, opened 2026-04-07
Pr/macos direct downloads

boxabirds:
I want to download Kiwix files to an external disk and the files are bigger than the space available on my system drive. 

So this change adds a macOS download preference with Chrome-style behavior:

  - new preference: `Always ask where to save files`
  - default is `on`
  - when `off`, downloads use a saved default folder
  - when `on`, the user is prompted for a folder before each download

  This also changes macOS download transport behavior:

  - downloads now stream directly to the selected destination volume
  - they no longer download to the system volume first and move afterward
  - free-space checks now use the actual destination disk

  macOS only. iOS behavior is unchanged.


BPerlakiH:
@boxabirds Thank you for your PR. 
I must say I am a little confused. In the other issue it seemed you could not even launch the macOS app: https://github.com/kiwix/kiwix-apple/pull/1540, and here you have a PR with only 3 commits for a topic that is size XL to say the least.
For me it looks like AI generated code to be honest. Note you are not the first one, see https://github.com/kiwix/kiwix-apple/pull/1494

At this stage, I won't spend time reviewing it. If you really wrote it yourself, I am happy to re-open it and have a meaningful discussion about it.
In the meantime I recommend reading the following interesting article on this very topic to understand my point of view as well:
https://observablehq.com/blog/the-medium-is-the-problem

boxabirds:
Sure it’s AI generated. It doesn’t mean I didn’t put a lot of thought and
care into it validating and testing it with appropriate test coverage

I’m just solving a problem that I had and I tried to follow contribution
guides and I didn’t see anything untoward.

The crash in the other poor request is confusing to me because it worked
locally. It looks like copilot added the crash.

I think other people could benefit from being able to choose the download
folder

But I’m okay just to use my fork it doesn’t really bother me.

kelson42:
@boxabirds In general, we don't welcome AI generated code.

rgaudin:
> @boxabirds In general, we don't welcome AI generated code.

This is both ill-defined and buried in a comment in a closed PR.

Given we have and advertise CONTRIBUTING rules, those should be updated. Despite being myself against all AI-generated contributions to Kiwix, I find it very rude to not specify it in our contributing rules and refuse complying contributions. It's not like it's a new or unexpected topic.

Issue #1520, opened 2026-03-25
[maps] - search does not move to expect location


benoit74:
This issue is about ZIMs from new maps scraper.

Sample test ZIM: https://mirror.download.kiwix.org/zim/.hidden/dev/maps_en_monaco_2026-03.zim

While I open the issue in kiwix-apple repo, I'm open to consider we need to modify scraper code to do things in a slightly different manner which will work in kiwix-apple

## Expected behavior

ZIM works as expected on kiwix-serve, visible at https://dev.library.kiwix.org/viewer#maps_en_monaco_2026-03/

https://github.com/user-attachments/assets/f815d925-c9bf-4a84-b104-9efa3eff6e10

## How it works

- Search (title or full-text) entry is at `search/Commune%20de%20Monaco` in the ZIM (see https://dev.library.kiwix.org/content/maps_en_monaco_2026-03/search/Commune%20de%20Monaco)
- This entry contains an HTML redirect with `    <meta http-equiv="refresh" content="0;URL='../index.html#lat=43.73216&lon=7.41981&zoom=6'" />`
- On `index.html` page, JS is triggered which detects the fragment and move the map to appropriate lat/lon/zoom

## What happens in kiwix-apple

Nothing, the map flickers but that's all

kelson42:
@BPerlakiH We are at the final stage of our new map ZIM. Please support @benoit74 in priority over your usual dev tasks so we can achieve to release the version 1.0 of these maps in a reasonable time. More info at https://github.com/openzim/maps

BPerlakiH:
@benoit74 @kelson42 After some hard digging into, I've found what the is causing this issue.

The requests that we really make on the Apple side to the content are through the "zim" scheme, such as:
"zim://47e76585-683d-e4c9-55bd-eab2fe6f7f86/planet"

Within this specific content, there's a JS library "Axios", that is too restrictive / trying to be too smart, and throws the following warning:

<img width="348" height="235" alt="Image" src="https://github.com/user-attachments/assets/37fb2eec-ffbb-45ad-bcf3-4f251effd7eb" />

This is because it is restricting the protocols to be:
```js
protocols: ["http", "https", "file", "blob", "url", "data"]
```

Which is in the file:
```
zim://47E76585-683D-E4C9-55BD-EAB2FE6F7F86/assets/index-iVlzh6x0.js
```

according to map sources, it is coming from:
```
[...]

PR #1554, opened 2026-04-22
Bridge HTML5 Geolocation API to CoreLocation in WebKit viewer

### LINKED ISSUE #1560: Allow location sharing
Allow ZIM files to use device location.
Bridge the web request to native location handling.
Related base PR: https://github.com/kiwix/kiwix-apple/pull/1554

jasontitus:
Fixes #1560

## Summary

Map ZIM files (e.g. the new `maps_*` scraper output, WikiVoyage) call `navigator.geolocation` to recenter on the user. WebKit on iOS/macOS does not surface a permission prompt for geolocation in app-hosted `WKWebView`s, so the call silently fails and the map sits idle. This PR bridges the HTML5 Geolocation API to CoreLocation through a small JS shim and a `GeolocationService`.

Related: #1520, #1521 (recent work on map ZIM behaviour in kiwix-apple).

## Problem

`navigator.geolocation.getCurrentPosition` / `watchPosition` in a `WKWebView` rely on `WKUIDelegate`'s implicit prompt path, which Apple ships only in Safari, not in app-hosted web views. The call returns no permission prompt and no result. Map ZIMs that try to recenter on the user therefore appear broken.

## Changes

- **`Model/Utilities/GeolocationService.swift` (new)** — wraps `CLLocationManager`. `@MainActor`-isolated; CoreLocation delegate callbacks land on a small nonisolated `CLLocationManagerDelegate` shim that hops to MainActor. Supports both one-shot (`requestLocation`) and continuous (`startWatching` / `stopWatching`) requests. Authorization is requested lazily on the first call.
- **`Support/injection.js`** — IIFE that installs a `navigator.geolocation` shim. Posts to a new `geolocation` script message handler. Falls back to defining on `Object.getPrototypeOf(navigator)` if the instance property is non-configurable, and surfaces `window.__kiwixGeolocationShimInstalled` so installation can be verified from the Web Inspector.
- **`ViewModel/BrowserViewModel.swift`** — registers the `geolocation` script message handler, lazily creates one `GeolocationService` per tab on first request, and replies through `callAsyncJavaScript` (no JSON-into-source interpolation, so U+2028 / U+2029 in error messages can't break the bridge). Stops active watches on `didCommit` and on tab teardown.
- **`project.yml`** — adds `INFOPLIST_KEY_NSLocationWhenInUseUsageDescription`, `INFOPLIST_KEY_NSLocationUsageDescription`, and the `com.apple.security.personal-information.location` sandbox entitlement. The CI step that strips `in-app-payments` for macOS unit tests is unaffected — the new entitlement sits alongside it in the same `targetTemplate`.
- **`Tests/GeolocationServiceTests.swift` (new)** — covers `LocationSnapshot` field handling (negative-sentinel values become `nil`) and the W3C `PositionError` code mapping the JS shim relies on (`PERMISSION_DENIED == 1`).

## Notes

- **Scope.** This wires geolocation only. No UI, no settings toggle. The first call surfaces the standard system permission prompt; if the user denies, the JS receives `PositionError { code: 1 }` and the watch is dropped from the JS pending map so it cannot leak.
- **Authorization model.** `requestWhenInUseAuthorization()` on both iOS and macOS — `Always` would be overreach for a page-driven prompt. macOS supports `WhenInUse` since 11.0.
- **Localization.** Usage-description strings are added as `INFOPLIST_KEY_*` build settings, matching the existing pattern for `NSPhotoLibraryUsageDescription`. Happy to route them through `localizations.py` if maintainers prefer.
- **No changes to** existing handlers, scheme handlers, Apple Pay entitlement, or the `headings` script message handler. The pre-existing `removeScriptMessageHandler(forName:"headings")` pattern is mirrored for the new handler so re-init on the same `WKWebViewConfiguration` is safe.
- **Tested locally** with iOS Simulator (Debug, iPhone 17 Pro), a physical iPhone 17 Pro Max device install, and macOS arm64 Debug build. Unit tests pass. `python localizations.py validate` passes.


kelson42:
@jasontitus Thank you for your PR. I don't think we have ZIM files relying on geolocalisation of the client device... or maybe one of the ZIM file you produce? Do you have a small demo ZIM file?
@benoit74 Coukd younplease confirm we have no issue open related to such a use case?

jasontitus:
I am in the process of evolving a whole bunch of them, but you kind of need to have one for your location to get a sense of it.

Check out https://streetzim.web.app/ for the set of map zims I am building out.  But only some of them have the routing and exploring UI baked into them.   Silicon Valley already works, California and Japan will be uploaded in an hour or two, and Australia/NZ in 5-6 hours.  Then overnight I will upload new US East Coast, US West Coast, and Europe.  More coming after that.

They all work fine in the PWA version and with these changes work fine with the iOS app.

Example video of using geolocation in Kiwix on one of these mapzims - 
https://archive.org/download/streetzim-washington-dc/osm-washington-dc-2026-04-28.zim

kelson42:
@jasontitus Can you please just provide a clear step by step reproduction case with one of your small ZIM files? Ideally in a separate issue? I really want to avoid the dev to try to long figure out how/when the bug surfaces.

EDIT: the only thing I see is the button (in blue in the screenshot) at the bottom right corner. Do you confirm?
image

jasontitus:
Yes - iterating on the automated feedback and will give a before and after screencast.  For it to be a small example zim file for a tester though, I need to know roughly where the tester will be located.  I can generate a state or country file for them to see what it looks like.  The location find doesn't do anything if your location isn't within the bounds of the zim map.  I can also add a message to that effect as well in the web UI.

BPerlakiH:
@jasontitus Thank you for the PR. I do have a couple of comments on this.
In order to validate this from end to end, please point to a small ZIM file we can use, it can be any place. Don't worry about the physical location, the iOS simulator can be set to any coordinates we want.

I do have some concerns though, so I am trying to understand the use-case here as there are requests for:
- current location
- watching the location

Plus from what I can see there might be multiple location requests (each with a different id) feeding back to the same ZIM file (via JS). Why is that needed?  The user will have a single location at any given time.

Also some technical things to point out, the MainActor usage is a bit convoluted, I will need to dig deeper into that.
We also use translations via LocalString, those in-lined strings also need some updates in that regard. 

@kelson42 @benoit74 Did you see this claim, about the required Patched libzim here:
https://github.com/jasontitus/streetzim#prerequisites

jasontitus:
Yes - the streetzim project exposed  some issues with libzim and I submitted a PR  ( https://github.com/openzim/libzim/pull/1053 )  which was  accepted  into  9.6.0.  I  have updated the README  on streetzim  to recognize that.

I have updated to address feedback and bugs found in  further review.

Here is a small zim (63MB) of Washington, DC that has can be used to test the location features -  [https://archive.org/download/streetzim-washington-dc/osm-washington-dc-2026-04-28.zim](url)

jasontitus:
> @jasontitus Thank you for the PR. I do have a couple of comments on this. In order to validate this from end to end, please point to a small ZIM file we can use, it can be any place. Don't worry about the physical location, the iOS simulator can be set to any coordinates we want.
> 
> I do have some concerns though, so I am trying to understand the use-case here as there are requests for:
> 
> * current location

"Show me  where I am on the map"

> * watching the location
> 

During driving directions you can leave  Kiwix open and it shows you moving  along the map.

kelson42:
> @kelson42 @benoit74 Did you see this claim, about the required Patched libzim

@BPerlakiH Yes. All of this is well known indeed!

BPerlakiH:
Overall I think it is a good use-case, I just want to make sure we can standardize it, so it could be beneficial for other ZIM files as well.

@jasontitus Can you please explain this part?

> Plus from what I can see there might be multiple location requests (each with a different id) feeding back to the same ZIM file (via JS). **Why is that needed?** The user will have a single location at any given time.

BPerlakiH:
I did try it out, it works, but the additional id param is not needed.
Currently it is doing the following:
- get location for id = 1
- stop location for id = 1
- get location for id = 2
- stop location for id = 2
and so on...

The original Geolocation API also not using the id value as the input param:
https://www.w3.org/TR/geolocation/#get-current-position

I did test a version of this without the id param (both on swift and JS side), and it is working equally fine without it.

@kelson42 this is how it looks with the driving simulation (from Apple):

https://github.com/user-attachments/assets/98807cff-873d-4df7-8fb5-f0e3863aa2f5

@kelson42 I can clean these changes up, as it's very much bloated, there are extra things here, which are not needed.

kelson42:
@BPerlakiH yes, please revamp in an other PR of you.

BPerlakiH:
@jasontitus Also there are some JS console log errors, that is worth to have a look into from the ZIM content side. If you connect Safari debug view to the app, you can see them:

Screenshot 2026-05-01 at 08 09 14
Screenshot 2026-05-01 at 08 08 42
Screenshot 2026-05-01 at 08 08 28

kelson42:
Superseeded by #1561

BPerlakiH:
@jasontitus There's a new simplified PR for this here: 
https://github.com/kiwix/kiwix-apple/pull/1562

Please also check the routing on these ZIM files, sometimes the route calculation is taking a really long time, like here:

Simulator Screenshot - iPhone 17 Pro Max 6 9  - 2026-05-03 at 17 15 49

If the current location and the destination are close to each other, it usually works fine.

If those are somewhat distant - like in the example - it takes forever to find the route, despite that the current location is immediately supplied. 
On the screenshot, I waited over 50 sec, and still no route results...

jasontitus:
Yes - I have plans to optimize the routing but first I just wanted to get offline routing working at all.  First I am getting the search, routing, and local finding all partitioned so they can work on US or Europe scale maps without running out of memory on phones.  Then I can rework routing to be faster.  Thanks for the help!

PR #1561, opened 2026-05-01
Allow location sharing

### LINKED ISSUE #1560: Allow location sharing

BPerlakiH:
Fixes: #1560 

Superseeds #1554 

@kelson42 you were right!
It's enough to just enable the location services and it works out of the box:

## Simulated driving in California, with request prompt:

https://github.com/user-attachments/assets/cc31b0d6-f4e8-4997-b95b-04fbabedc8be

-------

## Disabling location, gives a crossed out location icon:
no_location

### DIFF
--- a/project.yml
+++ b/project.yml
@@ -24,6 +24,8 @@ settings:
     INFOPLIST_KEY_UISupportedInterfaceOrientations_iPhone: UIInterfaceOrientationPortrait UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight
     INFOPLIST_KEY_UISupportedInterfaceOrientations_iPad: UIInterfaceOrientationPortrait UIInterfaceOrientationPortraitUpsideDown UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight
     INFOPLIST_KEY_NSPhotoLibraryUsageDescription: "Kiwix needs permission to saves images to your photos app."
+    INFOPLIST_KEY_NSLocationWhenInUseUsageDescription: "Kiwix can share your location with map content from ZIM files you open."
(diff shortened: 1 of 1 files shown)


BPerlakiH:
FYA: @jasontitus

project.yml
BPerlakiH, line 1: 3 lines of code vs [781 lines](https://github.com/kiwix/kiwix-apple/pull/1554) of code ;)

kelson42:
@BPerlakiH Works fine on macOS too?

BPerlakiH:
> @BPerlakiH Works fine on macOS too?

Not yet ... trying to figure it out now.

BPerlakiH:
@kelson42 
Well.. it's not working for macOS out of the box.
There's also another issue on iOS, there are 2 confirmation pop-ups to get the location, which is a bit annoying...

We can do 2 things:
- either go with the full solution
- do the workaround only for macOS. We can even simplify it to a single location request,  instead of continued location watching, in the hope that no-one will drive / travel around with their mac open. That means that iOS users need to confirm 2 popups.

It seems we are not the first ones, who bumped into this issue, see:
https://stackoverflow.com/questions/39665367/how-to-prevent-wkwebview-to-repeatedly-ask-for-permission-to-access-location

kelson42:
Do the full solution and document well in the code why this is complicated like this

jasontitus:
Thanks - sorry if there was too much in the previous PR.  I will work on the JS errors - I have cleared them out before on Chrome (for the PWA) but not on Safari since it seemed to be working fine w/ the PR

BPerlakiH:
replaced by: #1562

PR #1579, opened 2026-05-16
1032 recurring payments with apple pay email

### LINKED ISSUE #1032: Implement Apple Pay recurring payment for iOS

### LINKED ISSUE #1580: Identify clearly donation as coming from Kiwix Apple
For the moment we are not able to list precisely the donations coming from Kiwix Android.

We should be able to identify clearly in our backend the donations coming from Kiwix Apple.

BPerlakiH:
Fixes: #1032 
Fixes: #1580

### Adding the following to the payment requests:
- required email address via Apple Pay
- currently preferred user language
- current origin of the payment (mac, iphone, ipad)

### DIFF
--- a/Model/Payment.swift
+++ b/Model/Payment.swift
@@ -175,6 +175,7 @@ struct Payment {
         return request
     }
 
+    // swiftlint:disable:next function_body_length
     func onPaymentAuthPhase(selectedAmount: SelectedAmount,
                             phase: PayWithApplePayButtonPaymentAuthorizationPhase) {
         switch phase {
--- a/Model/Payment.swift
+++ b/Model/Payment.swift
@@ -67,7 +67,7 @@ struct Payment {
     // optional to be implemented:
     // for reference see:
     // https://developer.apple.com/documentation/merchanttokennotificationservices
-    static let paymentSubscriptionTokenCallbackURL = URL(string: "http://api.donation.kiwix.org/v1/stripe/token-callback")!
+    static let subscriptionTokenCallbackURL = URL(string: "http://api.donation.kiwix.org/v1/stripe/token-callback")!
     static let supportedNetworks: [PKPaymentNetwork] = [
         .amex,
         PKPaymentNetwork.pagoBancomat,
@@ -253,7 +253,7 @@ struct Payment {
             managementURL: URL(string: Self.paymentSubscriptionManagingURL)!
         )
         payRequest.regularBilling.intervalUnit = .month
-        payRequest.tokenNotificationURL = Self.paymentSubscriptionTokenCallbackURL
+        payRequest.tokenNotificationURL = Self.subscriptionTokenCallbackURL
         return payRequest
     }
 
--- a/Model/Payment.swift
+++ b/Model/Payment.swift
@@ -207,7 +207,11 @@ struct Payment {
                                                    returnURLPath: nil,
                                                    usingClientSecretProvider: { @Sendable in
                     await StripeKiwix
-                        .clientSecretForPayment(endPoint: endPoint, selectedAmount: selectedAmount, email: email)
+                        .clientSecretForPayment(endPoint: endPoint,
+                                                selectedAmount: selectedAmount,
+                                                email: email,
+                                                deviceName: Device.current.rawValue
+                        )
                 }, withAPI: StripeAsyncAPI())
 
                 switch result.status {
--- a/Model/Payment.swift
+++ b/Model/Payment.swift
@@ -190,6 +190,9 @@ struct Payment {
             Log.Payment.info("onPaymentAuthPhase: .didAuthorize")
             // call our server to get payment / setup intent and return the client.secret
             Task { @MainActor [resultHandler] in
+                let emailAddress = payment.shippingContact!.emailAddress!
+                Log.Payment.info("Email? \(emailAddress)")
+                // let paymentServer = StripeKiwix(endPoint: URL(string: "http://localhost:8000/v1/stripe")!)
                 let paymentServer = StripeKiwix(endPoint: Self.kiwixPaymentServer)
                 do {
                     let publicKey = try await paymentServer.publishableKey()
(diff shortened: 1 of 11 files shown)


Model/Payment.swift
```diff
@@ -247,6 +236,41 @@ struct Payment {
...
+            regularBilling: .init(
+                label: LocalString.payment_monthly_support_label,
+                amount: NSDecimalNumber(value: selectedAmount.value),
+                type: .final
+            ),
+            managementURL: URL(string: Self.paymentSubscriptionManagingURL)!
+        )
+        payRequest.regularBilling.intervalUnit = .month
+        // not yet sure how this is used, or if needed
+//        payRequest.tokenNotificationURL = Self.paymentSubscriptionTokenCallbackURL
```
BPerlakiH: @rgaudin I wasn't sure about this one if this is needed or not.

Model/Payment.swift
```diff
@@ -64,6 +64,8 @@ struct Payment {
     static let merchantSessionURL = URL(string: "https://apple-pay-gateway.apple.com" )!
     static let merchantId = "merchant.org.kiwix.apple"
     static let paymentSubscriptionManagingURL = "https://www.kiwix.org"
```
BPerlakiH, line 61: @rgaudin we should probably change this to a real subscription managing URL.

Model/StripeKiwix.swift
```diff
@@ -46,9 +46,8 @@ struct StripeKiwix {
         selectedAmount: SelectedAmount
     ) async -> Result<String, Error> {
         do {
-            // for monthly we should create a setup-intent:
-            // see: https://github.com/kiwix/kiwix-apple/issues/1032
-            var request = URLRequest(url: endPoint.appending(path: "payment-intent"))
+            let requestPath = selectedAmount.isMonthly ? "setup-intent" : "payment-intent"
```
BPerlakiH, line 56: @rgaudin This is the part where the monthly subscription is calling the "/setup-intent" endpoint instead of the "/payment-intent" one.

Model/Payment.swift
```diff
@@ -64,6 +64,10 @@ struct Payment {
...
+    // optional to be implemented:
+    // for reference see:
+    // https://developer.apple.com/documentation/merchanttokennotificationservices
+    static let paymentSubscriptionTokenCallbackURL = URL(string: "http://api.donation.kiwix.org/v1/stripe/token-callback")!
```
BPerlakiH: @rgaudin I added an example URL for this, I think it could be beneficial to implement this part as well, see the docs in the link above. 
@kelson42 On a pure MVP level, we can launch recurring payments without this being implemented, and can add it later.

rgaudin:
Thank you @BPerlakiH ; those missing URLs will have to wait for server-side implementation

Model/StripeKiwix.swift
```diff
@@ -104,15 +119,24 @@ private struct ClientSecretKey: Decodable {
 private struct SelectedPaymentAmount: Encodable {
     let amount: Int
     let currency: String
+    let email: String
+    let device: String
+    #if os(macOS)
+    let locale: String = NSLocale.preferredLanguages.first ?? ""
+    #else
+    let locale: String = Locale.preferredLanguages.first ?? ""
+    #endif
```
BPerlakiH: with these properties in place, we are also solving: https://github.com/kiwix/kiwix-apple/issues/1580

kelson42: please update PR description accordingly. I'm not sure to understand how it works either, do you pass parameter to the domation portal? Where is the parameter about the periodicity?

BPerlakiH: We have 2 endpoints:
- /setup-intent: this is for monthly subscriptions
- /payment-intent: this is for one time donations

We use the same set of parameters for both, so the email + origin + lang is added  to both one-time donations and monthly subscriptions.

Model/Payment.swift
```diff
@@ -165,36 +163,27 @@ struct Payment {
...
         return request
     }
 
+    // swiftlint:disable:next function_body_length
     func onPaymentAuthPhase(selectedAmount: SelectedAmount,
                             phase: PayWithApplePayButtonPaymentAuthorizationPhase) {
         switch phase {
         case .willAuthorize:
             Log.Payment.info("onPaymentAuthPhase: .willAuthorize")
+        // Important! do not attempt to do anything UI related after
```
rgaudin, line 184: Good to know!

Model/Payment.swift
```diff
@@ -204,36 +193,48 @@ struct Payment {
...
                 }
-                // we should update the return path for confirmations
-                // see: https://github.com/kiwix/kiwix-apple/issues/1032
+                // Note: the returnURL path is not needed, we are never leaving the app. According to stripe docs:
+                // "The URL to redirect your customer back to after they authenticate or cancel
+                // their payment on the payment method’s app or site.
+                // This should probably be a URL that opens your iOS app."
                 let stripe = StripeApplePaySimple()
                 let endPoint = paymentServer.endPoint
+                let email = payment.shippingContact?.emailAddress ?? ""
```
rgaudin, line 206: Is this on purpose? A missing/empty/non-valid-looking email will fail the server call (HTTP 400). Is this handled/explained to the user?

rgaudin (approved):
Looks good ; please see inline comment about email address.

I noticed in the ApplePay confirmation sheet that it contains an icon of a laptop (because I'm testing on Mac I believe) with a big K for single donation and a big M for monthly.
It looks like it's taking the letter from the first letter in the description. We might want to craft the message in a way that it's consistent.
Do you know if this is customizable with an actual icon?

Model/Payment.swift
BPerlakiH: Yes this is on purpose.

What we do is require the email as part of the payment process, so from the UI point of view **the user cannot proceed further without providing a valid email address, and the user cannot leave it empty**.
On the object level it's defined as optional though. We can very safely assume that a valid email value is always going to be there.

For the mostly theoretical, and rather impossible scenario of not being set, it's OK to fall back to an empty string and get a server error. All HTTP error  status codes will trigger a visible payment error for the user.

The only other option (not recommended) is to do a force-unwrap (payment.shippingContact!.emailAddress!), but that will crash the app, if it's nil.

BPerlakiH:
> I noticed in the ApplePay confirmation sheet that it contains an icon of a laptop (because I'm testing on Mac I believe) with a big K for single donation and a big M for monthly.
It looks like it's taking the letter from the first letter in the description. We might want to craft the message in a way that it's consistent.
Do you know if this is customizable with an actual icon?

I am not sure, which part is that you are referring to. My suspicion is that it might be related to the user icon associated with the email...
I do have the following on my side, is it this screen? 

Screenshot 2026-05-27 at 00 35 11 2

rgaudin:
> I do have the following on my side, is it this screen?

I have the same on my mac but it then asks me to confirm on my phone. I was talking about the phone sheet.

BPerlakiH:
> I have the same on my mac but it then asks me to confirm on my phone. I was talking about the phone sheet.

Interesting...
I don't have the phone confirmation step at all, maybe because of having touch ID on my mac.

PR #1587, opened 2026-05-24
BrowserTab and content preloading for quicker switching

### LINKED ISSUE #1479: Switching tab is not smooth
See here:

https://github.com/user-attachments/assets/7f4aa977-65b5-4cda-b774-6ac39832a8c3

Can we not find a solution to make it more comfortable for the eyes?

BPerlakiH:
Fixes: #1479

The iPad version of the app allows to have multiple tabs open,
and change between them with 1 tap.
We are talking about the scenario, when the application and the initially selected tab is already loaded.
Now, if we switch from 1 tab to another, the first time we did that it wasn't smooth.
Many steps had to be taken to load another tab:
- if the other tab comes from another ZIM file, the ZIM file had to be opened
- the browser instance related to that tab had to be created
- the content of that page had to be load

Solution:
Preload browser tabs and ZIM files for the inactive (not currently selected) tabs,
but only do this on iPad, and only after the first tab has been loaded,
in order not to make the start up time worse.

## BEFORE

https://github.com/user-attachments/assets/ce28aae5-05b5-42bb-9c52-2cc7619cfe06

## AFTER

https://github.com/user-attachments/assets/55d04e7b-51b3-4572-ae10-e5e106f4eaa7

### DIFF
--- a/App/CompactViewController.swift
+++ b/App/CompactViewController.swift
@@ -13,16 +13,12 @@
 // You should have received a copy of the GNU General Public License
 // along with Kiwix; If not, see https://www.gnu.org/licenses/.
 
-//
-//  CompactViewController.swift
-//  Kiwix
-
 #if os(iOS)
 import Combine
-import SwiftUI
-import UIKit
 import CoreData
 import Defaults
+import SwiftUI
+import UIKit
 
 // iPhone portrait only
 struct CompactTabView: View {
--- a/App/SplitViewForiPad.swift
+++ b/App/SplitViewForiPad.swift
@@ -123,6 +123,9 @@ struct SplitViewForiPad: View { // swiftlint:disable:this type_body_length
             if let currentItem = navigation.currentItem {
                 selection = MenuItem(from: currentItem)
             }
+            if case let .tab(selectedTabId) = selection {
+                BrowserTabPreloader.shared.start(with: tabs, selectedTabId: selectedTabId)
+            }
         }
         .onChange(of: navigation.currentItem) { _, newValue in
             updateSelection(newValue)
@@ -290,10 +293,7 @@ struct SplitViewForiPad: View { // swiftlint:disable:this type_body_length
         if let newNavItem, let newSelection = MenuItem(from: newNavItem) {
             if selection != newSelection {
                 selection = newSelection
-                while !navPath.isEmpty {
-                    navPath.removeLast()
-                }
-                navPath.append(newSelection)
+                navPath = NavigationPath([newSelection])
             }
         }
     }
--- a/ViewModel/BrowserTabPreloader.swift
+++ b/ViewModel/BrowserTabPreloader.swift
@@ -1,0 +1,82 @@
+// This file is part of Kiwix for iOS & macOS.
+//
+// Kiwix is free software; you can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 3 of the License, or
+// any later version.
+//
+// Kiwix is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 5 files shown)


kelson42:
We need here a proper PR description.

BPerlakiH:
> We need here a proper PR description.

Please see updated.

kelson42:
> > We need here a proper PR description.
> 
> 
> 
> Please see updated.

To me, what has been implmented is not what I proposed and I don't understand why:
* this background load does not happen in the background. It should not impact the start of the app?
* why is is not on macOS?
* what about the improvements of the previous closed PR?

BPerlakiH:
> To me, what has been implemented is not what I proposed and I don't understand why:
> 
>     * this background load does not happen in the background. It should not impact the start of the app?

It does happen in the background, here:
https://github.com/kiwix/kiwix-apple/pull/1587/changes#diff-c5054c083648d0150e9aa4a615151e4f8c0c4918187866de5806def16b0e9858R63

we use a detached utility Task for it, and making sure we are not affecting the start up time, and start preloading once the initial tab is fully ready.

>     * why is is not on macOS?

On macOS, in most cases you will only have 1 browser on start up:
Screenshot 2026-05-25 at 09 43 39

There are ways to have other "tabs", but those are actually windows, containing a new instance of the entire view hierarchy. That window management from Apple is problematic, has completely different properties and restoration challenges, we are yet to face.
 
>     * what about the improvements of the previous closed PR?

Those were animations only, the core of the issue is how long it actually takes to go from one tab to another, and if that browser instance is ready or not (plus the content).

kelson42:
So, macOS does not remember tabs over restarts? and iOS does not have tabs at all?

> Those were animations only, the core of the issue is how long it actually takes to go from one tab to another, and if that browser instance is ready or not (plus the content).

OK, but would that improve the overall feeling? Because from your screencasts, the difference between before and after is everything but obvious to me.

BPerlakiH:
> So, macOS does not remember tabs over restarts? and iOS does not have tabs at all?

Yes, macOS won't remember tabs over the restarts. 
iPhone do have tabs but the switch between them is different:

https://github.com/user-attachments/assets/7e351f7a-705f-46ed-9338-ecd40d049327

kelson42:
> > So, macOS does not remember tabs over restarts? and iOS does not have tabs at all?
> 
> Yes, macOS won't remember tabs over the restarts. iPhone do have tabs but the switch between them is different:

Still see no valid reason to slow down the loading of tabs by not applying the pre-loading!?

kelson42:
@BPerlakiH Any news on this PR? This is the last one to complete the tav loading speedup effort

BPerlakiH:
> @BPerlakiH Any news on this PR? This is the last one to complete the tav loading speedup effort

Yes, I am on it, will update soon.

App/CompactViewController.swift
BPerlakiH, line 1: just some cleanup

App/SplitViewForiPad.swift
```diff
@@ -290,10 +293,7 @@ struct SplitViewForiPad: View { // swiftlint:disable:this type_body_length
         if let newNavItem, let newSelection = MenuItem(from: newNavItem) {
             if selection != newSelection {
                 selection = newSelection
-                while !navPath.isEmpty {
-                    navPath.removeLast()
-                }
-                navPath.append(newSelection)
+                navPath = NavigationPath([newSelection])
```
BPerlakiH, line 296: minimal optimisation

BPerlakiH:
@kelson42 I've updated this PR.
It is now applicable for both iPad and iPhone.
I did check macOS as well, that doesn't need this optimization. MacOS is already loading the window contents in parallel on start and on session restoration (regardless of the window arrangement being side by side or in forms of tabs).

Issue #1589, opened 2026-05-25
Prevent external PR from changing workflows


rgaudin:
While https://github.com/kiwix/kiwix-apple/issues/1588 will change PR event to `pull_request` and ensure any foreign PR runs in `internal`, it is agreed that workflows are not something that external contributors should change.

We thus want to test and fail if an external PR contains edits to `.github/` prefix.

 ⚠️ ⚠️  this is just a convenience check but it doesn't limit the responsibility of allowing external contributors PR run. Secrets can be leaked from everywhere, not just from the workflow.

kelson42:
Do you think creating a `CODEOWNERS` file with the following content would be satisfying?:
```
# Assigns ownership of all files in the .github/workflows directory to `@kiwix/lieutenants`
.github/workflows/* @kiwix/lieutenants
```

See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

rgaudin:
No this is already covered in the `external` call. This ticket is about directly failing PR changing workflows

PR #1590, opened 2026-05-25
Stop using pull_request_target trigger in CI

### LINKED ISSUE #1588: Run CI on PR branch
Our current workflow setup for CI has several specificities:

- triggers on `pull_request_target`.
- use either `internal` (all secrets) or `external` (dev-related apple certs and creds) environment
- if the PR is from a different repo, it uses `external`, otherwise `internal`.
- `external` has a protection rule requiring review from either @BPerlakiH, @kelson42 or @rgaudin 

This was done on purpose to allow external contributors to make PR and have the CI run.

The main difference between `pull_request_target` event and `pull_request` one is that the first one runs on the PR target environment (usually `main` branch of this repo). This means that workflow updates are not tested because `main` workflows are executed. It's an additional protection from external contributors trying to leak our secrets.

The problem it poses is that team members can't test workflow updates in PR.
What we've been doing is to push to branches (changing the workflow to accept said branches) to validate the changes and then make a PR (with potentially failing CI).

Because we can't have both `pull_request_target` and `pull_request` events together (or tests would run twice), it's been decided that we'd switch to `pull_request` and be extra careful about accepting external contributor runs.

In addition, we'd improve the `authorise` job to fail to `external` should there be any change to the `.github/` prefix.

kelson42:
Fixes #1588

Instead use "simple" `pull_request` trigger which implies:
* Using the workflow in the CI (for everybody)
* Keep approving the CI run for external PR (but approver has to check properly the PR content to be sure CI secrets are not exposed)
* Implement #1589 in a second PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,9 @@
 name: CI
 
+permissions:
+  contents: read
+  pull-requests: write
+
 on:
   workflow_dispatch:
   pull_request:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,8 @@
 name: CI
 
 on:
-  pull_request_target:
   workflow_dispatch:
+  pull_request:
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


rgaudin (dismissed):
Good 👍 

Following CodeQL warning, can you also add permission settings on the workflow:

```yaml
name: CI
permissions:
      contents: read
      pull-requests: write
```

I believe we don't need anything else

rgaudin (approved):
👍

Issue #1592, opened 2026-05-25
Reactivate UI tests on iPadOs and iOS


kelson42:
... and fix broken CI on main

PR #1606, opened 2026-06-12
Persist state of tabs when backgrounding the app

### LINKED ISSUE #1604: Kiwix does not seem to remember anymore which tab was open last
On my iPad, with the latest Testflight version:
* I open many books with many tabs
* Last tab open is not the last one in the list
* I leave the app
* I do many activities with the tablet, so the Kiwix apps goes to hibernation

What I get 
* When I restart Kiwix, the tab open is the last of the list

What I expect
* See the last tab I was on

BPerlakiH:
Fixes: #1604 

## BEFORE:

https://github.com/user-attachments/assets/5e50bde7-08eb-45ac-aa4b-f75e5bae90f7

## AFTER:

https://github.com/user-attachments/assets/21f67332-69d8-4db0-934f-cedbc68becbe

### DIFF
--- a/App/SplitViewForiPad.swift
+++ b/App/SplitViewForiPad.swift
@@ -158,6 +158,7 @@ struct SplitViewForiPad: View { // swiftlint:disable:this type_body_length
                         labelFor(tab: tab)
                     }
                     .id(tab.id)
+                    .accessibilityIdentifier(tab.title ?? LocalString.common_tab_menu_new_tab)
                     .swipeActions(edge: .trailing, allowsFullSwipe: true) {
                         Button(role: .destructive) {
                             Task { [weak navigation] in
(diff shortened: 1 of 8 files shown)


kelson42:
@BPerlakiH we really need an automated test for this

BPerlakiH:
> @BPerlakiH we really need an automated test for this

I have started to adding the UI test modifications to this PR.
While doing that, I've discovered some navigation issues described in:
https://github.com/kiwix/kiwix-apple/issues/1607
for which there's a fix here:
https://github.com/kiwix/kiwix-apple/pull/1608

That should be merged first, and then we can include the proper UItests for this.
It won't be running on CI first, only locally. The CI part will be handled in:
https://github.com/kiwix/kiwix-apple/issues/1592

BPerlakiH:
@kelson42 I've updated the UITests with this case, since it's currently not running yet on CI, it can be merged now separately from the other PR.

PR #1620, opened 2026-06-26
Rename Main Page and New Tab to Start Page

### LINKED ISSUE #1618: New tab is now first in the list
It should stay the last item like before

Image

BPerlakiH:
Fixes: #1618 

As mentioned in the issue, I think we are trying to solve the wrong problem.
We are already doing what Safari does, ordering the tabs by the time of creation, it is just named the wrong way.

## Safari

https://github.com/user-attachments/assets/f68165fa-febc-41f2-bd16-01019b4a53bd

## Kiwix iPad

https://github.com/user-attachments/assets/c0430af1-5e47-4f34-9186-3a8b9aa1be77

## Kiwix iPhone
IMG_3615 Large

### DIFF
--- a/Support/en.lproj/Localizable.strings
+++ b/Support/en.lproj/Localizable.strings
@@ -57,7 +57,7 @@
 
 "common.tab.manager.title" = "Tabs Manager";
 "common.tab.navigation.title" = "Tabs";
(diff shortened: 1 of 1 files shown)


BPerlakiH:
Closing it as it is not addressing the updated requirements.

PR #1623, opened 2026-06-28
Update tab bar menu for iOS, replace crashing overlay sheets

### LINKED ISSUE #1621: iOS crash on popover animation
We do have a crash on iPhone, which is rare, but can be achieved.
It seems to be related to the more tab icon and the navigation.
It is related to this warning in Xcode:
```
Trying to dismiss the presentation controller while transitioning already. (<UIPopoverPresentationController: >)
```

I am still investigating it, so far I got the stack trace, but it's not really helpful

BPerlakiH:
Fixes: #1621

After lots of back and forth, I could not find a reliable way to "fix" the former solution.
Due to the fact that it was a bit "hacky" in the first place, I decided to do it in a different way, that is way less prone to the same problem, namely:
- the overlays are not synchronized in an understandable way on iOS, and having more than one at the time is causing issues

After my changes here's how it looks

# iPhone:

https://github.com/user-attachments/assets/948ddda8-bdd7-4853-85dc-aa6ad197f1da

# iPad:
**it is not affected** when in full-screen. 
But the iPhone "version" **can be tested on an iPad** by resizing the window to "compact" size, like so:

simulator_screenshot_B589BD95-3FD3-447B-B27E-E66E21B6E723


kelson42:
What is different from a user perspective?

BPerlakiH:
@kelson42 the original version is in the issue, but putting it here as well for comparison:
https://github.com/user-attachments/assets/b3cf4c90-73ab-47b8-8b88-45255de32fb7

The only minor drawback that I can see, is that the random button cannot be pressed many times as before, as you need to open the menu and press the random button, then open the menu again and press it again. I don't think it's going to be a big problem though, still better than the app crashing.

kelson42:
I still don't get it because the PR description does not state with simple and clear sentence what has changed from user perspective. What seems to be is that you have change significant,y the UI to fix a crash... does not seem OK as we didn't requested to change the user interface.

BPerlakiH:
@kelson42 I've updated it so it looks very close to what we had:

## BEFORE

https://github.com/user-attachments/assets/daf32925-b015-4e42-81ad-cd336d2e1aa3



## AFTER


https://github.com/user-attachments/assets/d38bbe5d-a3b5-40e6-a55d-c46250252b50

kelson42:
OK, solution sounds reasonable, lets give a try

Issue #1436, opened 2026-01-24
Impossible to save in image as file


kelson42:
The option is simply not given on my iPad

<img width="2160" height="1620" alt="Image" src="https://github.com/user-attachments/assets/1f4b57c3-da11-4909-bc34-4d61d877f06b" />

BPerlakiH:
@kelson42 I am not sure about this. iOS is very specific in this regard, and I find it very counterintuitive to work across apps.
On my side I see the following options, and I could save the image by long pressing on it, and selecting "Save to Photos":

<img width="2048" height="1536" alt="Image" src="https://github.com/user-attachments/assets/fb8e0b1b-974d-48d2-b277-683ed3b889ea" />

kelson42:
This is what is displayed for the same article, but online, with FF on the same device

<img width="2160" height="1455" alt="Image" src="https://github.com/user-attachments/assets/baf90223-504d-4c55-8b2b-8999acebe030" />

And on safari

<img width="2160" height="1620" alt="Image" src="https://github.com/user-attachments/assets/07303901-2c19-463a-8171-32e65061b6f6" />

Looks good to me. It does not seems to be a problem with iOS.

BPerlakiH:
I did try the same content by switching the category language to Bambara, and downloading the Wikipedia as:

<img width="1024" height="768" alt="Image" src="https://github.com/user-attachments/assets/cba200f3-35da-4735-9cab-bbd8c07a2b5d" />

I assume it is the same content.
I cannot reproduce this issue, I do see a way to save the image if I press and hold on it.

<img width="1024" height="768" alt="Image" src="https://github.com/user-attachments/assets/b42a3c26-70a3-490a-aeab-3e1fe5f794c5" />


I've noticed, that once I press and hold on the image it becomes "larger". I do not see the same effect on your original (1st) screenshot though.

kelson42:
@BPerlakiH I feel sorry you can not reproduce the problem, but we still have to do something. I simply guess, your iPad does not have latest beta iPadOS!?

BPerlakiH:
Mine is on latest stable: iPadOS 26.2.1

kelson42:
I have 26.3

BPerlakiH:
@kelson42 I have retested this with the latest main branch of Kiwix, and on iPad 26.3.1.
Downloaded the same ZIM file as above, and I can save the image properly, after "long pressing" it:

[...]

PR #1626, opened 2026-06-28
Change photos permission to write only

BPerlakiH:
Related to: #1436 

It might fix the issue... I am not sure, still cannot reproduce it. 
As far as I know we don't need to read from Photos, so a write only permission to Photos folder should be enough for the Kiwix app.


kelson42:
This PR just changes a label!?!

The fact that your iPad with exactly the same configuration behaves differently is the first thing to understand and fix. We don't fix things if we are blind.

BPerlakiH:
It changes the kind of permission we are requesting:
[INFOPLIST_KEY_NSPhotoLibraryUsageDescription](https://developer.apple.com/documentation/bundleresources/information-property-list/nsphotolibraryusagedescription)
vs
[INFOPLIST_KEY_NSPhotoLibraryAddUsageDescription](https://developer.apple.com/documentation/bundleresources/information-property-list/nsphotolibraryaddusagedescription)

[As the Apple documentation says](https://developer.apple.com/documentation/bundleresources/information-property-list/nsphotolibraryusagedescription#Discussion):

> If your app only adds assets to the photo library and does not read assets, use the [NSPhotoLibraryAddUsageDescription](https://developer.apple.com/documentation/bundleresources/information-property-list/nsphotolibraryaddusagedescription) key instead.

kelson42:
pretty sure it does not change anything, but at list you fixed the typo.

PR #1628, opened 2026-06-29
Keep new tab at the end of the list and reuse existing empty tabs

### LINKED ISSUE #1618: New tab is now first in the list

garvitkaushik-123:
Fixes #1618

## Summary
- Changed tab list sort order on iPhone from descending to ascending (by creation date), so new tabs appear at the **end** of the list — matching the iPad sidebar which already sorted this way.
- Added `openNewTab()` in `NavigationViewModel` that checks for an existing empty tab (no associated ZIM file) before creating a new one. If one exists, it switches to it instead of creating a duplicate.
- Updated all "New Tab" button actions (iPhone tab manager + iPad sidebar) to use `openNewTab()`.

## Test plan
- [ ] Open the app on iPhone, create a new tab — verify it appears at the bottom of the tab list
- [ ] Tap "New Tab" again — verify it switches to the existing empty tab instead of creating a second one
- [ ] Open a ZIM article in the empty tab, then tap "New Tab" — verify a new empty tab is created at the end
- [ ] Repeat on iPad sidebar — same behavior expected

### DIFF
--- a/App/SplitViewForiPad.swift
+++ b/App/SplitViewForiPad.swift
@@ -81,7 +81,7 @@ struct SplitViewForiPad: View { // swiftlint:disable:this type_body_length
                 } label: {
                     Label(LocalString.common_tab_menu_new_tab, systemImage: "plus.square")
                 } primaryAction: {
-                    navigation.createTab()
+                    navigation.openNewTab()
                 }
             }
         } detail: {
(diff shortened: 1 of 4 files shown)


BPerlakiH:
@garvitkaushik-123
Your solution is not covering many cases that we are already handling in the other PR,
that has been opened before yours:
https://github.com/kiwix/kiwix-apple/pull/1622

Issue #1633, opened 2026-07-03
Update library listing to support multiple languages


kelson42:
via drag-drop and librairies should be sorted accordingly.

BPerlakiH:
@kelson42 I think the display part of it will need some further discussions.
I dig deep into it, but it's not so obvious.

Currently we are displaying ZIM files in categories 2 ways, either as a list or as a grid:
https://github.com/kiwix/kiwix-apple/blob/b7f506f81a3778ea13ec09ee2f86006c9ee2195c/Views/Library/ZimFilesCategories.swift#L75-L80


## Grid (grouped by ZIM file name)

<img width="1280" height="889" alt="Image" src="https://github.com/user-attachments/assets/1f69c780-b1f6-4105-a39c-f39fdba285a8" />

Here the original idea was, that many files come in different flavours (max, nopic, mini) all named the same, so they are grouped together by that. In most cases this works, however for a multi-language user, file names can sometimes match even in multiple languages, and they all will be thrown into the same section (not to mention that it's hard to tell which one is in which language). For example here the title "Wikipedia" matches for both English and Spanish:

<img width="1280" height="341" alt="Image" src="https://github.com/user-attachments/assets/7a26be8e-6fe2-44e2-bf8c-f7e038391e97" />



## List (no grouping)

<img width="1280" height="889" alt="Image" src="https://github.com/user-attachments/assets/22242290-b6d8-4cd0-9350-94b0f605b4c1" />


-----------------

## Limitations / Recommendations by Apple

Applying the very same group strategy, won't work for a multi language user as we would need to group the files by both language and by title as well. However, it is not recommended and not even easily supported by Apple to have 2 level nested grids, such as:
- English:
  - Tenis by Wikipedia:
     - max
     - nopic
  - Wikipedia:
     - max
     - nopic
     - mini
- Spanish:
  - Wikipedia:
     - max
     - nopic
     - mini

Even if we force it, it might have performance issues.

# Things I have tried

## Multi language files
Most files have 1 language, however there are exceptions, where they come with a very long list of language codes (such as Phet) or a pair of languages. Trying to group sort them might be an issue. Even if we create a multi-language category, we should still need to handle the sorting of that multi-language group: should it come first in the list or last in the list...

### Pair of languages:

<img width="1280" height="889" alt="Image" src="https://github.com/user-attachments/assets/98f5cdad-dc64-4ec6-9654-c2909a80534f" />

### Multi languages:

[...]

PR #1634, opened 2026-07-04
Migrate, and make category languages sortable

BPerlakiH:
Related to #1633 

I've split the task into 2 parts, this is part 1 making the category languages sort-able.
Changed:
- the saved structure of the preferred languages was changed from a Set to an Array
- for that to work fluently, I've added a migration, so updating users won't loose their existing preferences
- the macOS interface was fully updated
- improved the iOS one, making sure that the keyboard is not fully covering the languages we are actually searching
- added a custom group to sort by name or count (inspired by the table formerly used on macOS)

# iPhone

https://github.com/user-attachments/assets/927fa057-26c6-498a-9773-4ea81f8a912d

# iPad

https://github.com/user-attachments/assets/f49b6fda-5893-4eec-a796-5c778c093de5

# mac

https://github.com/user-attachments/assets/700245e6-c1b6-4594-9424-9b63896a0816


kelson42:
@BPerlakiH Looks good, but IMHO we should add the standard visual on sortable items to show that we can do that.

BPerlakiH:
@kelson42 please see it updated:

ipad

macOS

PR #1644, opened 2026-07-18
Update the logic for catalog sync

### LINKED ISSUE #1641: Update the logic for catalog sync
As it came to light here:
https://github.com/kiwix/kiwix-apple/issues/1636#issuecomment-4952834616

> Is it possible that the book entries were updated, especially their download URL, but were published under the same ID ?

>> Absolutely. The UUID is about the content, not the download URL. That's mistake on the reader side to assume that the download URL wouldn't change.

We need to update the logic to reflect it.
My only question remains if it should apply to all metadata fields not just the downloadURL ?

(eg: can the title, article count, size ect. of a ZIM file can be updated without changing UUID) ?

BPerlakiH:
## Impact for end user
After updating the the application, on the first successful catalog fetch the catalog entries will be up to date, ready for download.

## Removed:
Old fallback logic from 2024 (see: #672) for users that were on still on old ISO language codes.
Chances of users updating the app from 2024 to today will be very slim, so it's safe to remove this.

## Changed to a new approach to sync:
Sync the local DB entries with the parsed feed:
1) do not touch already downloaded files, those are not synced
2) remove local items that are not in the feed
2) insert / replace entries that are new from the feed

Notes: the remaining discussion, and notes will be continued under:
https://github.com/kiwix/kiwix-apple/issues/1673

### Tested on
- [x] **iPhone**
- [x] **iPad**
- [x] **mac**

### DIFF
--- a/Model/OPDSParser/Parsed.swift
+++ b/Model/OPDSParser/Parsed.swift
@@ -18,12 +18,4 @@ import Foundation
 /// Sendable data that is comming out of the OPDSParser
 struct Parsed {
     let results: [UUID: ZimFileMetaStruct]
-    
-    /// An empty result we can use to delete zim entries
-    /// Based on the assumption we insert new ones, delete the ones not on the list
-    /// Therefore an empty list will delete everything, using the same method
-    /// @see: LibraryViewModel.process(parsed: Parsed)
-    static func deletingResult() -> Parsed {
-        Parsed(results: [:])
-    }
 }
--- a/SwiftUI/Model/DefaultKeys.swift
+++ b/SwiftUI/Model/DefaultKeys.swift
@@ -41,7 +41,6 @@ extension Defaults.Keys {
         "libraryLanguageSortingMode", default: LibraryLanguageSortingMode.byCounts
     )
     static let libraryAutoRefresh = Key<Bool>("libraryAutoRefresh", default: true)
-    static let libraryUsingOldISOLangCodes = Key<Bool>("libraryUsingOldISOLangCodes", default: true)
     static let libraryLastRefresh = Key<Date?>("libraryLastRefresh")
 
     static let isFirstLaunch = Key<Bool>("isFirstLaunch", default: true)
--- a/Tests/TestDefaults.swift
+++ b/Tests/TestDefaults.swift
@@ -25,7 +25,6 @@ final class TestDefaults: NSObject, Defaulting {
         self[.categoriesToLanguages] = [:]
         self[.libraryAutoRefresh] = false
         self[.libraryETag] = ""
-        self[.libraryUsingOldISOLangCodes] = false
         self[.libraryLanguageCodes] = []
     }
     
--- a/ViewModel/LibraryViewModel.swift
+++ b/ViewModel/LibraryViewModel.swift
@@ -47,6 +47,11 @@ enum LibraryState {
 
 @MainActor
 final class LibraryViewModel: ObservableObject {
+    private struct SyncCount {
+        var inserted: UInt = 0
+        var deleted: UInt = 0
+    }
+    
     @Published private(set) var error: Error?
     /// Note: due to multiple instances of LibraryViewModel,
     /// this `state` should not be changed directly, modify the `process.state` instead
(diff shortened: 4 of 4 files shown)


kelson42:
Not bad, but not optimal either:
* don't empty the whole catalog, because in case of problem you will manage to have a full empty catalog
* Never update downloaded books metadata (no idea why you should update the download URL!)
* Update books in local cache with books of OPDS if same uuid. Insert otherwise.
* Delete books from the cache if it has disappeared from the OPDS
* Hide books in catalog if it is already downloaded. If local version removed, then unhide.

BPerlakiH:
> no idea why you should update the download URL

My thinking behind it is the following, that refreshing the catalog is a separate process, plus it's behind a cache, if we don't update the download URL of an already downloaded book, that information "gets lost".
At any point in time, the user can do the following (independently from the catalog refresh):
- right click or long tap on the already downloaded file to see it's download URL
- delete the downloaded file, and try to re-download it again

BPerlakiH:
> Hide books in catalog if it is already downloaded. If local version removed, then unhide.

This might be tricky...
Currently the very same DB row is re-used. It starts as an entry from ODPS, which we list for the user. Once the download starts, we "enrich" the same row with:
- download progress
- resume data (it's small in size)
Once the download completes, we further "enrich" the same row with:
- file system path to the real ZIM file
- updating the metadata with the metadata from the ZIM file (the same process is used for side loaded ZIMs)
- plus as on the diagram above: include/exclude from search results, integrity check result

From the UI perspective, it looks like this:
- list of entries
- details of one
- start downloading 
- see the progress in the details
- download complete

Therefore hiding the book would mean, that once we finish the download it should disappear from the list of entries. That might complicate the handling of the details page at this point (or maybe not).

kelson42:
> > no idea why you should update the download URL
> 
> My thinking behind it is the following, that refreshing the catalog is a separate process, plus it's behind a cache, if we don't update the download URL of an already downloaded book, that information "gets lost".
> At any point in time, the user can do the following (independently from the catalog refresh):
> - right click or long tap on the already downloaded file to see it's download URL
> - delete the downloaded file, and try to re-download it again

This is not something needed and can not work that way because the library keeps moving every hour.

kelson42:
> Currently the very same DB row is re-used.

Doing so can only be a source of problems... Local library is different from OPDS catalog. Once decided to download (not even once downloaded because the book can disappear from the OPDS catalog one second after the click... Actually it can already have been removed from the online catalog before the download click), then we should split.

BPerlakiH:
> > Currently the very same DB row is re-used.
> 
> Doing so can only be a source of problems... Local library is different from OPDS catalog. Once decided to download (not even once downloaded because the book can disappear from the OPDS catalog one second after the click... Actually it can already have been removed from the online catalog before the download click), then we should split.

@kelson42 This has been the approach on the Apple side since I remember. The only good reason that I could see for this was to be able to display the local files in the same listing as the online catalog... Since you are requesting to keep the listing separated, plus - from what I am understanding from these use-cases - we always want the "most up to date" version of the online catalog, we should possibly rethink how to achieve that.
Downloading and caching everything with these requirements start to have little sense. Even the "details page" we show right before the download should be up to date (ideally). So perhaps we could move the download items into a separate DB table with only the relevant metadata for those (and migrate existing users), and optimize on the catalog request / response side to make sure we can have everything up to date, and think about what to display if the user is offline and tries to use the catalog.
Maybe we need a separate plan for all this.

BPerlakiH:
@kelson42 I've updated this PR, as an intermediate solution as we discussed before a view "live" what is online in the catalog solution.

On this point: 
> don't empty the whole catalog, because in case of problem you will manage to have a full empty catalog

It is only really about the the not downloaded entries in the DB.
I've tried to do an update of the entries that are in the catalog already, but it's really slow (no bulk operation solution for that one). So instead we can stick to the solution of removing the entries that are not yet downloaded and inserting those from the parsed entries, it's quicker and simpler.

BPerlakiH:
@kelson42 can we merge this ?

kelson42:
@BPerlakiH Can you please update the PR description because now, I'm not sure what it exactly does. Please open issues if there is anything left to do.

BPerlakiH:
> @BPerlakiH Can you please update the PR description because now, I'm not sure what it exactly does. Please open issues if there is anything left to do.

@kelson42 done, and opened a new task for the bigger piece of work.

PR #1647, opened 2026-07-19
Multi-language category filtering

### LINKED ISSUE #1633: Update library listing to support multiple languages

BPerlakiH:
## Impact for end user
Now it is possible to filter down the list of items in the categories by selected language. The order of the languages is set in the app settings.

For users with a single language, there's no difference, since we won't display any additional buttons to filter.

## **iPhone**

https://github.com/user-attachments/assets/eb12f1c3-d40a-446f-af84-e1932bfb51c4

## **iPad**
https://github.com/user-attachments/assets/f973f39a-249e-4a0f-82e9-48742e26b3f9

## **mac**

https://github.com/user-attachments/assets/422ee1e9-19cf-4edb-ba33-825d7a9a31ac

### DIFF
--- a/App/CompactViewController.swift
+++ b/App/CompactViewController.swift
@@ -118,9 +118,9 @@ struct CompactTabView: View {
         .sheet(item: $presentedSheet) { presentedSheet in
             switch presentedSheet {
             case .library(downloads: false):
-                Library(dismiss: dismiss)
+                LibraryTab(dismiss: dismiss)
             case .library(downloads: true):
-                Library(dismiss: dismiss, tabItem: .downloads)
+                LibraryTab(dismiss: dismiss, tabItem: .downloads)
             case .customHotspot:
                 SheetContent {
                     HotspotZimFilesSelection()
--- a/App/RootView_macOS.swift
+++ b/App/RootView_macOS.swift
@@ -16,6 +16,7 @@
 #if os(macOS)
 import SwiftUI
 import PassKit
+import Defaults
 
 // swiftlint:disable:next type_body_length
 struct RootView: View {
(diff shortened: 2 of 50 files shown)


kelson42:
@BPerlakiH what about the "new" tab... Which should also be filtered by language.

Otherwise it works significantly differently on iPhone and iPad... But the only one big problem is on iphone where we can only filter categories by languages.... Not the books!

BPerlakiH:
@kelson42 the new tab is already included:

# iPhone

https://github.com/user-attachments/assets/88034680-b61b-41c0-a261-6b205b56fb9b

# iPad

https://github.com/user-attachments/assets/8fa4e6bc-bbf4-4aa9-9eb9-5e643c216a69

# macOS


https://github.com/user-attachments/assets/f3022611-1214-4228-ba24-b82a78384895

BPerlakiH:
> But the only one big problem is on iphone where we can only filter categories by languages.... Not the books!

On the iPhone you filter both the categories and the books, it is just that the filter is at the "top level", and the selected language gets "passed down" to the next screen where we list the books.


https://github.com/user-attachments/assets/ef592fde-7a93-4b91-9200-782e6bd9564c

kelson42:
> > But the only one big problem is on iphone where we can only filter categories by languages.... Not the books!
> 
> On the iPhone you filter both the categories and the books, it is just that the filter is at the "top level":
> 
> 
> https://github.com/user-attachments/assets/ef592fde-7a93-4b91-9200-782e6bd9564c

Yes, this is the problem: people are all the time at book level and can not switch easily.

BPerlakiH:
@kelson42 OK, I've updated it:

https://github.com/user-attachments/assets/bf40ca6f-dec6-4454-9af2-72d24ea7dd09

kelson42:
We are OK, would that mean we are good for  merging?

BPerlakiH:
yes, all green, good to be merged.

PR #1672, opened 2026-08-08
1671 Fix crash on accessing fileID after download cancellation

### LINKED ISSUE #1671: Download task crashes
**Describe the bug**
The clean up process of a DownloadTask can lead to crashes:

Image

After digging into this, there's only 1 place where we really read the .fileID property that can result in such crash:
https://github.com/kiwix/kiwix-apple/blob/f5965f7fb91a38f9f30a3b7fa1b9df64e31ae808/Views/Library/ZimFilesDownloads.swift#L40

**iPhone/iPad (please complete the following information):**
 - Device: iPhone, iPad
 - Kiwix Version 3.16.1

BPerlakiH:
## Impact for end user
Cancelling download tasks should not cause sporadic crashes any more.

## Description

- Changed:
Make sure deletion is not colliding with the UI update cycle
Rely on a copy of plain data where possible rather than CoreData references to properties, as CoreData can pull the plug on any of the ``@NSManaged`` properties on deletion.

### Tested on
- [x] **iPhone**
- [x] **iPad**
- [x] **mac**

### DIFF
--- a/Model/Downloads/DownloadService.swift
+++ b/Model/Downloads/DownloadService.swift
@@ -109,7 +109,7 @@ final class DownloadService {
             let errorMessage = LocalString.download_service_error_option_directory
             DownloadUI.showAlert(.downloadErrorZIM(zimFileID: zimFileID, errorMessage: errorMessage))
             task.cancel()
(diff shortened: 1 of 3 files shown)


kelson42:
I would be happy to merge, but the CI fails

PR #1687, opened 2026-08-23
Immerse reading based on webview scroll

### LINKED ISSUE #1682: Searchbar is not always visible properly with maps
I open a map ZIM file, here the one of Greece.

Then I go to the bottom of the map whith swiping up. Searchbar disappears.

Finally at the bottom, I want to search for a city and swwipe down to make the searchbar reappearing... but it does not. It seems I have to go back to the top of the map to see it again!

I want to see the searbar reappearing as soon as I swipe down, like in any other book.

image

BPerlakiH:
## Impact for end user
The navigation bars will be hidden / show in a more appropriate manner.
The user won't be "stuck" on pages without being able to access the navigation icons.

## Description
It turned out that this is a wider issue not specific only to maps, and depends on the content itself. If the scroll is happening on the "top level" everything works OK. However maps and certain wiki home pages do the scrolling not on the top level, but inside a container within the html DOM. This one was not triggering the native scroll events.

## Solution
Instead of relying on the native top level scroll position, we can rely on the html document element, which is more accurate in our cases. For the content where the scroll is happening within these inner containers, like map and wiki home pages we won't be hiding the navigation/tool-bar items.

## BEFORE:

https://github.com/user-attachments/assets/398f0680-4bf6-488f-870f-7585a9093d21
https://github.com/user-attachments/assets/8c6987b3-d1c6-4e20-b0c1-39b7ed83d1b9

## AFTER:
https://github.com/user-attachments/assets/f52a3dd0-21d5-4fe1-972c-a2d225f197bc

### Tested on
- [x] **iPhone**
- [x] **iPad**
- [x] **mac**

### DIFF
--- a/Support/injection.js
+++ b/Support/injection.js
@@ -28,14 +28,23 @@ let observer = new IntersectionObserver(function(entries) {
 	}
 }, { rootMargin: '-50px 0 -35% 0', threshold: 1.0 });
 
-// register scroll view to handle heading on top of the page
+// register scroll view to handle custom show/hide navigation bars
+var lastYOffset = 0;
+var lastDirection = undefined;
 window.onscroll = function() {
-	if (document.documentElement.scrollTop <= 0) {
-		const headingVisible = window.webkit.messageHandlers.headingVisible
-		if(headingVisible !== undefined && headingVisible.postMessage !== undefined) {
-			headingVisible.postMessage({id: headings[0].id})
-		}
-	}
+    const element = document.documentElement;
+    const maxYOffset = element.scrollHeight - element.clientHeight;
+    // trim off the values created by bounce-back animation at the top & bottom of the page
+    // which are over the normal scroll min/max values
+    const yOffset = Math.min(Math.max(element.scrollTop, 0), maxYOffset);
+    if(lastYOffset < yOffset && 50 < yOffset && lastDirection != "down") {
+        lastDirection = "down";
+        window.webkit?.messageHandlers?.scrollHandler?.postMessage("down");
+    } else if(yOffset < lastYOffset && lastDirection != "up") {
+        lastDirection = "up";
+        window.webkit?.messageHandlers?.scrollHandler?.postMessage("up");
+    }
+    lastYOffset = yOffset;
 }
 
 // expand all detail tags
--- a/ViewModel/BrowserViewModel.swift
+++ b/ViewModel/BrowserViewModel.swift
@@ -108,6 +108,8 @@ import CoreKiwix
     private var canGoForwardObserver: NSKeyValueObservation?
     private var titleURLObserver: AnyCancellable?
     private let bookmarkFetchedResultsController: NSFetchedResultsController<Bookmark>
+    /// Maps to injected javascript: window.webkit?.messageHandlers.*
+    private let jsHandlers = ["headings", "geolocation", "scrollHandler", "zimCookies"]
 
     // MARK: - Lifecycle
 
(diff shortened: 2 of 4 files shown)


Support/injection.js
```diff
@@ -28,14 +28,25 @@ let observer = new IntersectionObserver(function(entries) {
 	}
 }, { rootMargin: '-50px 0 -35% 0', threshold: 1.0 });
 
-// register scroll view to handle heading on top of the page
+// register scroll view to handle custom show/hide navigation bars
+var lastYOffset = 0;
+var lastDirection = undefined;
 window.onscroll = function() {
-	if (document.documentElement.scrollTop <= 0) {
-		const headingVisible = window.webkit.messageHandlers.headingVisible
```
BPerlakiH, line 34: This "headingVisible" was not triggering anything actually (dead code).

ViewModel/BrowserViewModel.swift
```diff
@@ -108,6 +108,8 @@ import CoreKiwix
     private var canGoForwardObserver: NSKeyValueObservation?
     private var titleURLObserver: AnyCancellable?
     private let bookmarkFetchedResultsController: NSFetchedResultsController<Bookmark>
+    /// Maps to injected javascript: window.webkit?.messageHandlers.*
+    private let jsHandlers = ["headings", "geolocation", "scrollHandler", "zimCookies"]
```
BPerlakiH, line 113: joined together the javascript message handler ids, so it's less repetition in code, and less error prone.

Views/BuildingBlocks/WebView.swift
```diff
@@ -172,8 +173,21 @@ final class WebViewController: UIViewController {
         super.didMove(toParent: parent)
         // navigated to webView
         if !Brand.disableImmersiveReading, let parent {
-            parent.navigationController?.hidesBarsOnSwipe = true
```
BPerlakiH, line 175: This was the native handling of the scroll to show/hide nav items, which wasn't always triggering.

Views/BuildingBlocks/WebView.swift
```diff
@@ -172,8 +173,21 @@ final class WebViewController: UIViewController {
...
+                    case "up":
+                        navController?.setNavigationBarHidden(false, animated: true)
+                        navController?.setToolbarHidden(false, animated: true)
+                    case "down":
+                        navController?.setNavigationBarHidden(true, animated: true)
+                        navController?.setToolbarHidden(true, animated: true)
+                    default: break
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
BPerlakiH: That's not completely true. ``WebViewController`` is inheriting ``@MainActor`` from ``UIViewController``:
```
@available(iOS 2.0, *)
@MainActor open class UIViewController
```
and the new task created will inherit ``@MainActor`` by default as well.

For the sake of clarity and to be on the safe side, I've marked it as such, updated the ``==`` to ``===`` and retested on iPad. Looks good to me.

Support/injection.js
```diff
@@ -28,14 +28,23 @@ let observer = new IntersectionObserver(function(entries) {
...
+    if(lastYOffset < yOffset && 50 < yOffset && lastDirection != "down") {
+        lastDirection = "down";
+        window.webkit?.messageHandlers?.scrollHandler?.postMessage("down");
+    } else if(yOffset < lastYOffset && lastDirection != "up") {
+        lastDirection = "up";
+        window.webkit?.messageHandlers?.scrollHandler?.postMessage("up");
+    }
+    // trim off the values created by bounce-back animation at the top & bottom of the page
+    // which are over the normal scroll min/max values
+    lastYOffset = Math.min(Math.max(yOffset, 0), maxYOffset);
```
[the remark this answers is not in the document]
BPerlakiH: Updated and retested.

PR #1717, opened 2026-09-12
Unify ZIM categories to grid listing

### LINKED ISSUE #1712: Impossible to have "no category" selected
See

Image

The problem is that you can not make a global search based on keyword!

Worse: At first install time, the user as per default the category "Wikipedia", and might not even fully and quickly understand that actually he searches around a minority part of the catalog.

We should be able to search without category filter, with many categories at the same time.... and per default there ahould not be a filter at all.

BPerlakiH:
## Impact for end user
- Enable to search globally across all categories.
- While searching the category selection is hidden (it doesn't make a difference when changing it)
- Updated ZIM tiles and details for all sections (unified approach)

# Changed:
- Unified the cells to be one type
# Added:
- ZIM file description, and media count
- New localization for "pages" and "media" counts
# Removed:
- Old localization

## iPad

https://github.com/user-attachments/assets/8c7a4a3d-08d9-47a8-ae8f-9439cb4c40a4

## mac

https://github.com/user-attachments/assets/28f50f68-87bd-409e-b873-9f3a64196cf9

## iPhone

https://github.com/user-attachments/assets/4a19ae1f-67b5-4ac3-aecd-d9e4f158a9a6

### Tested on
- [x] **iPhone**
- [x] **iPad**
- [x] **mac**


BPerlakiH:
> I don't really understand how to filter with no categories or a selection of categories?!

My thinking behind it is the following, search is global, so you always search without the selection of categories. As you type into the search field, the category selector disappears. 
I think we can do even better than this as a follow up in: #1720

BPerlakiH:
> The logo is very small?! It was bigger before, isn't it?

I have noticed that the logo size is not always consistent across languages, this logo you mentioned has a text under it, which is also localized, and it is smaller than the same logo in other languages:

Simulator Screenshot - iPad Air 11-inch (M4) - 2026-09-12 at 20 26 46 2
Simulator Screenshot - iPad Air 11-inch (M4) - 2026-09-12 at 20 27 02 2
Simulator Screenshot - iPad Air 11-inch (M4) - 2026-09-12 at 20 26 59 2

BPerlakiH:
> Put something like 42K pages, 54K media on the same line, so we gain a whole line in total.

Yes I thought about that. The thing is that the shortening "K" is localized as well (eg: literally saying "thousand" or some short form of that), so it can become quite a bit longer in some cases, and not always fitting on 1 line:

punjabi_listing

I will try to gain some space, by trying to find something we could use instead of the "no picture" text in that "blue pill" in the bottom-right corner, the translation of that is too long.

kelson42:
Please put the file size below the counts.

BPerlakiH:
I've replaced the "no pic" text with an icon, and moved the size to the bottom:



Simulator Screenshot - iPad Air 11-inch (M4) - 2026-09-12 at 22 02 05 2
Simulator Screenshot - iPad Air 11-inch (M4) - 2026-09-12 at 22 00 55 2

kelson42:
Lets merge this and I will open issues if I still have things to improve

PR #1730, opened 2026-09-18
Inline donation link for compact screens

### LINKED ISSUE #1726: Donation visibility on iOS (not iPadOS, not macOS)
Since [`3.16`](https://github.com/kiwix/kiwix-apple/releases/tag/3.16.0) (July 20th), the app supports recurring donations.
At of today (**less than 2 months later**), there has only been two such subscriptions ; one of which being @kelson42.
Given there are approx. 500k users of the app, it seems low.

- There hasn't been single donations neither during that period
- Only one person started the payment flow (iPhone) but did not get through.
- Both subscriptions happenned on iPad

One hypothesis is that the _feature_ is not vsible enough: on macOS and iPad, it's visible in the sidebar. On iPhone, the extremely dominant platform, it's only available at the post-scroll bottom of the settings.

@siemsie @BPerlakiH @kelson42, can we suggest **unobtrusive** ways to improve visibility of this?
For instance, @kelson42 suggested that the heart icon could be _sometimes/randomly_ animated to bring attention. That would only work for the sidebar platforms of course.

BPerlakiH:
## Impact for end user
Ability to find donation more easily.

## Screenshots

IMG_3747 Large
IMG_3748 Large

### Tested on
- [x] **iPhone**
- [x] **iPad**
- [x] **mac**


kelson42:
@BPerlakiH LGTM. Can younplease make the border of this donation box red (thin line, 1 or 2 pixel in red like the heart logo)?

BPerlakiH:
@kelson42 I've added the border, also with reduced opacity, the full red was just too aggressive to my taste:

IMG_3749 Large
IMG_3750 Large

BPerlakiH:
~~I need to add one more thing before merging ...~~

Now it should be good, I've added the usual conditional check to see if the device is actually supporting payments, otherwise we won't display it.

